PR #18, opened 2014-07-31
Updated the names "vessels" and "vessel" to "sources" and "source"

fabdrol:
In a discussion on Slack about the API footprint, a point was made by Pavel Kalian about not naming the list of vessels "vessels", as this list could also include marinas, ground stations, ActiveCaptain backbone servers, etc:

> **tkurki** List is pretty ambiguous - vessels?
> **fabdrol** hmm
> **fabdrol** no problem with that
> **nohal** Don't like "vessels", it can be more
> (...)
> **nohal** What if the entity is a marina?
> **nohal** Or Jeff
> **tkurki** Targets? Items? entities?
> **fabdrol** *@nohal* that would call for a change in the specification as well

In this pull request I changed the name of the list with vessels to `sources` and an individual vessel is a `source`. The definition of a "source", in our context, is:
> Anything that produces valid Signal K data and makes this data available over an available network.

I'm looking forward to feedback!


fabdrol:
Outdated. Closing.

PR #49, opened 2015-01-08
Multiple values

### LINKED ISSUE #48: Multiple sources for the same value
It is quite possible for a key value to come from more than one device. eg position (lat/lon) could come from several gps enabled devices, and multiple depth sounders are not uncommon. We need a consistent way to handle this.

All the incoming values may well be valid in their own context, and it is feasible that all of them may be wanted, for instance, dispalying depth under each hull on a catamaran.

Hence discarding or averaging is not a solution, and since signalk is unable to derive the best way to handle multiple values it must always fall to a default action, with human over-ride when needed. 

I propose we should simply store all the options in the tree, and have the main 'source' reference the options?

Then simple rules can apply:
1) If its the first value for the key, it becomes the first option and the default value.
2) If another value with different source arrives, we add to options - if its our preferred source (from persistent config) we auto-switch to it, otherwise we just record it
3) Users can then view the source options for a key and select from list for a specific display. They might select courseOverGroundTrue.options.ttyUSB1 as the source for a specific need or they may select an average/mean/min/max etc. (This facility is not part of the spec, but implementation specific)
4) maybe needs a rule to trim unreferenced options (or all options) on output.
```
{"vessels":
    {"self":
        {"navigation":
            {"courseOverGroundTrue":
                {
                 "value": 102.29,
                 "source": "options.actisense",
                 "options":
                    {
                    "ttyUSB1":  **just a unique name, have to generate somehow from incoming
                        {  
                            "value": 99.2900009155,
                            "source": "/dev/ttyUSB1"
                            "timestamp": "2014-08-15-16:00:00.081",
                        },
                     "actisense":
                        {
                            "value": 102.29,
                            "source": "/dev/actisense",

                            "timestamp": "2014-08-15-16:00:00.081",
                            "src": "115",
                            "pgn": "128267"
                        }
                    }
                }
            }
        }
    }
}
```

tkurki:
Fixes #48

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "SignalK specification schema as an npm module with tests",
   "main": "index.js",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "mocha"
   },
   "repository": {
     "type": "git",
@@ -25,8 +25,9 @@
   },
   "homepage": "https://github.com/SignalK/specification",
   "dependencies": {
+    "JSONStream": "0.7",
     "chai": "1.9",
-    "tv4": "latest",
-    "JSONStream" : "0.7"
+    "mocha": "^2.1.0",
+    "tv4": "latest"
   }
 }
--- a/package.json
+++ b/package.json
@@ -4,7 +4,8 @@
   "description": "SignalK specification schema as an npm module with tests",
   "main": "index.js",
   "scripts": {
-    "test": "mocha"
+    "test": "mocha",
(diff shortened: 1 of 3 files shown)


fabdrol:
@tkurki should this still be merged? If/after the checks stop failing?

fabdrol:
@tkurki what should I do with this?

tkurki:
I'm working on a concrete impl & test for this.

fabdrol:
cool :)

PR #51, opened 2015-01-24
Group roll pitch and yaw

### LINKED ISSUE #45: Group roll, pitch and yaw under "attitude"
`roll`, `pitch` and `yaw` should be grouped together under `navigation.attitude` and get a single timestamp and source value.

timmathews:
Fixes #45 

Not sure how this fell off the radar, but this change groups roll, pitch and yaw into navigation.attitude. I think we should merge.


rob42:
Yes but we should also move rateOfTurn into attitude too. In fact rateOfYaw (rateOfTurn?), rateOfRoll, rateOfPitch

rob42:
No - we should avoid having 'value' in branches. It should be at the bottom only, even if its an object.
In my case in particular I use 'value' to denote the leaf.
However there is nothing to stop having a complex object, like navigation.position
          attitude.timestamp, 
          attitude.source, 
          attitude.roll, 
          attitude.rollRate, 
          attitude.pitch, 
          attitude.pitchRate, 
          attitude.yaw, 
          attitude.yawRate, 
etc

timmathews:
@rob42, I'm a little confused. When you say
> we should avoid having 'value' in branches. It should be at the bottom only, even if its an object.

do you mean something like this?

```json
"attitude": {
  "timestamp": "...",
  "source": "...",
  "value": {
    "roll": "...",
    "rollRate": "...",
    "pitch": "...",
    "pitchRate": "...",
    "yaw": "...",
    "yawRate": "..."
  }
}

**Comment by rob42** [comment]
Like this:
```
"attitude": {
  "timestamp": "...",
  "source": "...",
   "roll": "...",
    "rollRate": "...",
    "pitch": "...",
    "pitchRate": "...",
    "yaw": "...",
    "yawRate": "..."
  
}
```
My server uses these rules:
* timestamp/source parent is the name of the key, so above key is `attitude`
* siblings of timestamp/source are the values
* `.value` is an invisible  key, signifies a primitive value of the parent, eg courseOverGound(.value)=234
* `attitude`  has several keys, so its an object, with keys as object content

So far it allows my server to correctly differentiate primitive and object keys without external config help.

**Comment by timmathews** [comment]
Can you show an example with the invisible `.value`? I apologize for my denseness here.

**Comment by rob42** [comment]
By 'invisible' I mean that we dont use it as the values name. eg
```
"speedOverGround":{
          "timestamp":"2015-03-23T01:57:02.256Z",
          "source":"vessels.motu.sources.nmea.0183.RMC",
          "value":0.1517598
        }
```
is equivalent to `speedOverGround=0.1517598` where as in `position` we do care about the values name:
```
 "position":{
          "timestamp":"2015-03-23T01:57:02.256Z",
          "longitude":4.58006417,
          "latitude":51.9485185,
          "source":"vessels.motu.sources.nmea.0183.RMC"
        }
```
`position = {object}`, `position.latitude=51.9485185`

fabdrol:
@timmathews should this branch be removed or updated and merged?

rob42:
Update then merge I think?

PR #52, opened 2015-01-25
Electrical, AC

timmathews:
New addition to the spec: Electrical AC parameters. This covers PGNs 65001 - 65030.

### DIFF
--- a/schemas/groups/electrical_ac.json
+++ b/schemas/groups/electrical_ac.json
@@ -10,39 +10,15 @@
             "title": "Bus #1",
             "description": "Main AC power bus on the vessel",
             "properties": {
-                "timestamp": {
-                    "$ref": "../definitions.json#/definitions/timestamp"
-                },
-                "lineLineVoltage": {
-                    "type": "integer",
-                    "description": "Average RMS voltage measured between two hot legs of a 3-phase or split-phase bus.",
-                    "units": "Vrms",
-                    "minimum": 0,
-                    "maximum": 64255,
-                    "example": 480
-                },
-                "lineNeutralVoltage": {
-                    "type": "integer",
-                    "description": "Average RMS voltage measured between any hot leg and neutral in a 3-phase, split-phase, or single phase bus.",
-                    "units": "Vrms",
-                    "minimum": 0,
-                    "maximum": 64255,
-                    "example": 277
-                },
-                "frequency": {
-                    "type": "number",
-                    "description": "Average frequency of the bus.",
-                    "units": "Hz",
-                    "minimum": 0,
-                    "maximum": 501.9921875,
-                    "example": 60.0
-                },
                 "phaseA": {
                     "type": "object",
                     "title": "Bus #1, Phase A",
                     "properties": {
                         "timestamp": {
                             "$ref": "../definitions.json#/definitions/timestamp"
+                        },
+                        "source": {
+                            "$ref": "../definitions.json#/definitions/source"
                         },
                         "lineLineVoltage": {
                             "type": "integer",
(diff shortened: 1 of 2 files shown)


schemas/groups/electrical_ac.json
```diff
@@ -0,0 +1,998 @@
...
+                    "type": "object",
+                    "title": "Bus #1, Phase A",
+                    "properties": {
+                        "timestamp": {
+                            "$ref": "../definitions.json#/definitions/timestamp"
+                        },
+                        "source": {
+                            "$ref": "../definitions.json#/definitions/source"
+                        },
+                        "lineLineVoltage": {
```
tkurki: Is the 'lineLine' correct? Sounds strange.

schemas/groups/electrical_ac.json
```diff
@@ -0,0 +1,998 @@
...
+                            "units": "VA",
+                            "minimum": -2000000000,
+                            "maximum": 2211081215,
+                            "example": 10500
+                        }
+                    }
+                }
+            }
+        },
+        "generator": {
```
tkurki: What if there are multiple generators? Like batteries, engines.

tkurki:
First of all I'm no AC expert and don't know what all these readings are about.

I like the electrical.ac prefix.

There is a *lot* of repetition at least between the phases judging from a cursory glance at this. Should the schema be more DRY or are there notable differences that prohibit using one definition?

tkurki:
I think it would be a good idea to come up with a realistic, full blown example document. That could be used as a runnable test against the schema as well as part of the documentation, as schema by itself makes a bit confusing reading. See https://github.com/SignalK/specification/blob/511260c9de4b8ccb8a390bc2d54cdf516bf38df1/test/multiple_values.js.  

If you can come up with a sample json document I can patch in the test.

canboat:
Aw, another instance where the NMEA has made matters worse in trying to be precise.

In general AC electricity is generated in 3 phases. Three coils in the generator, each generating one phase, conducted using three 'line' conductors and one 'neutral' conductor. Depending on the setup, you either connect single phase consumers between phase and neutral (="Y" or "star") or between two phases (="Delta"). The voltage between two 'line' conductors is colloquially know as the _line voltage_ and the voltage between neutral and one line or phase as the _phase voltage_.

Side note: common usage in Europe is to generate phase-to-phase voltage of 400V and neutral-to-phase 230V, and connect consumers between 1 phase and neutral. The problem starts when people then say that the 'line voltage is 230 V' which is not correct -- the line voltage is 400 V and the phase voltage is 230 V.

I suggest we either use a strictly correct wording, that everybody understands, `neutralToPhaseVoltage` and `phaseToPhaseVoltage` or the more practical terms `lineVoltage` and `phaseVoltage`.

timmathews:
We can't actually blame NMEA for this one, at least not directly. They chose to adopt the J1939-75 standard for gensets and AC power, so these PGNs come from that.

lineLineVoltage is their terminology, I agree that it could probably be simpler if we used phasePhaseVoltage, phaseNeutralVoltage, etc. (Not sure the To is necessary). But, at least in North America, most non-commercial power isn't 3-phase, it's split-phase (which is a bad term meaning that we get our power from center-tapped transformers fed from a single leg of a three-phase generator). So, typical household power is 120/240V with two hot (Line in common terminology here) and a Neutral. And we call each of them a line and both together a line. Lazy yanks.

While we muse over names I will try to simplify the schema somewhat by using definitions to remove the repetition.

schemas/groups/electrical_ac.json
fabdrol: @timmathews maybe the `To` would be good to include, I agree with @tkurki that `lineLine` sounds weird.

fabdrol:
@timmathews this can be merged, shall I?

timmathews:
:+1: Go for merge.

This will give us parity with NMEA for AC info. We can address multiple generators, etc. later.

PR #53, opened 2015-01-25
Pull of initial ELECTRICAL schema

thomasonw:
Initial schema to allow wider comments and edits.


schemas/groups/electrical.json
```diff
@@ -0,0 +1,326 @@
+{
+    "type": "object",
+    "$schema": "http://json-schema.org/draft-03/schema",
+    "id": "https://signalk.github.io/specification/schemas/groups/electrical.json#",
+    "description": "Schema describing the electrical child-object of a Vessel.",
+    "title": "electrical",
+    "properties": {
+
+        "battery": {
```
tkurki: What if there are multiple batteries? I think a sample document, even a simple one, would clear up things a lot as a pure schema is hard for humans (at least for me) to reason about. The sample document can be used to test the schema and vice versa. If you need help with this let me know.

tkurki:
See also #52 as a start for AC schema.

thomasonw:
Restructuring draft greatly - so canceling this pull request.

PR #54, opened 2015-02-07
Pull of restructured ELECTRICAL JSON schema

thomasonw:
Restructured to support multi-instance / multi-source – builds on assumable common ‘source’ definitions with simple extensions applied to each device.


fabdrol:
@thomasonw @tkurki what's the difference with the other electrical pull request? should one be pulled and the other be closed?

thomasonw:
@fabdrol @tkurki :  I thought I had canceled the Pull request "3513b35" dated Jan 25;  on my screens it shows as 'closed' - but perhaps I do not have sufficient access right to "fully" close it.

The correct one to use is the last one,  5a59572 dated Feb 6th.  In it I have attempted to use Definitions as a way to avoid repeating the same information each time  (Example, the dcSource object with volts, amps, etc..) 

-al-

fabdrol:
@thomasonw you're referring to commits, not to PRs. The latter commit you mentioned (5a59572) supersedes 3513b35 because you committed it after that. My question focussed on the two different pull requests about electricity #54 (this page) and https://github.com/SignalK/specification/pull/52 (older)

fabdrol:
Ah my mistake, I thought this was similar to @timmathews AC schema (I didn't read it properly). For now I shall merge with a small change to the group name. After all PRs have been merged I'll raise an issue for everyone to review the master branch, see if new issues arise.

PR #55, opened 2015-03-03
Labels for SignalK data items

tkurki:
SignalK data model is mostly self-documenting, but paths from the model are not really meant for human consumption in a UI.

A UI needs nice, potentionally localizable names. These should be universally available, eg. usable in any SignalK UI and since they need to be in sync with the spec the natural home for them is the specification repository.

This PR is not meant to be merged yet, but an invitation to anybody interested to participate. Please help fill in all the missing data either by sending pull requests, in email or by adding ready to copypaste comments on this thread.

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -185,14 +185,16 @@
     "longName": "Depth in Meters"
   },
   "environment.depth.belowKeel": {
-    "shortName": "Depth Below Keel",
-    "longName": "Depth Below Keel in Meters"
+    "shortName": "DBK",
+    "longName": "Depth Below Keel",
+    "unit": "m"
   },
 
 
   "environment.depth.belowTransducer": {
-    "shortName": "Depth Below Transducer",
-    "longName": "Depth Below Transducer in Meters"
+    "shortName": "DBT",
+    "longName": "Depth Below Transducer",
+    "unit": "m"
   },
   "environment.depth.belowSurface": {
     "shortName": "Depth Below Surface",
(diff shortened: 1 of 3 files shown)


bobmor99:
I'll take a first crack at this.

rob42:
Good idea, but these should all go into .meta{} ?

tkurki:
I have no definite view on where this data should be or be served from. 

If it is in a separate file we can save & eat the cake: a client can package it internally and servers can serve it from whatever path they want.

About .meta - an example would come in handy for this?

rob42:
See https://github.com/SignalK/specification/blob/messageSchemas/samples/signalk-depth-meta-attr.json

fabdrol:
This one is on the roadmap, I assume it hasn't been finished yet? @tkurki

tkurki:
@rob42 About these belonging to the .meta: this data may well be served by the server optionally as part of the full tree, say with a query string option like `?meta?true` or requested specifically by path `/signalk/api/v1/self/environment/depth/belowKeel/meta`, but they need to be mastered someplace.

It would be also very handy for UIs to share common short and long names in a localised format. Usually UI localisation is oftern done with this type of key-value file format.

So my suggestion is to merge this. Just adding this makes it usable for front end components as part of the signalk-schema npm module.

If we want to serve this data merged in the full tree or as a single file under a well defined url that can be easily built on top of this.

rob42:
Lets merge it as its good for early adopters understanding the schema. By doing this we are saying the default langauge is english. We could create a structure later that replaces these with names_XX_xx as in java, eg names_EN_NZ, or names_NL. Then the server looks in locales, finds the right local file, and substitutes names.

PR #58, opened 2015-03-08
Add waypoint.distanceActual, correct units for log & logTrip

zapfware:
added distanceActual for current distance to the next waypoint.
Has to be clarified if Meters or Nautical Miles


tkurki:
@zapfware lower case, please. Could you change log & logtrip to meters as well?

zapfware:
Done.

PR #59, opened 2015-03-12
Added performance group to vessel.json

zapfware:


tkurki:
Hrmph, I should have caught this on the first round.

How about creating a test, say performance.js, that contains a sample document with these fields & validation code like foo.should.be.validSignalK? This way we get a sample document & a runnable test for future reference?

See validation.js for inspiration.

PR #63, opened 2015-04-10
More lenient delta schema

tkurki:
The current delta schema does not explicitly say what a 'value' should be and by itself does not work if we want to validate against extra properties. 

This change specifies value as string/number/object and explicitly allows additionalProperties.

Please notice that I stupidly I created this branch off the one for #62


rob42:
Good to go - merge it

PR #64, opened 2015-04-11
Message schemas

rob42:
Added meta and _attr  objects to definitions.json
Created messages dir, and added subscribe and unsubscribe schemas


tkurki:
@rob42 how about some examples & tests that validate those? We can cooperate, you provide the examples and I can add the tests.

rob42:
samples are added, not sure how you run the tests?

tkurki:
I'll create the tests, but attr and meta would need real usage examples, since they are meant to be used in context, right?

rob42:
Added https://github.com/SignalK/specification/blob/messageSchemas/samples/signalk-depth-meta-attr.json

tkurki:
Sorry about the extra commits, just 5baad1c7418b0a20119d13a14bf2eae0cd660713 is relevant.

fabdrol:
@rob42 should this be merged?

rob42:
Yes, @tkurki - can you pls look at the tests, seems to fail on environment/depth/belowKeel/source (not) being an object - which should be ok i think.
Then pls merge.

fabdrol:
@rob42 @tkurki I'll have a look at the tests, push some fixes and merge when checks pass.

fabdrol:
@rob42 @tkurki I changed the definitions file and to make the tests pass. Could you check if everything is still as intended?

rob42:
I think this is ok now. I put a description of alarms handling in the wiki too. This is how the signalk-jacva-server does it at present

fabdrol:
Okay, then I'll merge. We can review master after all merges have been processed.

PR #67, opened 2015-07-30
Separate group for sails

fabdrol:
This PR is about separating `sails` from the `design` group, as the sails of a vessel are not a design feature.

### DIFF
--- a/schemas/definitions.json
+++ b/schemas/definitions.json
@@ -56,6 +56,54 @@
             "minLength": 8,
             "example": "550E8AB0"
         },
+
+        "sail": {
+            "type": "object",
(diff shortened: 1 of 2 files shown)


fabdrol:
@tkurki @timmathews @rob42 @canboat feedback?

rob42:
I think we should merge this and consider it experimental until we have some experience using it. It all quite theoretical till then. eg I can see sail dimensions (luff/foot/leach/curvature etc) being useful for maintenance and  performance stuff, but its not needed yet, so we can alter later.

fabdrol:
I agree, let's do that. It's not hard to change things later :)

PR #72, opened 2015-08-17
Update tanks.json

MariusVolkhart:
This is a breaking change.

The addition of the `greywater` tank type and the change from `water` to `fresh water` is motivated by the fact that some boats have rainwater collection systems. If I recall correctly, greywater is also available in the N2K spec.

The addition of the `name` field is explained inline.

The change from `tankType` to `type` reflects the naming convention used in other parts of the spec.

The other changes are just to standardize the formatting and provide reader clarity.

### DIFF
--- a/schemas/groups/tanks.json
+++ b/schemas/groups/tanks.json
@@ -2,29 +2,33 @@
     "type": "object",
     "$schema": "http://json-schema.org/draft-03/schema",
     "id": "https://signalk.github.io/specification/schemas/groups/tanks.json#",
-    "description": "a tank, named by a unique name within this vessel",
+    "description": "A tank, named by an UUID",
     "title": "tank",
     "properties": {
-            "tankType": {
-                "type": "string",
-                "description": "The type of waypoint",
-                "enum": [
-                    "petrol",
-                    "water",
-                    "holding",
-                    "lpg",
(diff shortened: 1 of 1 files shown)


fabdrol:
Not sure about "named by UUID" (`29ad443a-25be-4c12-9b8e-e3ab4330f408` doesn't look very friendly to me). I would propose to add an ID field of some kind (maybe good to add to everything that indicated a physical "thing" on the boat, like in the design group (one or more masts, etc), the sensor group, etc) and keep name as a label for the user

MariusVolkhart:
@fabdrol agreed, the "named by UUID" is incorrect. "Fresh Water 1" and "Fresh Water 2" serve as appropriate unique keys already. Thanks, I'll fix that up.

After realizing the "item 1", "item 2" scheme take care of uniqueness, I'm actually inclined to say skip the ID field.

fabdrol:
@MariusVolkhart could you change the description, so we can merge this PR? Thanks!

fabdrol:
@MariusVolkhart did you see my last comment? Could you update your branch and commit? Thanks!

fabdrol:
@MariusVolkhart as you never updated your fork, I've copied your changes into another PR and I'm closing this one.

PR #75, opened 2015-08-17
Format all enums to the same style

### LINKED ISSUE #73: Enum formatting
Enums have various formats throughout the spec. For example, `waypoint.type` is all lower case while `navigation.state.value` starts with a upper case letter. What is the casing style that is preferred? I'd like to submit a PR with the appropriate changes.

MariusVolkhart:
Lowercase with spaces for multiple words

Fixes #73


fabdrol:
looks good, doesn't involve much change so I'll go ahead and merge.

PR #77, opened 2015-08-18
Format the resources group

MariusVolkhart:
Format the descriptions in the `resources` group


fabdrol:

PR #78, opened 2015-08-18
Format the propulsion group

MariusVolkhart:
Format the descriptions in the `propulsion` group.

This introduces a breaking change in that the engine type enum has a spelling change.


fabdrol:

PR #80, opened 2015-08-18
Format the navigation group

MariusVolkhart:
Format the descriptions in the navigation group.

### DIFF
--- a/schemas/groups/navigation.json
+++ b/schemas/groups/navigation.json
@@ -78,7 +78,7 @@
                         },
                         "xte": {
                             "type": "number",
-                            "description": "Cross track error in METERS"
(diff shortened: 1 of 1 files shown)


fabdrol:
These descriptions can be better, we just put in the unit (prob. temporary change). I propose to put full sentences in there, like we have in the other groups. Furthermore, the longitude and latitude fields are in "decimal notation", that should be "decimal degrees"

MariusVolkhart:
@fabdrol I gave this one another go. Thanks for the feedback :)

timmathews:
Hmm. Not sure we should have merged. Units ought to be in the `units` field, no need to be in the description

Issue #71, opened 2015-08-17
Vessel without MMSI


MariusVolkhart:
In the root document, a vessel can be identified by its MMSI or an UUID. The `vessel.json` specifies only a `MMSI` field. In the event that a vessel is identified by an UUID, what is the correct value of the MMSI field in `vessel.json`: `null`, empty string (`""`), or the UUID?

fabdrol:
@MariusVolkhart In that case there should be an `uuid` field on `vessel`. It should have been in there, thanks for spotting it. I'll update the document asap.

tkurki:
@fabdrol doc change is not quite enough, requires a schema change. How does json schema model "either of these properties"?

fabdrol:
@tkurki "document" == `vessel.json` ;). I'm not sure you can do "required unless", but I'll check it out.

fabdrol:
See #83

fabdrol:
@tkurki I don't think "required unless" is an option, let's just solve that in documentation?

tkurki:
Isn't this really `either MMSI or UUID`? 

See 
- http://json-schema.org/example2.html nfs example
- https://github.com/json-schema/json-schema/wiki/anyOf,-allOf,-oneOf,-not

MariusVolkhart:
`oneOf` is new in v4 of json-schema :(

fabdrol:
Yes but that all relates to the value, not the key. An option would be to use `oneOf` with a `RegExp` key

timmathews:
No need for "required unless". UUID is *always* required and valid as a hash key or the value of `self`. If the optional MMSI exists it is also valid as a value of `self`.

tkurki:
Sorry about my casual early comments that added just confusion.

So this is how it is in master:
- `root.self`is a required `string`
- `root.vessels.<id>.mmsi` is optional

@fabdrol's change in #83 
- same as in master plus
- `root.vessels.<id>.mmsi` is optional

So in effect the only thing that is required in addition to self is the `<id>` key under vessels. If the key is mmsi then it should appear also as  `root.vessels.<id>.mmsi`. In case it is uuid you can tell that it is uuid (eg. some generated key) by the fact that there is no mmsi matching the `<id>` key.

I would go with this and cancel the PR.

fabdrol:
Reason I like the `UUID` on vessel is for this use case:
- I subscribe to local boat
[...]

PR #83, opened 2015-08-19
Added an UUID property to vessel.json

fabdrol:
Fix for #71

### DIFF
--- a/schemas/vessel.json
+++ b/schemas/vessel.json
@@ -6,8 +6,12 @@
     "title": "vessel",
     "properties": {
         "mmsi": {
-            "description": "MMSI number",
+            "description": "MMSI number of the vessel, if available.",
             "$ref": "definitions.json#/definitions/mmsi"
+        },
+        "uuid": {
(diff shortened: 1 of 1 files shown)


tkurki:
Should we make either UUID or MMSI mandatory under vessels.XXX? It is perfectly possible with JSON Schema.

```
diff --git a/schemas/vessel.json b/schemas/vessel.json
index fc6ab2b..c096b73 100644
--- a/schemas/vessel.json
+++ b/schemas/vessel.json
@@ -103,5 +103,9 @@
             "description": "Performance data",
             "$ref": "groups/performance.json#"
         }
-    }
+    },
+    "oneOf": [
+      {"required": ["mmsi"]},
+      {"required": ["uuid"]}
+    ]
 }
```

sumps:
Yes I think UUID has to be mandatory and Name as well, everything else is recommended but not mandatory. Reason for Name being mandatory, is that we need some human readable label when visualising or using the data.
MMSI, Call Sign and IMO number are the other marine identifiers which can be recommended not mandatory.

Issue #91, opened 2015-09-28
Vessel Identity


timmathews:
Intro
---
The `vessels` object at the root of the Signal K object is a key-value store holding information for 1 or more boats. We need to determine exactly what this key should be and what requirements it must fulfill.

This issue is intended to discuss that. Please keep in mind that this is not the same as the identifier for other Signal K hosts on a network (e.g. on a boat with multiple Signal K sources). **This is for identifying a single vessel which may be a collection of one or more Signal K devices**.

Background
---
We were originally going to use MMSIs as the unique identifier, but that won't work for a variety of reasons, not the least of which is not everyone has one. MAC address of the SK server device was floated as an option, but that changes whenever the hardware does, so it may not be a good option either. In fact, a boat may have more than one SK gateway and we need to keep the ID the same across devices.

Requirements
---
- Globally unique
- Easily reproducible
- Portable across devices
- No central agency required to assign them
- Do all of this and protect end users' privacy

These are what I believe the requirements to be. I would like to use this discussion to firstly agree on a set of requirements, vetted against as many use cases as possible, and then float options which satisfy these requirements so that we can agree on something.

Use Cases
---
- Route sharing
- Fleet management (races, cruising rallies, etc.)
- Crowd sourced marine survey data
- Automated log keeping

tkurki:
I believe that a single identification scheme won't cover all situations, unless it is something like a generated UUID, which sort of doesn't make sense - nice APIs, non-human-friendly IDs and too brittle.

Here we run into limitations of the JSON approach: there is no way to build a composite key (scheme + value) without just stuffing the values into a string key.

On the other hand if we think about this in terms of an API and not a single JSON document we could define different routes / paths such as
- `/signalk/v1/api/byMMSI/<mmsi>`
- `/signalk/v1/api/byUUID/<uuid>` ( per the real RFC 4122)
- `/signalk/v1/api/byIMO/<IMO number>`
- `/signalk/v1/api/byNational/fin/<finnish vessel registration id>`
- `/signalk/v1/api/bySailId/<sailId>`
- `/signalk/v1/api/byName/<vesselname>`

Each could offer a different view (in Signal K format) to the data the server has. The paths above illustrate just some of the ways to identify a Signal K entity.

Without the trailing id parameter you would get a collection of all the vessels' (items')  data keyed by the ids in JSON safe format.

`/signalk/v1/api/` could list all the different access schemes that the server supports.


[...]

PR #95, opened 2015-10-13
Uuid and mmsi as main vessel identifiers

tkurki:
Changes:
- either vessels.`<id>`.mmsi or vessels.`<id>`.uuid mandatory
- vessels.`<id>`.uuid consist of
 - UUID in canonical format and
 - checksum calculated with CRC-32
- the id under vessels to be either `signalk:<uuid>` or `mmsi:<mmsi>`

See #91 for more discussion


fabdrol:
+1

PR #98, opened 2015-10-17
Identities

tkurki:
Schema implementation & tests for three methods to specify vessel identity. See #91

### DIFF
--- a/index.js
+++ b/index.js
@@ -75,6 +75,10 @@ function validateFull(tree) {
   }
 
   var valid = tv4.validateMultiple(tree, signalkSchema, true, true);
+  var result = tv4.validateResult(tree, signalkSchema, true, true);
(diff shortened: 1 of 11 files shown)


fabdrol:
@tkurki there is an issue with the URNs you've used, if we want to follow that newspaper. Currently you use `urn:uuid:<uuid>` and `urn:mmsi:<mmsi>`, but as I understand it we need to follow the namespace, e.g. `urn:mrn:signalk:uuid:<uuid>` and `urn:mrn:imo:mmsi:<mmsi>`

timmathews:
I agree, let's update it to match the MRN proposal and then merge.

fabdrol:
Superseded by https://github.com/SignalK/specification/pull/110

PR #102, opened 2015-11-09
Add NMEA1038 talker and sentence source fields

tkurki:


fabdrol:
Two points:
- I find "src" unclear, could we use something that's more descriptive and not N2K-specific
- I propose to change `talker` to `source`, which can then later be a reference to a "source" meta data object.

sumps:
I see no reason to not use industry standard expressions where they exist otherwise the level of documentation explaining these new SK versions has to increase. For someone use to NMEA expressions there is immediate recognition and for someone not use to NMEA, there is a lot of info already out there and it helps to get these NMEA expressions in to the open world.

tkurki:
My thinking:especially delta should capture all the relevant details of the incoming message. For NMEA0183 that is
- talker id  eg. "The next five characters identify the talker (two characters)"
- sentence
- data in Signal K data model
For N2K
- priority (not included right now btw)
- src
- pgn
- data in Signal K data model


Remember that delta and full share the same `source` definition. In full the source could/should be a reference to the /vessel/X/sources entry, but in delta the message should be self contained and carry all the relevant information.

So I'd go with source specific field structure - the fields are optional, eg. NMEA0183 originated messages will carry talker and sentence and N2K src and pgn (and probably priority as well). 

For data originating from non-NMEA sources, for example sensors directly connected to the server, I think label should be enough for now.

Somebody better versed with Seatalk should add Seatalk specific entries. Ping @rob42 .

timmathews:
I think the only reason `src` is `src` and not `source` is because that's the abbreviation @canboat used.

But that's sort of irrelevant, because we can simplify replace the entire `source` object with its `label` property. If the client needs additional info (for diagnostics or debugging), then it should make an HTTP request to the server asking for the details of the particular source: `GET /signalk/v1/api/sources/<src-foo>`.

The source device is the only variable. The other stuff - like sentence and PGN never changes - those are based on the static mapping of PGNs and sentences to SK path. In the case where a device transmits multiple PGNs or sentences with the same data, we should only use one of them. Whether the user chooses which to use or we rank them and choose the best is an implementation detail. Regardless, it any given system, for a particular device it will always be the same message used to populate that data point.

With the current design, we end up sending _a lot_ of redundant data. If we can remove some bytes here, we should do it.

tkurki:
If we replace most of the data in `source` with something similar to the `$source` reference in  https://gist.github.com/tkurki/84cd71d2329f3ffae9f0 we are encoding label plus N2K source / NMEA0183 talker  in a single field. Why do we argue that json verbosity is ok in some places and not ok here?

I think that the output delta stream should carry all the data that the input stream carries. 

If a consumer wants to have access to all updates regardless of the any source preference related logic in the server we should send all input data fields as delta out. We should include pgn/sentence in it, otherwise a consumer has no access to this information, as it can not be divined from just a source reference.

So in the end the only datum that we could do without is the `type` field. It is redundant for NMEA data if source+pgn/talker+sentence are available. If it will be just something like 'other' for all non-NMEA data it doesn't carry much information.

timmathews:
> Why do we argue that json verbosity is ok in some places and not ok here?

Being verbose (no abbreviations, strings instead of ID numbers in enums, that sort of thing) is OK. This is being positively garrulous. There just isn't any value in wasting bandwidth to repeatedly send static data.

> We should include pgn/sentence in it, otherwise a consumer has no access to this information, as it can not be divined from just a source reference.

Why? If the client wants to know about the specific pgn/sentence that its data came from, it can make an HTTP request to get that info. Then it can cache it and never have to ask again.

>  [Type] it will be just something like 'other' for all non-NMEA data it doesn't carry much information.

Why would it be `other`? Seems to me that it would be the type of network that the data came from. That could be `canbus` for generic CAN, `j1939` for engine stuff, or `seatalk`, `fastnet`, `modbus`, `fieldbus`, `profinet`, `devicenet`, `lonworks`, or any other transport you can think of.

tkurki:
>>we are encoding label plus N2K source / NMEA0183 talker in a single field.
>There just isn't any value in wasting bandwidth to repeatedly send static data.

My point was that if we change `..."label":"n2k1","source":"115"...` to `..."$ref":"n2k1.115"...` we are encoding that data into a single field and do not have it readily available in json fields. So this is not about resending static data. If we follow that path why stop here, because the whole delta could be encoded more efficiently in one json field...

>> We should include pgn/sentence in it, otherwise a consumer has no access to this information, as it can not be divined from just a source reference.

>Why? If the client wants to know about the specific pgn/sentence that its data came from, it can make an HTTP request to get that info.

It can not if a sensor sends out multiple pgns that contain the same datum.

>>  [Type] it will be just something like 'other' for all non-NMEA data it doesn't carry much information.

>Why would it be `other`? Seems to me that it would be the type of network that the data came from. That could be `canbus` for generic CAN, `j1939` for engine stuff, or `seatalk`, `fastnet`, `modbus`, `fieldbus`, `profinet`, `devicenet`, `lonworks`, or any other transport you can think of.

You're right, can't understand what I was thinking there.

tkurki:
@timmathews I think that how source information is encoded in delta is separate from this issue. If you think that we should change that I think it should be a separate issue - this PR is in line with what we have imho. 

Can we merge this to keep WIP down?

fabdrol:
Let's merge for now. I'm in favour of moving to a structure like we are talking about in multiple sources, where we have a key with a certain pattern that references a source under `/vessels/self/sources/<source>` or `/sources/<source>`  (IMO that is more inline with various other things in the spec, like the vessels map (`vessels/<urn>`), the `propulsion` map, etc).

timmathews:
@tkurki, you're right. I'm fine with the contents of `source` (actually there are a few more fields that I think could be useful, but they're low priority), I just don't think the whole thing should be in the delta every time. We can discuss that elsewhere.

fabdrol:
This one is superseded by https://github.com/SignalK/specification/pull/109

PR #105, opened 2015-11-14
Housekeeping: cleanup of all schemas

fabdrol:
Major cleanup of all existing schemas:
- Moved "units" from descriptions to a separate key (to improve documentation)
- Cleaned up indentation differences
- Small changes to various descriptions (mostly related to consistency)
- Fixes in various schemas where they didn't comply with JSONSchema standard or contained JSON parsing errors. 
- Includes changes to "tanks" suggested by @MariusVolkhart

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,5 +32,9 @@
     "chai": "1.9",
     "mocha": "^2.1.0",
     "tv4": "latest"
+  },
+  "devDependencies": {
+    "chai": "^1.9.2",
+    "tv4": "^1.2.7"
   }
 }
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+# editorconfig.org
+root = true
+
+[*]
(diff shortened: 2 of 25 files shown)


package.json
```diff
@@ -32,5 +32,9 @@
     "chai": "1.9",
     "mocha": "^2.1.0",
     "tv4": "latest"
+  },
+  "devDependencies": {
+    "chai": "^1.9.2",
+    "tv4": "^1.2.7"
```
tkurki: Chai & mocha should be dev, tv4 a real dependency.

fabdrol: Okay, will update

rob42:
Basically all good, except the units key. The current `units` array should be in the `description` field as its the human readable version, and the `units` array should be just the unit code eg `°/s`, which is the machine readable version

fabdrol:
Sounds good, will update today.

PR #110, opened 2015-11-20
Identities, take two

fabdrol:
In this PR I re-did @tkurki's work on the cleaned-up version of the schema.

Another thing we should discuss before merging: the keys in the `vessels` array can be any valid MRN. If the MRN is a MMSI, the respective vessel has a `mmsi` key with the 9-digit number in it - without the rest of the MRN (authority, etc). 
For UUID, the schema currently demands the whole URN/MRN in the `uuid` key of the vessel. This is inconsistent with the way we treat MMSI's, however I think it does make sense from the point of view of the UUID, as it doesn't tell much on it's own and, if data comes in from other sources that use UUID but that are not SK, they can fit here as well if it includes the namespace. 

Another thing: current regexp pattern allows URLs as a valid ID for a boat... which doesn't play well with JSON (slashes). We could either disallow this and force everything to use URNs, or leave this in for now for flexibility sake..

What do you guys think?


tkurki:
Fixed, ready to merge by me.

rob42:
nice, merge!

Issue #112, opened 2015-11-22
References to units should be in ASCII only


canboat:
When a unit has a Unicode string (for instance \u00b0 for degrees) label this ends up in various places in the model, which might not be readable in all editors.

The 'units' description in `definitions.json` should contain a map `"label": { "display": "\uxxxx" .... }`

jboynes:
The display text for these units in some places uses characters from the CJK compatibility block that are discouraged (e.g. "\u33ad" for "rad") even though they do semantically represent the unit. This is not consistent though with some using canonical forms even though there are characters to represent them (e.g. "\u33ad/s" rather than "\u33ae" for "rad/s" or "Hz" rather than "\u3390" for frequency).

How about moving the display information from `definitions.json` to `i18n/en.json` where the display form and description could be localized. This could then include localization of units as well, for example displaying speed in knots. However, localization would also add the problem where different units are used depending on application (for example, boat speed in knots vs. line speed in ft/min, or fuel consumption in gph vs. air flow in cu. ft./min).

This could be addressed by using a semantic reference in the metadata rather than a raw unit. For example, storing `unitType` in `meta` rather than units, something like:

```json
"vessel": {
  "speedThroughWater": {
    "value": 1.0,
    "meta" : {
      "unitType": "speed.boat"
    }
}
```
In `en_gb.json`:
```json
"unitTypes" : {
  "speed.boat" : {
    "unit": "m/s",
    "display": "m/s",
    "ratio": 1.0
  }
}
```
In `en_us.json`:
```json
"unitTypes" : {
  "speed.boat" : {
    "unit": "m/s",
    "display": "kts",
    "ratio": 1.94384
  }
}
```

tkurki:
The specification json schema files are the master for unit information. Did you notice https://github.com/SignalK/specification/blob/master/keyswithmetadata.json ? That is generated from json schema.

I'm not 100% convinced that the extra indirection layer for specifying the display units is worthwhile. In the end the display units are up to the Signal K consuming application.

canboat:
If we do start catering for localization we should also cater for locales that use more than one presentation for a particular value. Nautical distance is expressed differently in some locales depending on the distance.

For example
1 NM = 1 NM (UK) = 1 NM (NL)
but
0.1 NM = 1 Cb (UK) = 182 m (NL)

I guess this is why my Navico chartplotter has a "small distances" and "large distances" unit selection mechanism.

[...]

PR #113, opened 2015-11-22
Implement my own suggestion in issue #112

canboat:
Here's my suggested patch.


timmathews:
Temperature values should be in Kelvins (K), not degrees Celsius (C).

Issue #30, opened 2014-11-06
Use (derived) SI units for all values


tkurki:
"For the most part, we're using SI units for values (meters, seconds, etc.), but angles are represented as degrees in the spec. I think we ought to change this to radians.

[23:39] tim: n2k uses radians and most math library functions which operate on angles also use radians. So right now, we get an angle over n2k in radians, convert it to degrees, then if any math has to be done, we convert back to radians, do the math, then convert back to degrees again.

[23:39] tim: Wouldn't it be simpler and more consistent if we specified radians on the wire and only convert to degrees at the display?"

canboat:
CANboat `analyzer` now supports emit of angles and rotations in rad and rad/s, using the new `-rad` option. See https://github.com/canboat/canboat/commit/91d0582cbe526b1ee560cd2fa311e8c9d4ccde2c

rob42:
Closing as this is agreed and almost done now, and there are no commits against this issue.

PR #114, opened 2015-11-23
Change all units to (derived) SI units (fix issue #30)

canboat:
This should be everything.

The list of used units at the moment is:

```"A"
"C"
"Hz"
"J"
"K"
"Pa"
"V"
"W"
"deg"
"kg"
"m"
"m/s"
"m2"
"m3"
"m3/s"
"rad"
"rad/s"
"ratio"
"s"
```


fabdrol:
+1 for merging

canboat:
I've also pushed a change to canboat so analyzer supports `-si` option. It didn't need much conversion though except for Kelvin, as most of the weirder stuff is not yet supported by the PGN list (such as battery capacity.)

PR #115, opened 2015-11-27
Update propulsion.json

sumps:
Added all missing values mapped from N2K

Not sure about Percentage units or seconds for Engine Hours.

First Pull Request and editing of JSON so please check carefully !!

### DIFF
--- a/schemas/groups/propulsion.json
+++ b/schemas/groups/propulsion.json
@@ -42,7 +42,7 @@
           "enum": [
             "stopped",
             "forward",
-            "neutral"
+            "neutral",
             "reverse"
           ]
         },
--- a/schemas/groups/propulsion.json
+++ b/schemas/groups/propulsion.json
@@ -131,22 +131,10 @@
       "units": "ratio"
     },
     
-     "engineTorque": {
+    "engineTorque": {
       "description": "Engine torque in Percent",
       "$ref": "../definitions.json#/definitions/numberValue",
       "units": "ratio"
-    },
-    
-     "transmissionOilPressure": {
-      "description": "Transmission oil pressure in kPa",
-      "$ref": "../definitions.json#/definitions/numberValue",
-      "units": "kPa"
-    },
-    
-    "transmissionOilTemp": {
-      "description": "Transmission oil temperature in degrees C",
-      "$ref": "../definitions.json#/definitions/numberValue",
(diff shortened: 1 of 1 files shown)


sumps:
I have corrected the typo, but I am struggling to restructure the Propulsion Group so that it has Engine and Transmission object groups. Basically it needs to be....
Propulsion
    -  Engine Port
          -  properties....
    -  Engine Starboard
          -  properties....
    -  Transmission Port
          -  properties....
    -  Transmission Starboard
          -  properties....

The example above is a larger vessel with multiple engines. 

Could someone tell me how to restructure the Propulsion Group in this way and I will complete this PR ?

canboat:
Hm, you can't assume that there are 1 or 2 engines. Even assuming there are
as many transmissions as there are engines might not be true in all cases
(one or more may not have a NMEA 2000 interface connection.) In other words
it needs to cater to an arbitrary number of both, with 1 or 2 being the
most common.

2015-11-30 12:31 GMT+01:00 sumps <notifications@github.com>:

sumps:
Absolutely agree. The number of engines and transmissions will be decided by the gateway based on the data it is receiving. So in Propulsion there may be one Engine, one engine plus a transmission, two engines, two engines plus two transmissions, n engines, n engines plus n transmissions or I guess even some transmissions and no engines but I think this is unlikely, unless there is a problem with the J1939 interface or engine sensors

This comes down to the Multiple and Duplicate sources again

So there are Two Engines and they are multiple sources that must be treated equally and there are two GPS units and they are duplicate sources where one must be given preference over the other (in most situations)

Multiple Sources (Engines, transmissions, batteries, etc.) where all are equally important and all must be displayed, versus duplicate sources (GPS, depth, etc.) where you only want to know the preferred (best) one.

Propulsion, Tanks, Electrical and Sources (currently Sensors) are all groups which deal with multiple sources and need a "map" type schema

Not sure about the Alarm Group which is currently marked as a "map" but that is another important subject that we will need to look at soon.

sumps:
OK - have removed the transmissionOilTemp and transmissionOilPressure fields that I had previously added as we can use the existing oilTemerature and oilPressure fields used for an Engine. So now the Transmission data will be mapped to ...propulsion/transmission123/oilTemperature and ....propulsion/transmission123/oilPressure

PR #116, opened 2015-11-27
Add Percentage unit of measure

sumps:
To simplify conversion of values, it would be really useful if SK supported the storage of Percentages rather than just ratios.


canboat:
Sorry, but we have decided that the model will only store SI or unitary values. Since we already have 'ratio' why would you want percentage as well? Compare this to not having 'hPa' (very convenient when expressing air pressure that is naturally expressed in mbar, and 1 mbar = 1 kPa) or 'h' or 'l'.

sumps:
Some data sources use percentages (+/- 124) so it would make it easier for the gateway to just grab the integer data and pop it in to SK with no conversion.

PR #118, opened 2015-11-29
Multiple values (take 2)

tkurki:
Supercedes #49. See discussion at #48.

Node implementation at https://github.com/SignalK/signalk-server-node/commits/multiple-values.


rob42:
merged into v1-beta1

tkurki:
I think this is in pretty good shape and worth merging, even if there are some loose ends, the motivation being that it is easier to start using this for validation from master.

The multiple values implementation needs a human readable piece of documentation. I wish somebody else would try this out with fresh eyes & brains. The node server branch https://github.com/SignalK/signalk-server-node/tree/multiple-values has `bin/multiple-sources` that starts the server with four file readers, two each for `aava-n2k.data` and `plaka.log`. An example of the resulting full tree: https://gist.github.com/tkurki/f7a4b3b316f7f65eae1f

The main differences to master are
* source is now a reference, `$source`
* the value holds the timestamp and pgn/sentence producing the value
* there is not a top level `sources` object with a hierachy describing the different input sources
  * the first level is a human-readable label/id for the input source
  * second is the n2k device / nmea0183 talker
* if there are multiple values they are under  `values` object, keyed by their source; "preferred" value stays where it has been previously

Node-server has "last one wins" policy for the preferred value.

Reasoning for having the timestamp and pgn/sentence in the value is that a single pgn/sentence produces multiple values in different parts of the tree.

Sources also hold timestamps of the last received updates per pgn/sentence.

I think that `definitions,json` should be cleaned up from stuff that is not used anywhere, such as nullValue. 

I changed some of the object-valued properties to use the shared definition of common values in 64add9a0eeff0736f6b677c1d1d66108034edc2b. This is not complete - I only changed stuff that was being exercised by n2k-signalk tests. There is probably more on the schema to update.

@fabdrol I rewrote multiplexing logic as part of the multiple values server implementation. I don't think it is really needed as a sepate module. The JS validation code in `specification` needs a way to map deltas to full tree to allow validation for the data item paths, so that would be the place to put it, as all the other modules depend on it at least for validation in devDependencies.

schemas/definitions.json
```diff
@@ -229,8 +234,7 @@
 
     "url": {
       "type": "string",
-      "description": "A valid URL.",
-      "pattern": "^(http(s?):.*|mailto:.*|tel:(\\+?)[0-9]{4,})$"
+      "description": "A location of a resource, potentially relative. For hierarchical schemes (like http), applications must resolve relative URIs (e.g. './v1/api/'). Implementations should support the following schemes: http:, https:, mailto:, tel:, and ws:."
```
fabdrol: @tkurki can you clarify the relative stuff? Is that for resources hosted by the server itself (and can non-hosted or resources on another server be non-relative?) Would we store the location of a USB device here as well? E.g. `file:///dev/tty.USB0`?

fabdrol:
From reading the changes I'm not sure how multiple values look in JSON.. Is it like this (quoting @tkurki)?

@tkurki what needs to be done still?

```
"sources": {
    "nmea1": { //"NMEA0183 input 1 of this server"
        "label": "nmea1",
        "type": "NMEA0183",
        "II": {
            "talker": "II",
            "sentences": {
                "VHW": "2015-11-15T14:52:01.092Z",
                "VTG": "2015-11-15T14:52:01.192Z",
                ....
            }
        }
    },
    "nmea2": { //NMEA0183 input 2 of this server
       ..... 
    },
    "n2k1": { //"NMEA2000 bus 1 connected to this server"
        "115": {
            "src": "115",
            "pgns": {
                "128259": "2014-08-15-19:00:01.086",
                "128267": "2014-08-15-19:00:00.591",
                ...
            }
        },
```

schemas/definitions.json
tkurki: Not related to multiple values.

fabdrol: Ah, my mistake. I was fooled by GitHub inserting a crumble. I noticed after expanding it :|

tkurki:
@fabdrol did you check the gist above?

fabdrol:
Ah, missed that. Going too fast I suppose. Looks good to me, except maybe one thing: the `source` property has a `$` in front (`$source`). Is that because it's a reference?

fabdrol:
@tkurki how are we looking on this/what needs doing before we can merge? I think this point has been discussed over and over, and maybe we should move on beyond it.

tkurki:
I'll continue with
- [ ] clean up the definitions so that all values either support multiple values like `numberValue` or are removed
- [ ] add testclient support

Ping @rob42 how about java server support?

And `source` has `$` to indicate that it is a reference - somebody else indicated this mechanism during the discussion.

fabdrol:
@tkurki ah, makes sense. I just pushed a PR which moves `string` types to `stringValue` in definitions, make sure you ready that one too for multiple values. I'm working on an update to the lite 0183 parser - should I take anything into account?

tkurki:
I think this is about ready to be merged. *Value cleanup for unused values in definitions remains undone.

Issue #96, opened 2015-10-15
How will Signal K handle Alarms/Alerts ?


sumps:
It would be very good if a Signal K gateway/server could pass through NMEA2000 Alerts, AIS MOBs, etc. or there was a mechanism for a gateway/server to trigger an alert/alarm. One way would be to create a new object in the Root called Alarms;
- Alarms
    - aismob#12
    - alert#367
    - alarm#490

Each Alarm would reference the Signal K object that the alarm was related to i.e. a sensor, engine, aisSART, etc. so that it would be easy to access the relevant properties of the alarm by calling the object....  signalk/api/v1/vessels/self/propulsion/transmissionOilPressure/value

Obviously there are a lot of things to consider such as resetting/timing out the alarm and it maybe out of scope for V1 but we need to start considering this issue.

sumps:
OK I have found Alarms in the Vessel object, not sure how I missed that before, and I will start thinking how this can be mapped to NMEA2000 alerts. 

The gateway will need to have a setting for alarms allowing some of the fields to be set like notification method, enabled/disable, etc.

I'll include my ideas in this issue.

rob42:
See https://github.com/SignalK/specification/wiki/Alarm-Handling for a description of alarm use. However the method described assumes there is a signalk key to set the alarm against. While most alarms are related to some sort of measurement and therefore have signalk keys, there are some that dont have obvious homes. I'm thinking about MOB, fire, etc. Maybe post a list of any that surface?

rob42:
@sumps Can you list a few NMEA alarms we need to deal with, other than those that will be handled by the current 'check against limits' scheme. MOB comes to mind - we dont want to compare 'crew' with limits in `.meta` and raise alarms if we loose any:-)

rob42:
Some more thoughts on this. The `vessels.uuid.alarms.` tree is assumed to be populated by the server or an alarms daemon from the rest of the signalk tree, by comparing with the values bounds in `.meta`

Then the alarms that exist there are notified to screens etc by an alarm monitoring process. So plugging into this would be a good thing.

There is no reason that a value cannot be simply posted directly in there, eg `vessels.uuid.alarms.mob`. All the better if they map back to a signalk key, eg engine fire would be `vessels.uuid.alarms.propulsion.engine1.fire` - which would be more useful than a simple fire alarm, as you immediately know where to look.
Will need to look at `alarmState`, probably needs another state `emergency`

rob42:
[...]

Issue #86, opened 2015-08-28
Boat Brand and Model, plus standard "Self" Settings File


sumps:
There are a large number of available settings within SK and although I am sceptical that your average user would set all of these settings, I think we need to have a Boat Brand field, Model field and YearBuilt field so that standard boat setup files can be created.

For instance a Jeaneau Sun Odyssey 41DS will have a set of dimensions and polars that are the same for every boat and by storing the Boat Brand, Model and YearBuilt data we could encourage the collection and publishing on the SK site of the standard settings file for common boats. This would make it easier for customers as once one customer has gone to the trouble of entering all of the data, it will be available for others to use.

Thinking about this, it would be a good idea to define a standard "Self" settings file that all consumers/servers/gateways used to store this data. The file would have to have a Version Number that was linked to the SK specification version but it could be a useful and social way for SK users to share boat data.

tkurki:
It would be really cool to a have global shared & shareable (Creative Commons license or similar) repository of yacht technical data, addressable by url, say

http://yachts.registry.signalk.org/Jeanneau/SunOdyssey/41DS

fabdrol:
It would definitely make setting this up easier; after installing a sk device you simply connect to the internet, search for your boat, click, done. If it's not there, a user could choose to add it which means crowdsourced data..

rob42:
This is an idea I have had for Freeboard too. We could provide a service that has templates based on typical boat types, which does basic config for you, eg 11M sailing yacht. Then subclasses of increasing specialization. So eventaully Jeanneau/SunOdyssey/41DS. But more than just config you could collect and compare reliability, polars, comparisons, tracks, and social meets, swaps etc.

webmasterkai:
I'm interested in building out this resource. Perhaps something in partnership with sailboatdata.com? I posted a message on the forum asking about creating and API for the website http://sailboatdata.com/bbs/index.php?topic=2236.0

It would be trivial to extract the fields in the SignalK schema and save using the appropriate field ids. Initially read-only but eventually accepting info as @rob42 suggests.

sumps:
I think this is a great idea but it is important that the data is freely available to all Signal K users and developers. This is the first time that I have heard of sailboat data.com but it looks like it is a site setup by yachtsmen for yachtsmen so perhaps it would be a good partner.

sumps:
i am sure the core Signal K team would welcome any work you could do on this, have you joined the Signal K Slack channel ?

[...]

PR #119, opened 2015-11-30
Rework electrical dc: batteries

tkurki:
DC schema had a bunch of reference & structural errors and could not handle more than one battery.

This pull request fixes batteries, but inverter part still needs work.

Test includes a sample document with multiple batteries.

### DIFF
--- a/schemas/groups/electrical_dc.json
+++ b/schemas/groups/electrical_dc.json
@@ -202,16 +202,15 @@
           "description": "Batteries, one or many, within the vessel",
 
           "allOf": [{
-            "$ref": "#/definitions/identity",
+            "$ref": "#/definitions/identity"
+          }, {
             "$ref": "#/definitions/dcSource"
           }, {
             "properties": {
               "mode": {
                 "type": "object",
-                "description": "Operational mode of battery",
                 "properties": {
                   "value": {
-                    "type": "string",
                     "enum": [
                       "charging bulk",
                       "charging acceptance",
--- a/schemas/groups/electrical_dc.json
+++ b/schemas/groups/electrical_dc.json
@@ -4,329 +4,324 @@
   "description": "Schema describing the electrical child-object of a Vessel.",
   "title": "DC Electrical Properties",
   "type": "object",
-  "properties": {
-    "definitions": {
-      "title": "definitions",
+  "definitions": {
+    "identity": {
       "type": "object",
-      "description": "DC related definitions",
-
+      "title": "Electrical ID",
+      "description": " Common ID items shared by electrical items",
+      "required": ["name"],
       "properties": {
-        "identity": {
+        "name": {
+          "type": "string",
+          "description": "Unique ID of device (houseBattery, alternator, Generator, solar1, inverterCharger, Combiner1, etc.)"
+        },
+
+        "location": {
+          "type": "string",
+          "description": "Installed location of device on vessel"
+        },
+
+        "manufacturerName": {
+          "type": "string",
+          "description": "Manufacturer's name"
+        },
+
+        "manufacturerModel": {
+          "type": "string",
+          "description": "Model or part number"
+        },
+
+        "dateInstalled": {
+          "$ref": "../definitions.json#/definitions/timestamp",
+          "description": "Date device was installed"
+        },
+
+        "timestamp": {
+          "$ref": "../definitions.json#/definitions/timestamp"
+        },
+
+        "source": {
+          "$ref": "../definitions.json#/definitions/source"
+        }
+      }
+    },
+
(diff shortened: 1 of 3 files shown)


sumps:
You are making good progress with DC but if we are going to have lots of manual settings, which I think is correct, I think it is really important that we include these in a separate section of the scheme so that they can be saved, shared and published for other owners of similar boats to use, as mentioned in issue #86

sumps:
We have....
- Alarm values
- Visualisation properties i.e. Max RPM
- Equipment information
- etc.
In the data objects and I am not sure this is the best way.

timmathews:
@sumps, keep in mind that a lot (most?) of the schema objects are optional, so there is no hard requirement to support them.

I think it makes sense to keep all of this descriptive data in the object it describes (however, it probably(?) belongs in the `_meta` sub-object). So instead of:

```json
"batteries": {
  "house1": {
    "name": "House battery 1 (forward)",
    "location": "Under forward bunk",
    "manufacturerName": "MasterVolt",
    "manufacturerModel": "AGM 12/130",
    "chemistry": "FLA",
    "mode": {
      "value": "discharging"
    },
    "temperature": {
      "value": 290
    },
    "voltage": {
      "value": 12.5,
      "nominal": 12
    },
    "current": {
      "value": 0.1
    }
  }
}
```
perhaps
```json
"batteries": {
  "house1": {
    "mode": {
      "value": "discharging"
    },
    "temperature": {
      "value": 290
    },
    "voltage": {
      "value": 12.5,
      "_meta": {
        "nominal": 12,
        "minimum": 10,
        "maximum": 15
      }
    },
    "current": {
      "value": 0.1
    },
    "_meta": {
      "name": "House battery 1 (forward)",
      "shortName": "HB1",
      "location": "Under forward bunk",
      "manufacturerName": "MasterVolt",
      "manufacturerModel": "AGM 12/130",
      "chemistry": "FLA"
    }
  }
}
```

sumps:
@timmathews by putting this data in _meta sub-object, does this give us a way to get all of the _meta data with a restAPI call ? I am thinking of a way to efficiently extract this _meta data if we want to export it and share it and also to import it. This way if I have just bought a new boat and want to upload the _meta data for a similar/identical boat, there is an easy way to do it.

tkurki:
If all _meta is needed we can define an api route for retrieving that information. Just putting them in _meta doesn't add any additional functionality.

I don't see the whole _self_ json growing big enough to be unwieldy, one can also retrieve that and extract all _meta properties.

tkurki:
- [ ] move static information under _meta

rob42:
The idea behind the `_name` is that a server will strip out all `_*` names from the tree on normal transmission. Just a convenient way to easily mark a sub-tree as 'dont send'. But as @tkurki says, we just add a flag or request param `&meta=true` to get meta sent out too.

rob42:
Just looking over this, maybe these alarm and warning values for `vessels.*.electrical.dc.definitions.dcSource.current.measured"` belong in `.meta`?  If they are defined as zones they can be handled by the std alarms functionality.
```
vessels.*.electrical.dc.definitions.dcSource.current.faultLower";
vessels.*.electrical.dc.definitions.dcSource.current.faultUpper";
vessels.*.electrical.dc.definitions.dcSource.current.warnLower";
vessels.*.electrical.dc.definitions.dcSource.current.warnUpper";
```

sumps:
@rob42 what is the Std Alarm functionality ? I have read the Alarm section in the SignalK.org>Developers>Metadata but is there more info anywhere ? It looks to me that the Alarm method has been focused on a consumer being able to check a data value and see if an alarm should be triggered. There does not seem to be any consideration for NMEA alarms/alerts, which should be handled by Signal K.
Alarms needs to be given serious consideration or removed from the V1 spec. Maybe I have mis-understood something, but at the moment, it looks like we have alarms in the spec as a place holder but no clear mechanism for handling them.

rob42:
@sumps in this case (eg relative to dc power) it is a matter of compare and raise alarm. But your point re NMEA alarms is valid too. Jumping to  #96 for that one

fabdrol:
Shouldn't the definitions be moved to definitions.json?

tkurki:
@rob42 in schema (master + this branch) meta is without the underscore - the fix should be a separate issue.

tkurki:
@fabdrol If the definitions are only used in electrical_dc I want to keep this schema self-contained and not extract definitions to a separate file.

tkurki:
As the batteries part is done and the master version is unusable I think we should merge this asap.

tkurki:
Warnings and errors are for now just simple values, not zones.

PR #120, opened 2015-12-02
Improve propulsion

tkurki:
Propulsion needs a bit of work: reorganization & some additions.


tkurki:
Not ready to merge!

I've added just a sample document, without sources & timestamps, to illustrate the general organization of things under a propulsion item. There can be any number of propulsion items and one can mix & match engine data & transmission data:
- simple case: engine & transmission data under one id
- engine without transmission data: just leave transmission subgroup out
- handle engines & transmissions separately: use different id's

schemas/groups/propulsion.json
```diff
@@ -64,6 +41,12 @@
       "units": "Hz"
     },
 
+    "maxRevolutions": {
```
jboynes: Should this be in design?

schemas/groups/propulsion.json
```diff
@@ -82,6 +65,109 @@
...
+    },
+
+    "transmission": {
+      "type": "object",
+      "properties": {
+        "gear": {
+          "type": "object",
+          "properties": {
+            "value": {
+              "enum": ["Forward", "Neutral", "Reverse", "Fault"]
```
jboynes: Allow for multiple forward gears?

schemas/groups/propulsion.json
```diff
@@ -82,6 +65,109 @@
...
+          "units": "Pa"
+        }
+      }
+    },
+
+    "drive": {
+      "type": "object",
+      "properties": {
+        "type": {
+          "enum": ["saildrive", "shaft", "outboard"]
```
jboynes: jet?
fan? (for an airboat)

schemas/groups/propulsion.json
```diff
@@ -82,6 +65,109 @@
...
+        }
+      }
+    },
+
+    "fuel": {
+      "type": "object",
+      "properties": {
+        "type": {
+          "description": "Fuel type",
+          "enum": ["diesel", "petrol"]
```
jboynes: propane, coal, wood?
Should this be string rather than enum?

schemas/groups/propulsion.json
```diff
@@ -82,6 +65,109 @@
...
+        },
+        "used": {
+          "description": "Used fuel",
```
jboynes: Over what time period? Since installation, last tank refill, or this run?

schemas/groups/propulsion.json
```diff
@@ -82,6 +65,109 @@
...
+    "drive": {
+      "type": "object",
+      "properties": {
+        "type": {
+          "enum": ["saildrive", "shaft", "outboard"]
+        },
+        "trimState": {
+          "description": "Trim/tilt state",
+          "$ref": "../definitions.json#/definitions/numberValue",
+          "units": "K"
```
jboynes: rad or ratio?

schemas/groups/propulsion.json
```diff
@@ -82,6 +65,109 @@
       "units": "Pa"
     },
 
+    "alternatorVoltage": {
+      "description": "Alternator voltage",
+      "$ref": "../definitions.json#/definitions/numberValue",
+      "units": "V"
+    },
+
+    "engineTime": {
```
jboynes: How about simply `runTime`?

schemas/groups/propulsion.json
```diff
@@ -82,6 +65,109 @@
...
+      "description": "Total running time for engine",
+      "$ref": "../definitions.json#/definitions/numberValue",
+      "units": "s"
+    },
+
+    "engineTemperature": {
```
jboynes: Allow for `oilTemperature` and `coolantTemperature`?

schemas/groups/propulsion.json
```diff
@@ -82,6 +65,109 @@
...
+      }
+    },
+
+    "drive": {
+      "type": "object",
+      "properties": {
+        "type": {
+          "enum": ["saildrive", "shaft", "outboard"]
+        },
+        "trimState": {
```
jboynes: I'm assuming this is specifically angle from the horizontal e.g. for an outboard?
Do we need another for angle from the centreline for rotatable pods?

canboat: > On 17 Dec 2015, at 16:06, Jeremy Boynes <notifications@github.com> wrote:
> 
> In schemas/groups/propulsion.json <https://github.com/SignalK/specification/pull/120#discussion_r47915437>:
> 
> > +        },
> > +        "oilPressure": {
> > +          "description": "Oil pressure",
> > +          "$ref": "../definitions.json#/definitions/numberValue",
> > +          "units": "Pa"
> > +        }
> > +      }
> > +    },
> > +
> > +    "drive": {
> > +      "type": "object",
> > +      "properties": {
> > +        "type": {
> > +          "enum": ["saildrive", "shaft", "outboard"]
> > +        },
> > +        "trimState": {
> I'm assuming this is specifically angle from the horizontal e.g. for an outboard?
> 
Yes
> Do we need another for angle from the centreline for rotatable pods?	
> 
> 
I think that could be modeled as a rudder?

jboynes: My thought was a rudder is a passive system where this is actively thrusting in a specific direction, especially for a maneuvering unit like a bow thruster. One difference is that the thrust from a rudder is a function of boat speed, whereas thrust from the drive is a function of shaft speed.

For example, a small dinghy with an outboard that completely rotates has no rudder and steers purely based on drive thrust angle; it has no steering without drive thrust, same as a personal watercraft/jetski. Same principle applies to a large cruise ship or ferry with azimuth drives.

schemas/groups/propulsion.json
sumps: This could be added in the future but currently there is no mechanism for this info to be transmitted via NMEA2000. It is good to have an all encompassing spec but if the settings are not possible to set automatically from commercially available sensors and interfaces, it is a waste of time and I would prefer to add these as and when new sensors and interfaces become available.

schemas/groups/propulsion.json
sumps: The good thing is that Signal K is open and extensible and we should encourage people to think of developing new sensors that could detect the position and attitude of an outboard and then be the first interface that supports the new sensor, not try and second guess what may or may not be developed in the future.

sumps:
Do we have too many temperatures now? Should have engine oil, coolant and exhaust temps, plus transmission oil temp.

schemas/groups/propulsion.json
jboynes: The problem with using an enum here is that it does not allow other values to be used in a model and still have it validate. [Marine transmissions with multiple gears](http://www.coanracing.com/Catalog.asp?ProductLine=XMT) exist and it would be easy to monitor which gear was selected e.g. using GPIO inputs on a RPi running a SignalK server or on a micro-controller connected via Ethernet. 

That a _commercial_ vendor has not (yet) seen the ROI in bringing a sensor to market should not be a reason for preventing the model from representing this. Defining it as `string` with documented values would make this open.

sumps: I'll leave it to the software experts to argue the pros and cons of Enum versus Strings, but I am concerned that there are many other fields where Enum is used and to go through and identify and change them is a lot of work at this stage. Not sure how many extreme performance power boats will benefit from Signal K ;-)

tkurki:
@sumps Temperatures are as you listed.

schemas/groups/propulsion.json
tkurki: How about separating (clutch) `state` and `gear`?

tkurki:
@jboynes I've been thinking about your comment on enums.

In json itself there is no distinction between string and enum values. There is no schema validation in node server and I doubt there is any in the java server or other servers that provide a Signal K interface. 

Node server actually is not really aware of the schema at all, it just acts on incoming data and from what I gather so does java server.

So the added value of changing a current enum field to string would be...what exactly?

Now with enums we have the option to validate the structure, value types and values in Signal K data. This way we can check that widely used data such as NMEA0183 and NMEA 2000 data is mapped in a predictable way. 

If somebody chooses to use enum values that are not in the set specified in the schema it may break some consumers. With a string typed field the same thing would happen. Same thing if the software is not updated in sync with the spec.

My suggestion is that 
- we continue with the enum typed values for stuff that is naturally enumerable
- make sure the enum values cover most use cases
- write software defensively, eg. even if a field is an enum in the spec the software should not break with the appearence of "illegal" values (remember, somebody might add values to the spec behind your back)

How does that sound?

tkurki:
PS. @jboynes Please join signalk-dev Slack chat at http://slack-invite.signalk.org/

sumps:
I agree with this approach. Strings would give the flexibility but I think being able to validate and test the SK data in V1 and early iterations is very important and will give a bit more structure and consistency as we all start to use SK for the first time. Then with more practical experience and familiarity, we can introduce more flexible structures in V2.

rob42:
Ive added a few tweaks. I think that covering the more obscure drive and fuel types can be handled with 'other'. Ive added a `thrustAngle` to go with `trimState`, as steerable outboards etc are common, and not really rudders. (A friend has a cat with rudders and a steerable outboard). 


There is something wrong with `fuelRate` I think, it occurs in the engine, and `engine.fuel` tree. What is the thinking around the fuelRate* in `fuel` ? Is this acculmulated data for long trip reference or range calculations, or a record of the current averages etc. If so how is it timed, or reset?

rob42:
We could unclutter a bit by grouping the temperatures in `temperature.water` etc? Is that worth while?

sumps:
No I think we should keep the propulsion temperatures separated. They are well defined, common to many types of engine and are typically visualised in individual gauges on the boat dashboard.
You will get a chance to group other types of temperatures like, fridge, bait box, wine cellar (!), etc. When we support the NMEA2000 temperature PGN.

sumps:
On the fuel question @rob42 there is the instantaneous fuel flow data that comes from fuel flow sensors, fuel usage that can come from the EMU or from repeated fuel tank measurement, fuel remaining from fuel tank measurement and then the fuel efficiency and range type derived values. We need to start a PR on this subject as it needs some discussion, particularly when you consider multiple engines.

rob42:
Re the gears, we can cover all cases with a numeral to indicate the gear state.
-1 = reverse
0 = neutral
+1 = forward first gear
+2 = forward second gear, etc
Any invalid number for a given gearbox is a fault, defined in `.meta.zones`. Also allows you to mark reverse as a warning.

schemas/groups/propulsion.json
fabdrol: I'm also in favour of separating clutch and gear. One could imagine being in first, second, third etc gear forward or in reverse.

Another point: shouldn't `Fault` be implemented using an alarm (@rob42)?

schemas/groups/propulsion.json
fabdrol: Good question, raises a broader point:
- The type of engine(s), transmission(s) etc could be considered a part of design, where the *state* of the engine - being more static - could be separated. We had a similar discussion about sails. 
- The other side of the coin is - just as with sails - if you put the static properties in `design` propulsion related data becomes dispersed, making it harder for clients to retrieve information. That said, maybe we ought to designate `design` specifically as a group that contains information about the design of the vessel (hull, masts, interior, etc). Stuff that was decided upon when the boat was built and isn't dynamic at all.

tkurki:
We are starting to get overlapping work here ( #153 ) and it is high time to get this merged. Details can be amended in master.

sumps:
I agree it is very difficult to keep up with everything in so many branches

schemas/groups/propulsion.json
jboynes: It seems easier to me to write a consumer if the static design information is separated from the dynamic state information. My first thought would be to have the client pull the design data on first connection, use that to configure the display model, and then subscribe to the dynamic entries for ongoing updates. 

It would be easier and more efficient to subscribe to all dynamic values for the engine(s) (or sails or environment) at once rather than cherry-picking values out of a mix of static and dynamic ones, or subscribing to all engine data and receiving the same static data in each update.

tkurki: Having the dynamic and static data grouped by the entity they are about has advantages and disadvantages.

However I don't see the problem that you mention. If you subscribe to all propulsion values you will not get updates for the static data, as you can get delta updates only for data that has been updated, not the static data.

jboynes: This is probably my confusion. What happens if I subscribe to the parent path `vessels.foo.propulsion` e.g.
```json
"subscribe": [
    {
      "path": "vessels.foo.propulsion",
      "period": 1000,
      "format": "delta",
      "policy": "ideal",
      "minPeriod": 200
    }
]
```
Should I receive:
 * one update for the `propulsion` path with the full value?
 * one update for the `propulsion` path with a sparse value containing children that changed?
 * one update containing each source event that affected a value under the `propulsion` path, with each value affected?

tkurki: A good question. I was thinking about the simplest case, `instant` policy with no periods specified where you get updates as new data is received.

I think we have overengineered the subcription protocol and need to simplify.

Btw there's a bit of confusion also related to what goes into context and what in the path - in your example case `vessels.foo` should be in the context, not the path.

PR #121, opened 2015-12-03
Add Heave to environment.json

sumps:
Although it was not in my list of PGNs to be supported by iKommunicate V1, I know that some of you have been playing with 9 axis sensors and so perhaps it would be useful to have Heave included in SK. Basically negative measurement value means you are moving up on a wave and positive value means you are going down in to the trough of a wave.

### DIFF
--- a/schemas/groups/environment.json
+++ b/schemas/groups/environment.json
@@ -148,6 +148,12 @@
       "$ref": "../definitions.json#/definitions/numberValue",
       "units": "K"
     },
+   
+    "heave": {
+      "description": "Vertical movement of the vessel due to waves",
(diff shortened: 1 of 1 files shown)


rob42:
Merged into beta1

tkurki:
Should this be under `navigation` like attitude (roll, pitch, yaw) is?

sumps:
No Heave is effectively sea state and that is very much environment, although after the Time/Date discussions, I am no longer clear on what the scope of these two groups are ;-)

tkurki:
Makes sense. Attitude is boat, heave is sea.

PR #122, opened 2015-12-04
Add System Date and Time to navigation.json

sumps:
In order to allow SK devices to synchronise their date and time settings.

### DIFF
--- a/schemas/groups/navigation.json
+++ b/schemas/groups/navigation.json
@@ -332,6 +332,52 @@
       }
     },
 
+    "systemDateTime": {
+      "type": "object",
+      "title": "System Date and Time",
+      "description": "System Date and Time (UTC) for synchronisation",
+      "properties": {
+        "date": {
+          "type": "number",
+          "description": "Date in days since 1st Jan 1970",
+          "units": "days"
+        },
+
+        "time": {
+          "type": "number",
+          "description": "Time in seconds since midnight",
+          "units": "s"
+        },
+
+        "systemTimingSource": {
+          "type": "object",
+          "description": "Source of System Date and Time",
+          "properties": {
+            "value": {
+            "type": "string",
+            "enum": [
+              "GPS",
+              "GLONASS",
+              "Radio Signal",
+              "Internet",
+              "Local clock"
+              ]
+            }
+          }  
+        },
+
+        "source": {
+          "description": "Source of this data",
+          "$ref": "../definitions.json#/definitions/source"
+        },
+
+        "timestamp": {
+          "description": "timestamp of the last update to this data",
+          "$ref": "../definitions.json#/definitions/timestamp"
+        }
+      }
+    },
+
     "rateOfTurn": {
       "$ref": "../definitions.json#/definitions/numberValue",
       "description": "Rate of turn",
--- a/schemas/groups/navigation.json
+++ b/schemas/groups/navigation.json
@@ -194,7 +194,7 @@
         "timeAndDate": {
           "description": "GNSS Time and Date in ISO8601 format",
           "$ref": "../definitions.json#/definitions/timestamp",
-          "example": 2015-12-05T13:11:59Z
+          "example": "2015-12-05T13:11:59Z"
         },
 
         "gnssTimeSource": {
--- a/schemas/groups/navigation.json
+++ b/schemas/groups/navigation.json
@@ -190,6 +190,38 @@
             }
           }
         },
+       
+        "date": {
+          "type": "number",
+          "description": "Date in days since 1st Jan 1970",
+          "units": "days"
+        },
+
+        "time": {
+          "type": "number",
+          "description": "Time in seconds since midnight",
+          "units": "s"
+        },
+
+        "gnssTimeSource": {
+          "type": "object",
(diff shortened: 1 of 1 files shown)


timmathews:
Lets not repeat mistakes of the past, date and time should be one field. Also, lets be clear about the purpose of this field. It's fine for a rough approximation of the current time, but for synchronization between devices we need to use NTP.

So, lets combine the date and time fields, formatted as an ISO-8601 combined date/time stamp. [RFC 3339](https://tools.ietf.org/html/rfc3339) section 5.9 provides a formal definition of the format (it's the same as we are using for the timestamp fields currently).

In [argo](/timmathews/argo), I created a group called `system` to contain time, `~/system/currentTime` and `~/system/currentTimeSource`. I think this is still a reasonable approach (and makes more sense than time under `navigation`).

As a side note, we really need something on the network with a realtime clock. GPS time isn't UTC (although NMEA reports it as such). GPS time is GPS time, which as of June 30th of this year is 17 seconds ahead of UTC (and permanently 19 seconds behind TAI). GPS also transmits a leap-second correction in the subframe transmission, but that only occurs every 20 minutes. Which means that sometime in the first 20 minutes after you cold boot your GPS (assuming it recognizes and adjusts for the leap-second offset), your system time jumps backwards 17 seconds.

This is obviously very bad for timestamps. It would actually be better if we knew that no device made the correction and left it up to end users. Consistency and repeatability are more important that accuracy. But NMEA specified UTC for both their standards, which leaves us with this 20 minute period of uncertainty.

P.S. Other GNSS implementers have taken different approaches to time. GLONASS uses Moscow time (UTC+3) and adjusts for leap seconds. Galileo is synchronized with TAI. BeiDou is synchronized with UTC and does not adjust for leap seconds. Of the four (GPS, GLONASS, Galileo and BeiDou), only Galileo got it right IMO.

sumps:
I am not suggesting that GPS time and date is the only time and date that Signal K supports and I would be more than happy to see a combined time/date stamp as you have suggested (and not for the first time by the sounds of it ;-)

So many applications use GPS time/date and it is so widely understood now that I think we should include it as data in its own right under the navigation group, even if we have a separate Signal K System Time object.

When you are out at sea with no internet access and therefore no NTP, then GPS UTC is a very good fall back and matches the time used by other marine systems such as AIS base stations (AIS Message type 4), emergency DSC transmissions (maydays) and probably other systems I am not aware of.

I believe not including GPS time and date would be a mistake.

sumps:
Bearing in mind the important differences between the different GNSS systems, that @timmathews has made, it makes sense to call the objects gnssDate and gnssTime and a third field gnssTimeRef which could be GPS, GLONASS, GALILEO and BEIDOU. I bet we could get some good PR saying that Signal K supports Galileo and Beidou.

rob42:
Im with @timmathews on this - time is an attribute of the environment, not the boat. The idea of having gps time in the `navigation.gnss` makes sense, as thats where its most logical.
In my view the time is an attribute of the environment (or I could live with `system`), we should have an `environment.time` object,
```
"time":{
   "millis": 1234567890, //millis unix time
   "iso": "2015-12-06.....", //iso8601 timestamp
   "timezone": +/- hours //offset from UTC
}
```
Ideally its set from NTP, but in absence of that its up to the implementation to provide it - probably GPS time is next best.

```

timmathews:
We don't need to be connected to the Internet to do NTP. A Signal K server can be a very accurate NTP server as long as it has a GPS receiver connected which also provides a pulse-per-second (PPS) output. It's fairly straightforward to set up a stratum 1 NTP server with a GPS reference. A very nice second Signal K device would be a server with integrated GPS and NTP.

My argument against having time under navigation basically boils down to not wanting time values in two places, I think it will breed confusion. There will be multiple time sources on the boat, generally in the form of GPS receivers, but possibly also RTCs on SK servers or other sources (I've always wanted a Cesium clock). We already have a method to provide alternate sources for data and we should be using that for time as well. And at that point, it really doesn't make sense to have time from a non-navigational device under navigation.

@rob42, I'm good with either system or environment, but tend to prefer system (and also relocating sensors under system, but that's another discussion).

Keep in mind, that this has _nothing_ to do with timestamps on data or where the value for those timestamps comes from. We need to address that, because in an installation with only GPS time (which is only distributed via SK) we're going to run in to a situation where timestamps aren't accurate or worse where they jump backwards. Again, this is another discussion.

sumps:
OK, I have changed this so that we have "navigation.gnss.time", "navigation.gnss.date" and "navigation.gnss.gnssTimeSource" and this will be created from the GPS time and date in the NMEA2000 PGNs or the GPS time in the NMEA0183 "UTC" values common in many sentences and the GPS date from the NMEA0183 ZDA sentence if available.
It probably makes sense to use the "UTC" format HH:MM:SS.SS used in NMEA0183 sentences rather than the seconds since midnight format used in NMEA2000. However, I am not sure how to code this in the JSON schema, so at the moment the "navigation.gnss.time" field is still in seconds format.
There is currently no support for Galileo and Beidou in NMEA2000 and I will check on this with the NMEA committee.

tkurki:
Superceded by #136

PR #123, opened 2015-12-08
Added scope to anchor

rob42:
We probably should allow for chain counters etc.


canboat:
If we keep 'altitude' of the anchor, at least point out in the comment that the likely value is negative ;-)

rob42:
Heh, heh - I was going to do that but I thought it might be a bit flippant for a spec! Im just making another pull that makes position a $ref - so its moot now :-(

PR #126, opened 2015-12-09
Added time object and sample to environment.json

rob42:
No test, but sample data in vessel-time.json

### DIFF
--- a/samples/vessel-time.json
+++ b/samples/vessel-time.json
@@ -6,7 +6,10 @@
         "time": {
           "millis": 1449648657735,
           "timestamp": "2014-04-10T08:33:53Z",
-          "sourceRef": "sources.rtc",
+          "source": {
+	      "label": "Realtime clock",
+	      "type": "system"
+	    }
           "timezone": -1300
         }
       }
--- a/samples/vessel-time.json
+++ b/samples/vessel-time.json
@@ -6,7 +6,7 @@
         "time": {
           "millis": 1449648657735,
           "timestamp": "2014-04-10T08:33:53Z",
-          "source": "sources.rtc",
+          "sourceRef": "sources.rtc",
           "timezone": -1300
         }
       }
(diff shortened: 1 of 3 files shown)


tkurki:
- [ ] fix indentation
- [ ] fix test data to be valid, needs proper id as key & under vessel
- [ ] add test
- [ ] decision on system/environment/navigation
- [ ] one or several fields?

tkurki:
@rob42 why should we have alternative representations for the value in this case and not in any other?

In my view this is a "value" from a "source", for example pgn 126992, and should have the same value + source structure (including the possibility for multiple sources) like we have elsewhere.

tkurki:
Nmea0183 ZTG sentence doesn't have date information, should we have separate `timeofday` and `time`?

canboat:
I thought this environment time is supposed to be a real full UTC timestamp? Also we shouldn't  get distracted by silly deprecated NMEA 0183 sentences. ZTG is "Zulu Time To Go", full timestamp is available in "Zulu DAte" (ZDA): http://www.catb.org/gpsd/NMEA.html#_zda_time_amp_date_utc_day_month_year_and_local_time_zone

tkurki:
Sorry, my mistake, I mistook the NMEA0183 sentence timestamp for "system time". 

Luckily I can count on you Kees to correct my mistakes...

rob42:
formatted, and modified to hold `millis, timezone, timestamp and source` to be consistent with other value as Teppo points out. Since we have `timestamp`, which is iso format, we dont need `iso` so its removed.

This entry differs from gps time in #122 as this is supposed to hold the RTC or NTP derived real-time of the vessel, and info on its current time-zone, and hence required offset for display of local-time

tkurki:
@rob42 Reformatted & reworded a bit, and I added the test. 

And the test pointed out that the source references the source object that we have now in definitions.json, where it is defined as object with mandatory fields `type` and `label`. 

https://github.com/SignalK/specification/blob/6687c7eaf6f4b95203bf28df4bf2b228f7246f22/schemas/definitions.json#L18

PR #128, opened 2015-12-12
Removed CallsignDsc

sumps:
You just have one call sign for the VHF whether it is a DSC or non-DSC model.


rob42:
Merged into v1-beta1

PR #129, opened 2015-12-13
Replace last use of floatValue with numberValue

jboynes:
`floatValue` appears to be equivalent to `numberValue` and I only found one reference to it in the schemas. This changes `def_loader.json` to use `numberValue` and removes the declaration.


tkurki:
Thanks! I remember wondering about this myself.

PR #131, opened 2015-12-16
Update metadata and use rate for air pressure alarm

jboynes:
Refresh keyswithmetadata now new schema files were added.

Redefine `airPressureChangeRateAlarm` to use an actual rate rather than a quantity in an implicit time period.

### DIFF
--- a/keyswithmetadata.json
+++ b/keyswithmetadata.json
@@ -532,8 +532,8 @@
     "units": "C"
   },
   "environment.airPressureChangeRateAlarm": {
-    "description": "Change per hour which will cause an alarm",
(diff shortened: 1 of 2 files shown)


tkurki:
The simple script used to extract the keys doesn't quite work with real json schema structure. The keys in this update don't match actual schema structure, fir example compare against https://github.com/SignalK/specification/blob/5762ed0d02a025a6ec0431b117de404f35f2a121/test/data/electrical_dc.json. Furthermore there are upcoming changes to ac and dc schemas.

I'm going to close this so that it doesn't cause confusion. Could you please submit the air pressure change separately?

PR #133, opened 2015-12-16
Report engine revolutions using angular velocity

jboynes:


tkurki:
Why? This is `engine.revolutions`, not `engine.angularSpeed`.

jboynes:
When we say `revolutions` we mean revolutions per minute which can translate to either frequency or magnitude of angular velocity. By using magnitude we lose directionality i.e. that the shaft can rotate either way. A diesel won't but an electric drive could. Modeling this as the angular velocity of the shaft preserves this.

timmathews:
But regardless of the units, wouldn't the direction of rotation be indicated by sign? Might we be better off with adding a field to indicate rotation direction instead?

jboynes:
@timmathews negative values for frequency don't sound right to me - Hz is measuring the number of occurrences per second which could not be negative.

I don't see any reference to gearing in `propulsion.json` or elsewhere.

tkurki:
Please notice the work in progress in branches:  https://github.com/SignalK/specification/pull/120/files#diff-fe504afd1648650855f9167c0f8a541aR105

rob42:
My view is that both are awkward. The common term is  `rpm`, or slang `revs`. As a petrol-head all my life, Ive never heard it used as 'revolutions` which is a quantity not a rate, nor `angularSpeed`. I think  SI units are justified if they are a common alternative (meters/feet) or replace an awkward value with a std value (knots vs m/s) but not when we replace a common and useful value with an awkward one, with no practical benefit.

I think we should use the common term in this case `rpm`. Although `angularSpeed` may be SI, its just a confusion for users and a barrier to implementation. And we do need a holder for rotation direction. Maybe `rotation [CW|CCW]`. Apparenty SAE J 824 describes how to, relative to observer looking at the flywheel.

canboat:
Slang is exactly what that is: slang. 'revs' is short for 'revolutions'.

We've made the choice to use SI units, even if they are awkward and not the common term. Ergo pascal for pressure everywhere instead of mbar, psi, atm, hgm, and what have not. Awkward as it may be I think we should express this the same way. It is a rotation speed, expressed in radians per second in SI. Then we can have the usual conversion mapping it back to the usual `rpm` (which is actually one of the few properties that is expressed the same way in every locale that I know of.)

How about we call it `rotationSpeed` which avoids the worst aspect that it is unclear what physical property we're talking about.

Originally I thought we should use CW as positive and CCW as negative, but after reading http://www.sbmar.com/articles/which-way-does-my-engine-turn/ I am no longer so sure as it turns out that most marine engines are CCW, and it seems wrong to express CCW as a negative value, as we'd end up with most engines putting out negative rad/s!

One solution would be to say that positive is the "usual" or "value used when driving the boat forwards", with a possibility for negative values expressing the other direction. In the case of the common diesel + gear the engine rotationSpeed would still be positive when going backwards and the gear rotationSpeed would be negative. With an electric drive both could be negative (or just the engine if there is no gear.)

sumps:
I think using SI units where possible is the correct decision for SK but I do not think we should slavishly follow this in every case. Ultimately I think choice of units should be based on efficiency of system operation, followed by simplicity of understanding and consistency.
By using rad/s rather than RPM, a gateway will have to convert the RPM value to rad/s from the Engine Rapid Update PGN 10 times every second, plus another 10 more conversions per second for each additional engine. Then when the consumers receive the Signal K data they will have to convert it back from rad/s to RPM at this 10Hz rate, assuming you want nice smooth needle movement on the virtual engine gauges.
The same is true for engine turbo pressure which is also transmitted at 10Hz but as NMEA2000 and SK both use the SI Pascal unit, only the consumers will need to do the conversion.
For me RPM is one measurement that should not use SI units and if we come across other cases we should judge the choice of units in terms of system efficiency.

canboat:
@sumps we've been through this SI / not SI question before, and then we agreed to SI. As far as I am concerned we -should- slavishly follow SI.

For EVERY N2K to Signal K data point there WILL be a conversion because the N2K data is (according to the rumours I hear ;-) ) in fixed point with usually a very high precision, that does not match SI units. Length could be in nano meters or pressure in hecto pascals, etc. Whether you divide by 1000.0 or by some arbitrary floating point value does not matter.

fabdrol:
This shouldn't even be a question anymore, all values are SI. We had a lot of discussion about that. That said, some good points where made about the rotation direction (CCW/CW). Maybe the best way to go is to indicate `rotationSpeed` in `rad/s` and `rotationDirection` as an enum (CCW or CW). This is separate from gear, as an engine could be rotating CCW and be in forward gear or CW and be in forward gear. 

@jboynes - if nobody objects - could you update your PR?

tkurki:
Rereading this stuff my thoughts are
- For internal combustion engines revolutions per time unit has a physical meaning related to how the engine works. I'd like to keep the rate, Hertz is consistent with SI Units.
- For cases where  rotation direction is relevant we include `angularSpeed`, positive as the direction that usually creates forward thrust (or arbitrarily, think bow thrusters)

sumps:
I feel very strongly that we should do everything we can to make the conversion of the N2K rapid update PGNs as efficient as possible. These PGNs are transmitted at 10Hz and you can have multiple engines, GPS and Heading sensors all sending them out. 
By matching the units in the N2K rapid update PGNs to SK we will make the handling of these PGNs much more efficient. 
Values such as heading, COG, SOG, position, RPM, Turbo Pressure and a few other values are all coming off the network at 10Hz and it all adds up to a lot of pointless conversions if we simply use SI units for some reason that I have forgotten.

jboynes:
I'll update the PR to use Hz and add a direction indicator.

@sumps To me be advantage of using standard units is simply that they are
standard. This facilitates the interchange between applications and makes
us more open, especially for values that are not defined by NMEA and for
users who are not privy to the closed specs. I feel very strongly we should
not tie ourselves to factors that are not open.

re overhead, with reasonable modern hardware, I am skeptical as this should
be a single multiply-add operation given the equivalent dimensionality.
With devices connecting over high-speed protocols (even as low as a 10Mb
Ethernet class), I expect the minimum class of SoC needed to handle comms
would support that number of conversions. Is my expectation incorrect?

sumps:
There is no excuse for code inefficiency and the faster and more efficient we can make a process the more the hardware can do and the better the user experience, no one every complains of software running too quickly or smoothly.
This is not an open or closed standards argument, this is whether it would be better for some fields to use units that are globally recognised (anyone who has been in a car understands RPM), that are fit for purpose, easy to read in our "Human Readable" JSON format and can be handled in the most efficient way.
Over the years I have continuously seen embedded hardware platforms fall short on speed and performance due to inefficient code which then has to be optimised for the life of the product as new features are added. I would rather that did not happen to iKommunicate.
Also HTML5 apps are often sluggish on some mobile devices or in some browsers, so the more we can do to avoid unnecessary conversions at the browser end, should be considered.
Let us take a simple web app like Instrument Panel, if it is connected to iKommunicate which in turn is connected to an N2K network with two engines, two GPS units and a heading sensor. It will receive 50 PGNs a second containing Position, COG, SOG, Heading, Turbo Pressure, RPM and Engine Tilt, plus PGNs from whatever other devices are on the network and AIS data from perhaps 100 or more vessels in a popular sailing area.
All of that data has to be converted in to SK and then when it reaches the browser be converted back, so I think optimising and making this as efficient as possible is worth considering, above and beyond a blanket use of SI units, so that we can say the format is "standardised".
I would be happy to be proved wrong on this point, but it would be a crying shame if SK Version 1 was slow and inefficient on this first generation of SK devices.

jboynes:
@sumps I agree with you about eliminating inefficiencies but the loss of utility here (from inconsistency) does not seem worth it. As a newcomer to the spec, knowing everything is SI based made understanding the system very easy.

What's happening on the browser side would not impact headless components like I assume iKommunicate is. Server-side I would expect the conversion from/to the text on the wire to take more resources than the scaling. In the browser, rendering any graphic artifact (like an analogue dial widget, or even font text in a GUI) is going to take way more compute than the re-conversion.

jboynes:
Closing this and replacing with #154

PR #136, opened 2015-12-20
Add datetime with source information to the Signal K model

tkurki:
This adds a place in the Signal K data model for datetime information from GNSS, usually GPS.

This is essentially the same as #122 with the value & source grouped in one object and added timestamp & source to be in line with all other values.

### DIFF
--- a/schemas/groups/navigation.json
+++ b/schemas/groups/navigation.json
@@ -190,20 +190,20 @@
             }
           }
         },
-        
+
         "integrity": {
           "type": "object",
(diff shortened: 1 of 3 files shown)


rob42:
I think "exmple" is usually spelled "example"
Not your day on this one :-)

Otherwise good, just merge I think

tkurki:
E

sunnuntai 20. joulukuuta 2015 rob42 <notifications@github.com> kirjoitti:

Issue #138, opened 2015-12-21
Add support for navigation lights


rob42:
Add a `navigation.lights` enum that allows the nav lights status to be set. The options added dont cover all types and sizes of vessels, they assume that the state is correct for the current vessel, eg if the vessel is 6M the motoring lights will be different than a vessel of 14M, but both will be `motoring`

PR #139, opened 2015-12-21
Added lights to navigation 

rob42:
See issue #138


tkurki:
Extract to definitions?

rob42:
The enums are not quite identical. Lights have 'fault', off' etc, nav state has 'not under command'

rob42:
Pushed into v1-beta1

PR #141, opened 2015-12-21
Added environment.mode

rob42:
Allows us to set the vessel to `night` mode and based on `navigation.state` we can dim the instruments, set the nav lights, or warn if those are incorrectly set. In turn `navigation.state` can be largely derived from other vessel parameters like SOG, motor params, time, etc. So this also makes smart automation easier.


rob42:
Pushed into v1-beta1

PR #144, opened 2015-12-22
Allow for vessels not registered at the national level

jboynes:


jboynes:
In the US a vessel need not be registered at the national level (with the USCG). Instead, a state level registration would be displayed on the bow. I called this "local" to avoid confusion over state/province/district naming - the intention is some level below national.

tkurki:
LocalRegistration sounds like a good idea. Should it be composite with scheme + id instead of just a plain string?

jboynes:
I was avoiding coming up with a scheme for identifying registration authorities as that seemed complex without a lot of benefit :)

We could perhaps combine all registrations into a map, something like:

```javascript
vessel.registrations = {
  "imo": "IMO 9074729",
  "nz": { "partB": "NZ654" },
  "us": { "uscg": "123456", "wa": "WN 1234 XY" }
}
```

Use cases to cover:
* A undocumented vessel in the US may have just a state registration
* A US vessel may have a national USCG registration and a state registration
* A NZ owned vessel imported to the US may have a NZ registration, a US-CBP cruising permit, and a US state registration (for tax purposes)

I'm sure there are other combinations in other jurisdictions.

rob42:
`vessel.registrations` is a good solution - lets go with that - its simple, descriptive and covers all cases i can think of

rob42:
See #145, closing this

PR #145, opened 2015-12-23
Added registration structure

rob42:
Replaces #144 with a registrations tree structure, allowing multiple national, local and other registrations


rob42:
Merged into v1-beta1

rob42:
If there are no further comments I will merge tomorrow

PR #148, opened 2016-01-06
Add schema and tests for endpoint discovery

jboynes:
Added a schema for the endpoint discovery document.
Added simple validation test.
Relaxed definition of what will be accepted for a URL to allow for relative locations and arbitrary protocols.

### DIFF
--- a/schemas/definitions.json
+++ b/schemas/definitions.json
@@ -229,8 +229,7 @@
 
     "url": {
       "type": "string",
-      "description": "A valid URL.",
-      "pattern": "^(http(s?):.*|mailto:.*|tel:(\\+?)[0-9]{4,})$"
+      "description": "A location of a resource, potentially relative. For hierarchical schemes (like http), applications must resolve relative URIs (e.g. './v1/api/'). Implementations should support the following schemes: http:, https:, mailto:, tel:, and ws:."
     },
 
     "sail": {
--- a/schemas/discovery.json
+++ b/schemas/discovery.json
@@ -25,8 +25,22 @@
             "signalk-ws": {
               "description": "The URL of a WebSocket streaming endpoint e.g. http://localhost/signalk/v1/stream",
               "$ref": "definitions.json#/definitions/url"
-            }
-          }
+            },
+            "signalk-tcp": {
+              "description": "The host and port of a tcp socket streaming endpoint e.g. localhost:55555",
+              "type": "object",
+              "required": ["host","port"],
+              "properties": {
+                "host": {
+                  "description": "The host, fqdns or IP address.",
+                  "type": "string"
+                },
+                "port": {
+                  "description": "The tcp port",
+                  "type": "integer"
+                }
+              }
+            }          }
         }
       }
     }
--- a/schemas/discovery.json
+++ b/schemas/discovery.json
@@ -27,20 +27,10 @@
               "$ref": "definitions.json#/definitions/url"
             },
             "signalk-tcp": {
-              "description": "The host and port of a tcp socket streaming endpoint e.g. localhost:55555",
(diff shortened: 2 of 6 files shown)


schemas/definitions.json
```diff
@@ -229,8 +229,7 @@
 
     "url": {
       "type": "string",
-      "description": "A valid URL.",
-      "pattern": "^(http(s?):.*|mailto:.*|tel:(\\+?)[0-9]{4,})$"
+      "description": "A location of a resource, potentially relative. For hierarchical schemes (like http), applications must resolve relative URIs (e.g. './v1/api/'). Implementations should support the following schemes: http:, https:, mailto:, tel:, and ws:."
```
tkurki: What use case prompted you to include the relative urls? I'm just wondering where that might come useful and what it would be relative to?

tkurki: Why move the protocol stuff to description and not extend the regexp?

schemas/discovery.json
```diff
@@ -0,0 +1,48 @@
...
+                "host": {
+                  "description": "The host, fqdns or IP address.",
+                  "type": "string"
+                },
+                "port": {
+                  "description": "The tcp port",
+                  "type": "integer"
+                }
+              }
+            }          }
```
tkurki: indenting/line break

test/data/endpoints.json
```diff
@@ -0,0 +1,18 @@
+{
+  "endpoints": {
+    "v1": {
+      "version": "1.1.2",
+      "signalk-http": "http://192.168.1.2/signalk/v1/api/",
+      "signalk-ws": "ws://192.168.1.2:34567/signalk/v1/stream",
+      "signalk-tcp": {
```
tkurki: I suppose `tcp://localhost:5555` won't cut it, even if this makes tcp the oddboll of the three.

test/endpoints.js
```diff
@@ -0,0 +1,10 @@
+var chai = require('chai');
+chai.Should();
+chai.use(require('../index.js').chaiModule);
+var _ = require('lodash')
```
tkurki: lodash not needed.

test/data/endpoints.json
rob42: Re my comment above "I think we should use it anyway.." I meant lets use ```tcp://localhost:55555```.
As Teppo says - it keeps the format std, and we are going to get lots more in the future udp/rdp/mqtt/stomp/coap/thread, many of which will be unofficial

jboynes:
Switched to using a URL for TCP endpoints.

schemas/definitions.json
jboynes: @tkurki The regex pattern for URLs only describes a small subset of potential ones and does not reflect the actual syntax allowed. For example, the original `(http(s?):.*` fragment only says http(s) is a valid scheme but does not enforce any of the restrictions of what's allowed in the scheme-specific part. With the escaping and I10N aspects of valid URL syntax I did not think the regex added much.

Relative URLs would avoid having to generate full absolute values for references to resources on the same server. For example, the discovery result could be
```json
{
  "endpoints": {
    "v1": {
      "vers, ion": "1.1.2",
      "signalk-http": "/signalk/v1/api/",
    }
  }
}
```
which has the advantage of being constant for any installation of a server it it's following our convention. Actually, all same the advantages as for using relative references in HTML.

Resolution would be done [as defined by the RFC](https://tools.ietf.org/html/rfc3986#section-5.2), so if that data was returned by a GET to `https://10.0.1.1/signalk` then the resolved path would be `https://10.0.1.1/signalk/v1/api/`

tkurki:
Thank you!

PR #151, opened 2016-01-10
Use SPDX abbreviation of the license

tkurki:
ping @SignalK/core, I'd like somebody else to sign this off.


rob42:
Do we also need an SPDX file, and we probably need a LICENCE file?

PR #152, opened 2016-01-11
Improve alarms - add alert and emergency

rob42:
Add alert and emergency types, and improve descriptions

### DIFF
--- a/samples/signalk-depth-meta-attr.json
+++ b/samples/signalk-depth-meta-attr.json
@@ -17,8 +17,8 @@
                     "shortName": "DBK",
                     "gaugeType": "sparkLine",
                     "units": "Meters (m)",
-                    "warnMethod": "visual",
-                    "alarmMethod": "sound",
+                    "warnMethod": ["visual"],
+                    "alarmMethod": ["sound"],
                     "zones": [
                         [0.0,1.5,"alarm", "Running aground!"],
                         [1.5,3.0,"warn" , "Shallow water!"],
--- a/schemas/definitions.json
+++ b/schemas/definitions.json
@@ -47,7 +47,7 @@
         }
       }
     },
-    
+
     "version": {
       "type": "string",
       "description": "Version of the Signal K root object.",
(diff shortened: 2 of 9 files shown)


rob42:
Updated docs https://github.com/SignalK/specification/wiki/Alarm-Handling

sumps:
This approach to alarms seems to be very server centric and not so efficient.
Is polling through every data objects meta data to discover if any alarms are active the best solution ?
What about having a higher level (in the schema tree) section that lists the active alarms and the URL of the data object they refer to, then it is only ever a single poll to check if any alarms are active and a second poll to get the status of the data objects meta data alarm value.
Also for delta files a regular alarm status object sent every few seconds and immediately on the occurrence of an NMEA alarm/alert would be a better method for less complex consumers.

rob42:
"Is polling through every data objects meta data to discover if any alarms are active the best solution ?"

The intention is that the server does this as a background process, and when alarm conditions are met it sets/removes the appropriate alarm in `vessel.[uuid].alarms...`. 

If a client wants to be notified of alarms coming and going they should subscribe to `vessel.[uuid].alarms.*`

If a device wants to set an alarm explicitly it just send a deltas with `vessel.[uuid].alarms...`

Hmmm, `vessel.[uuid].alarms` might be better named `vessel.[uuid].notifications` ?

rob42:
Some breaking changes here - in the alarms object `alarmState` is now `state`, `alarmMethod` is now `method`. The 'alarm' prefix was un-needed and confusing when `meta` has alarm* keys.

Also the possible values are now enums in definitions so consistent throughout. The alarm object had to have a unique path to allow for multiple alarms on the same key, and alarm collisions, so uuid was added.

fabdrol:
@rob42 I like "notifications" over "alarms", as it can be used for more than alarms (e.g. "your coffee is ready", would be great. But not really an alarm. Or, more realistic, a notification to note your position in the log every hour). 

Aside from that.. personal preference (I really do like coffee), wouldn't be simpler to  assess the alarm state of every incoming value when it comes in, and do something to it? Each alarm has one or more "triggers", which are other properties that need to be monitored whenever they change. The related alarms' states are updated accordingly. That way, no polling is needed.

sumps:
@rob42 well done for getting this merged, but I would be really grateful for a simple example of an NMEA2000 alert PGN being received by a gateway and converted in to a Delta. I have read and re-read the Wiki post on Alarms and I am still not 100% sure how to specify this process.

rob42:
Ive updated the wiki with more info, and samples. Ive removed the alarm uuid. See #160

PR #153, opened 2016-01-19
added alternator potential

damianham:


sumps:
Hi Damian, 
In the "improve-propulsion" branch we have a similar data object but called alternatorVoltage which is probably more understandable than alternatorPotential....

"alternatorVoltage": {
      "description": "Alternator voltage",
      "units": "V"
    },

damianham:
Hi sumps

Yes I agree alternatorVoltage is more understandable - I used the naming convention from the json in 
n2k-signalk/test/todo.txt which I presume comes from canboat.

Futhermore if there is no requirement to stick to the naming convention from canboat then I would suggest the 'alternator' part is superfluous.  A 'voltage' engine parameter would always be from the alternator one would assume.

sumps:
I like having alternatorVoltage, as there can be a lot of voltage measurements in a boat and even though this data field is associated with an engine, not everyone may jump to that association and who knows, in the future we may start supporting Electric Outboards, in which case the voltage for that type of engine might need to be called supplyVoltage rather than just voltage.

canboat:
How about multiple alternators? Quite common nowadays. I’ve got a 24V for the house bank and a 12V for the engine starter battery.

sumps:
The alternatorVoltage value comes from the Engine Status PGN and is based on the assumption that additional alternators will be retro-fitted (only one alternator is standard fit on an engine) with a separate voltage sensor that will probably use the DC Current/Voltage PGN # 127751 or possibly the Battery Status PGN # 127508 if the alternator is charging a specific battery.

damianham:
OK sounds reasonable - I have changed the name and pushed new commits

timmathews:
Just because n2k assumes one alternator is no reason for us to assume one alternator. That said, it seems confusing to me to have alternator voltage under propulsion and not under electrical.dc.

Alternators aren't intrinsically part of an engine, there is no hard requirement for one. I sailed with an old Volvo Penta for two years without an alternator.

tkurki:
I agree that having electrical stuff under propulsion is a bit confusing.

Once we get multiple values #118 merged it will allow multiple alternator voltages under the engine.

There's plenty of stuff (coolant as an example) that is not intrinsically part of an engine but that is not a reason to exclude that feature. 

Now that DC schema is updated I think this pull request is no longer appropriate. If you guys feel that the issue needs further attention please open an issue and let's continue discussion there.

@damianham could you please update your n2k-signalk PR to use the just updated master schema?

sumps:
The reality is that nearly every marine petrol or diesel engine will have an alternator that is physically fitted to the engine and the output of that alternator will be directly related to the RPM of the engine. In most medium/large power boats, there is a volt meter in the engine gauge panel that shows the alternator charging voltage and that is why it is linked  to the engine PGN and why we should keep it in propulsion.
If your engine is running and the alternator voltage is low, you will go to the engine compartment, check the fan belt and fault find the alternator. A petrol engine will only continue working if the alternator is working correctly and even some modern diesels will fail if the ECM is not getting a voltage from the alternator. There fore the alternator is inherently linked to an engine, even if there are rare cases of older engines without one or some vessels having more than one.

PR #157, opened 2016-01-24
Changed plain "string" type properties to reference stringType

### LINKED ISSUE #156: AIS vessel name structure
In the current schema vessel name is a plain string. The `self` vessel's name doesn't normally appear in the delta stream (it might if it changes?).

AIS vessel names appear in the delta stream when AIS static data messages are received. This puts name in path `name` under vessel's context root in the delta message.

When deltas are added to full  tree there is a pretty simple logic: if the `value` received for a path is a plain value (number, string) it is inserted into the `value` property:
```
"speedOverGround": {
  "value": 3.53,
  "source": {},
  "timestamp": "2014-08-15-19:09:16.334"
}
```
and if the `value` is an object its properties will be inserted into the tree:
```
"position": {
  "longitude": 24.7276789,
  "latitude": 59.7086808,
  "source": {...},
  "timestamp": "2014-08-15-19:09:15.633"
}
```

There is no mechanism to distinguish the incoming AIS static update's `name` field to be used as it is and not wrapped without adding total schema awareness to this logic, which I would definitely like to avoid.

I suggest we alter the schema and wrap the `name` field in the normal value structure.

I guess there are / will be other cases like this.

fabdrol:
This PR fixes #156 and applies the same resolution throughout the specification. 

Properties of plain `type: "string"` have been moved to `$ref` `stringValue` which includes `source`, `timestamp`, `meta` & `_attr` just as `numberValue` and such.

### DIFF
--- a/schemas/groups/communication.json
+++ b/schemas/groups/communication.json
@@ -6,53 +6,53 @@
   "title": "communication",
   "properties": {
     "callsignVhf": {
-      "$ref": "definitions.json#/definitions/stringValue",
+      "$ref": "../definitions.json#/definitions/stringValue",
       "description": "Callsign for VHF communication",
       "example": "ZL1234"
     },
 
     "callsignHf": {
-      "$ref": "definitions.json#/definitions/stringValue",
+      "$ref": "../definitions.json#/definitions/stringValue",
       "description": "Callsign for HF communication",
       "example": "ZL3RTH"
     },
 
     "phoneNumber": {
-      "$ref": "definitions.json#/definitions/stringValue",
+      "$ref": "../definitions.json#/definitions/stringValue",
       "description": "Phone number of skipper",
       "example": "+64xxxxxx"
     },
 
     "emailHf": {
-      "$ref": "definitions.json#/definitions/stringValue",
+      "$ref": "../definitions.json#/definitions/stringValue",
       "description": "Email address to be used for HF email (Winmail, Airmail, Sailmail)",
       "example": "motu@xxx.co.nz"
     },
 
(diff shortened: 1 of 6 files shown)


timmathews:
Do we really need to do this? Are we committing to having a `timestamp`, `source`, etc. on every value, even things that don't change and don't have a source other than "boat owner"?

tkurki:
Timestamp and source are optional, but value is there to stay.

Not really happy with this. See original problem - any other suggestions?

timmathews:
Perhaps the simple thing to do is have n2k-signalk and signalk-parser-nmea0183 output a delta that looks like this for these sorts of values:

```js
{
  "updates": [{
    "source": {
      "label": "",
      "type": "NMEA2000",
      "pgn": 129809,
      "src": "43"
    },
    "timestamp": "2014-08-15T15:00:04.655",
    "values":[{
      "path": "",
      "value": {
        "name": "LAGUNA"
      }
    }]
  }],
  "context": "vessels.urn:mrn:imo:mmsi:230044160"
}
```

Then the object `{"name": "LAGUNA"}` will just be merged into the specific vessel object pointed to by the context. signalk-multiplexer-node may need to be modified to support an empty path.

tkurki:
You have to ask yourself one question: "now why didn't I think of that?".

Brilliant! 

Added this to the "new" multiplexer a k a FullSignalK: https://github.com/SignalK/specification/commit/4bd476f05749b62d88ccb43a29fa8290c6bd6884


https://github.com/SignalK/signalk-parser-nmea0183/issues/43

PR #159, opened 2016-01-30
Update vessel.json

joabakk:
Description misleading


tkurki:
UUID should be dropped as well, as it is not Uuid we have here.

PR #161, opened 2016-01-31
Convert zone in meta to an object

rob42:
As teppo pointed out the zones in meta is an array of arrays.The zone definitions are better described by objects as they have fixed formats
This converts the zones array of arrays to an array of objects

### DIFF
--- a/samples/signalk-depth-meta-attr.json
+++ b/samples/signalk-depth-meta-attr.json
@@ -22,7 +22,7 @@
                     "zones": [
                         {"lower":0.0,"upper":1.5,"state":"alarm", "message":"Running aground!"},
                         {"lower":1.5,"upper":3.0,"state":"warn", "message":"Shallow water!"},
-                        {"lower":3.0,"upper":9999.9,"state":"normal", "message":""}
+                        {"lower":3.0, "state":"normal"}
                     ]
                 }
             }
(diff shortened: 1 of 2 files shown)


samples/signalk-depth-meta-attr.json
```diff
@@ -20,9 +20,9 @@
                     "warnMethod": ["visual"],
                     "alarmMethod": ["sound"],
                     "zones": [
-                        [0.0,1.5,"alarm", "Running aground!"],
-                        [1.5,3.0,"warn" , "Shallow water!"],
-                        [3.0,9999.9,"normal",""]
+                        {"lower":0.0,"upper":1.5,"state":"alarm", "message":"Running aground!"},
+                        {"lower":1.5,"upper":3.0,"state":"warn", "message":"Shallow water!"},
+                        {"lower":3.0,"upper":9999.9,"state":"normal", "message":""}
```
tkurki: Lose the upper here?

schemas/definitions.json
```diff
@@ -523,40 +523,40 @@
           "description": "The zones defining the range of values for this signalk value.",
           "items": [
             {
-              "type": "array",
+              "type": "object",
```
tkurki: Make state required and leave others optional?

PR #167, opened 2016-02-01
Update navigation.json

sumps:
Add missing Data objects, modify Data Object names to be more standardised and familiar and change Destination to be compatible with Class A AIS Voyage Data.

### DIFF
--- a/schemas/groups/navigation.json
+++ b/schemas/groups/navigation.json
@@ -49,6 +49,13 @@
         }
       }
     },
+
+   "speedOverGround": {
+      "$ref": "../definitions.json#/definitions/numberValue",
+      "description": "Vessel speed over ground",
+      "units": "m/s"
+    },
+
     "courseOverGroundMagnetic": {
       "description": "Course over ground (magnetic)",
       "$ref": "../definitions.json#/definitions/numberValue",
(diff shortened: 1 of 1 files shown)


rob42:
This doesnt really work in this form, but its a useful start. There are several things mixed here, which need to be separated for a clean solution.
1) A route is composed of legs between waypoints. 
2) A leg is composed of segments between marks. The start waypoint is mark[0], the end waypoint is mark[n].
3) The vessel could be on a very simple route, (start waypoint to end waypoint), with a straight line between, (eg mark[0] to mark[1]), or multi-waypoints, each leg having many marks.

http://www.bandg.com/Glossary/ defines `course` as "Course is a planned route to a destination, Waypoint or Mark."
I have started setting up a structure with `navigation.course`, which will hold data on the current leg. This includes all the keys in the commit above.

sumps:
Hi Rob, I was not happy with the expression Active Route but it was already called that in the Spec and I struggled to think of something better but Navigation Course makes sense and I agree with that.
This section is all about the current navigational status of the vessel and whether we are doing a simple GOTO type function to a single Waypoint/Mark or a leg in a more complex route, it should not make any difference.
This PR is designed to match SK to the traditional NMEA0183 and NMEA2000 methods of transferring Next Waypoint data and other data required by an autopilot to steer a safe course. Although SK will not support Autopilot Control in V1, this will follow very soon, so it is important that we get the data structure right for V1.

rob42:
See #168 
Closing this

PR #168, opened 2016-02-03
Rework and extend the navigation.activeRoute into navigation.course

rob42:
Replaces #167 
Changes the `activeRoute` object to `course` and adds support for missing data including great circle values

### DIFF
--- a/schemas/groups/navigation.json
+++ b/schemas/groups/navigation.json
@@ -72,7 +72,7 @@
         },
         "geometry": {
           "title": "LineString",
-          "description": "The GeoJSON LineString that defines the course between the last and next waypoints.",
+          "description": "The GeoJSON LineString that defines the course between the last and next waypoints. When null, use a straight line",
           "properties": {
             "type": {
               "enum": [
@@ -80,17 +80,32 @@
               ]
             },
             "coordinates": {
-              "$ref": "#/definitions/lineString"
+              "$ref": "../definitions.json#/definitions/lineString"
+            },
+            "timestamp": {
+              "$ref": "../definitions.json#/definitions/timestamp"
+            },
+
+            "source": {
+              "$ref": "../definitions.json#/definitions/source"
+            },
+
+            "_attr": {
+              "$ref": "../definitions.json#/definitions/_attr"
+            },
+
+            "meta": {
+              "$ref": "../definitions.json#/definitions/meta"
             }
           }
         },
         "nextMark": {
           "$ref": "../definitions.json#/definitions/numberValue",
-          "description": "Number of next mark"
+          "description": "Array position (zero-based) of next mark in the geometry LineString"
         },
         "lastMark": {
           "$ref": "../definitions.json#/definitions/numberValue",
-          "description": "Number of last mark"
+          "description": "Array position (zero-based) of last mark in the geometry LineString"
         },
         "vmg": {
           "$ref": "../definitions.json#/definitions/numberValue",
@@ -122,23 +137,23 @@
           "description": "The estimated time of arrival at the end of the current leg of the route",
           "properties": {
             "value": {
-              "$ref": "#/definitions/timestamp"
+              "$ref": "../definitions.json#/definitions/timestamp"
             },
 
             "timestamp": {
-              "$ref": "#/definitions/timestamp"
+              "$ref": "../definitions.json#/definitions/timestamp"
             },
 
             "source": {
-              "$ref": "#/definitions/source"
+              "$ref": "../definitions.json#/definitions/source"
             },
 
             "_attr": {
-              "$ref": "#/definitions/_attr"
+              "$ref": "../definitions.json#/definitions/_attr"
             },
 
             "meta": {
-              "$ref": "#/definitions/meta"
+              "$ref": "../definitions.json#/definitions/meta"
             }
           }
         },
--- a/schemas/groups/navigation.json
+++ b/schemas/groups/navigation.json
@@ -37,12 +37,10 @@
             "mine clearance"
           ]
         },
-
         "source": {
           "description": "Source of this data",
           "$ref": "../definitions.json#/definitions/source"
         },
-
         "timestamp": {
           "description": "timestamp of the last update to this data",
           "$ref": "../definitions.json#/definitions/timestamp"
@@ -54,109 +52,171 @@
       "$ref": "../definitions.json#/definitions/numberValue",
       "units": "rad"
     },
-
     "courseOverGroundTrue": {
       "description": "Course over ground (true)",
       "$ref": "../definitions.json#/definitions/numberValue",
       "units": "rad"
     },
-
-    "activeRoute": {
+    "course": {
       "type": "object",
-      "title": "Active Route",
-      "description": "The currently active route",
+      "title": "Course to next mark",
+      "description": "The current leg of the planned course or route",
       "properties": {
-        "timestamp": {
-          "$ref": "../definitions.json#/definitions/timestamp",
-          "description": "Time of last update to active route"
+        "lastWaypoint": {
+          "description": "The position of the last waypoint in 3 dimensions",
+          "$ref": "../definitions.json#/definitions/position"
         },
-
-        "source": {
-          "$ref": "../definitions.json#/definitions/source",
-          "description": "Source of last update to active route"
+        "nextWaypoint": {
+          "description": "The position of the next waypoint in 3 dimensions",
+          "$ref": "../definitions.json#/definitions/position"
         },
-
-        "bearingActual": {
-          "type": "number",
-          "description": "The current bearing of the next waypoint relative to true North",
-          "units": "rad"
+        "nextMark": {
+          "$ref": "../definitions.json#/definitions/numberValue",
+          "description": "Number of next mark"
         },
-
-        "distanceActual": {
-          "type": "number",
-          "description": "The current distance to the next waypoint",
-          "units": "m"
+        "lastMark": {
+          "$ref": "../definitions.json#/definitions/numberValue",
+          "description": "Number of last mark"
         },
-
-        "bearingDirect": {
-          "type": "number",
-          "description": "The bearing relative to true North from last waypoint to the next waypoint",
-          "units": "rad"
+        "vmg": {
+          "$ref": "../definitions.json#/definitions/numberValue",
+          "description": "The velocity made good towards the next waypoint",
+          "units": "m/s"
         },
-
-        "courseRequired": {
-          "type": "number",
+        "courseRequiredTrue": {
+          "$ref": "../definitions.json#/definitions/numberValue",
           "description": "The course relative to true North towards the next waypoint",
           "units": "rad"
         },
-
+        "courseRequiredMagnetic": {
+          "$ref": "../definitions.json#/definitions/numberValue",
+          "description": "The course relative to magnetic North towards the next waypoint",
+          "units": "rad"
+        },
+        "timeToGo": {
+          "$ref": "../definitions.json#/definitions/numberValue",
+          "description": "Time to Go in seconds before we reach the next waypoint",
+          "units": "s"
+        },
+        "xte": {
+          "$ref": "../definitions.json#/definitions/numberValue",
+          "description": "Cross track error",
+          "units": "m"
+        },
         "eta": {
-          "$ref": "../definitions.json#/definitions/timestamp",
-          "description": "The estimated time of arrival at the end of the current route"
-        },
-
-        "route": {
-          "type": "string",
-          "description": "UUID of the current route, found in the routes list"
-        },
-
-        "startTime": {
-          "$ref": "../definitions.json#/definitions/timestamp",
-          "description": "The time this route was activated"
-        },
-
-        "waypoint": {
           "type": "object",
-          "title": "waypoint",
-          "description": "The last and next waypoints data",
+          "description": "The estimated time of arrival at the end of the current leg of the route",
           "properties": {
-            "lastTime": {
-              "$ref": "../definitions.json#/definitions/timestamp",
-              "description": "The time the last waypoint was reached"
+            "value": {
+              "$ref": "#/definitions/timestamp"
             },
 
-            "last": {
-              "type": "string",
-              "description": "UUID of the last waypoint, found in the waypoints list"
+            "timestamp": {
+              "$ref": "#/definitions/timestamp"
             },
 
-            "nextEta": {
-              "$ref": "../definitions.json#/definitions/timestamp",
-              "description": "Estimated time of arrival at the next waypoint",
-              "example": "2014-03-24T00:15:41Z"
+            "source": {
+              "$ref": "#/definitions/source"
             },
 
-            "next": {
-              "type": "string",
-              "description": "UUID to the next waypoint, found in the waypoints list"
+            "_attr": {
+              "$ref": "#/definitions/_attr"
             },
 
-            "xte": {
-              "type": "number",
-              "description": "Cross track error",
+            "meta": {
+              "$ref": "#/definitions/meta"
+            }
+          }
+        },
+        "startTime": {
+          "type": "object",
+          "description": "The time this leg of the route was activated",
+          "properties": {
+            "value": {
+              "$ref": "#/definitions/timestamp"
+            },
+
+            "timestamp": {
+              "$ref": "#/definitions/timestamp"
+            },
+
+            "source": {
+              "$ref": "#/definitions/source"
+            },
+
+            "_attr": {
+              "$ref": "#/definitions/_attr"
+            },
+
+            "meta": {
+              "$ref": "#/definitions/meta"
+            }
+          }
+        },
+        "bearingToNextMarkTrue": {
+          "$ref": "../definitions.json#/definitions/numberValue",
+          "description": "The current bearing to the next mark relative to true North",
+          "units": "rad"
+        },
+        "bearingToNextMarkMagnetic": {
+          "$ref": "../definitions.json#/definitions/numberValue",
+          "description": "The current bearing to the next mark relative to magnetic North",
+          "units": "rad"
+        },
+        "distanceToNextMark": {
+          "$ref": "../definitions.json#/definitions/numberValue",
+          "description": "The current distance to the next mark",
+          "units": "m"
+        },
+        "bearingDirectTrue": {
+          "$ref": "../definitions.json#/definitions/numberValue",
+          "description": "The bearing relative to true North from last mark to the next mark",
+          "units": "rad"
+        },
+        "bearingDirectMagnetic": {
+          "$ref": "../definitions.json#/definitions/numberValue",
+          "description": "The bearing relative to magnetic North from last mark to the next mark",
+          "units": "rad"
+        },
+        "greatCircle": {
+          "type": "object",
+          "title": "Great Circle data",
+          "description": "Data based on great circle calculations",
+          "properties": {
+            "distanceToNextMark": {
+              "$ref": "../definitions.json#/definitions/numberValue",
+              "description": "The current distance to the next mark (Great Circle)",
               "units": "m"
+            },
+            "bearingToNextMarkTrue": {
+              "$ref": "../definitions.json#/definitions/numberValue",
+              "description": "The current bearing to the next mark relative to true North (Great Circle)",
+              "units": "rad"
+            },
+            "bearingToNextMarkMagnetic": {
+              "$ref": "../definitions.json#/definitions/numberValue",
+              "description": "The current bearing to the next mark relative to magnetic North (Great Circle)",
+              "units": "rad"
+            },
+            "bearingDirectTrue": {
+              "$ref": "../definitions.json#/definitions/numberValue",
+              "description": "The bearing relative to true North from last mark to the next mark (Great Circle)",
+              "units": "rad"
+            },
+            "bearingDirectMagnetic": {
+              "$ref": "../definitions.json#/definitions/numberValue",
+              "description": "The bearing relative to magnetic North from last mark to the next mark (Great Circle)",
+              "units": "rad"
             }
           }
         }
       }
     },
-
     "magneticVariation": {
       "$ref": "../definitions.json#/definitions/numberValue",
       "description": "The magnetic variation (declination) at the current position",
       "units": "rad"
     },
-
     "destination": {
       "title": "destination",
       "description": "The intended destination of this trip",
@@ -165,24 +225,20 @@
         "eta": {
           "$ref": "../definitions.json#/definitions/timestamp"
         },
-
         "source": {
           "description": "Source of this data",
           "$ref": "../definitions.json#/definitions/source"
         },
-
         "timestamp": {
           "description": "timestamp of the last update to this data",
           "$ref": "../definitions.json#/definitions/timestamp"
         },
-
         "waypoint": {
           "description": "UUID of destination waypoint",
           "type": "string"
         }
       }
     },
-
     "gnss": {
       "type": "object",
       "title": "gnss",
@@ -192,12 +248,10 @@
           "description": "Source of this data",
           "$ref": "../definitions.json#/definitions/source"
         },
-
         "timestamp": {
           "description": "timestamp of the last update to this data",
           "$ref": "../definitions.json#/definitions/timestamp"
         },
-
         "methodQuality": {
           "type": "object",
           "description": "Quality of the satellite fix",
@@ -219,7 +273,6 @@
             }
           }
         },
-
         "integrity": {
           "type": "object",
           "description": "Integrity of the satellite fix",
@@ -235,65 +288,52 @@
             }
           }
         },
-
         "satellites": {
           "$ref": "../definitions.json#/definitions/numberValue",
           "description": "Number of satellites"
         },
-
         "antennaAltitude": {
           "$ref": "../definitions.json#/definitions/numberValue",
           "description": "Altitude of antenna",
           "units": "m"
         },
-
         "horizontalDilution": {
           "$ref": "../definitions.json#/definitions/numberValue",
           "description": "Horizontal Dilution of Precision"
         },
-
         "positionDilution": {
           "$ref": "../definitions.json#/definitions/numberValue",
           "description": "Positional Dilution of Precision"
         },
-
         "geoidalSeparation": {
           "$ref": "../definitions.json#/definitions/numberValue",
           "description": "Difference between WGS84 earth ellipsoid and mean sea level"
         },
-
         "differentialAge": {
           "$ref": "../definitions.json#/definitions/numberValue",
           "description": "Age of DGPS data",
           "units": "s"
         },
-
         "differentialReference": {
           "$ref": "../definitions.json#/definitions/numberValue",
           "description": "ID of DGPS base station"
         }
       }
     },
-
     "headingMagnetic": {
       "$ref": "../definitions.json#/definitions/numberValue",
       "description": "Current magnetic heading of the vessels",
       "units": "rad"
     },
-
     "headingTrue": {
       "$ref": "../definitions.json#/definitions/numberValue",
       "description": "The current true heading of the vessel",
       "units": "rad"
     },
-
     "position": {
       "description": "The position of the vessel in 3 dimensions",
       "$ref": "../definitions.json#/definitions/position"
     },
-
-
-
     "attitude": {
       "type": "object",
       "title": "Attitude",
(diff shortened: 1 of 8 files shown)


tkurki:
Multiple values and the restructured source ($source) is now merged, so at least `source`s need fixing.

schemas/groups/navigation.json
```diff
@@ -36,153 +38,188 @@
...
+        "xte": {
+          "$ref": "../definitions.json#/definitions/numberValue",
+          "description": "Cross track error",
           "units": "m"
         },
-
-        "bearingDirect": {
-          "type": "number",
-          "description": "The bearing relative to true North from last waypoint to the next waypoint",
+        "eta": {
```
tkurki: We have mostly everything in verbose format, I think this and xte above should be written out as well.

tkurki:
How about instead of `bearingToNextMarkMagnetic` and `distanceToNextMark` have `mark.distance` and `mark.bearingMagnetic`?

rob42:
Yes good idea - I wasnt very happy about the length of `bearingToNextMarkMagnetic` anyway. Will add tomorrow

sumps:
I would recommend using Waypoint rather than Mark, which is traditionally the generic description (at least in Chart Plotter Land) of any single point position stored by the end user to identify a location of interest. A Waypoint is a specific mark that you are navigating to and wish to reach or pass, it could be a temporary single GOTO location, an individual stored Mark or part of a route. The third type of Mark is an Event and this is a location where the boat actually sailed over and at which something happened or the user wished to record.

canboat:
I would suggest Trackpoint or something like that instead of Event. Or Waypoint.Event. But not just Event — too vague and has connotations of other things to me, like “Boat Sinking” or “Tank Full”.

sumps:
Sorry @canboat I think you misunderstood my intent. I was just giving some background on what I believe the most common de-facto terms are for some of these things.
I would use the term Waypoint rather than Mark in this Routing/Course situation as it is a specific sort of Mark that you are trying to reach.
We may or may not decide to support the storage of Events and Marks in Signal K but I think this is outside of the scope for V1.

canboat:
Sorry from me as well, I wasn’t very clear either.

I agree that these should be Waypoints not Marks.

tkurki:
FWIW I think the keyword here is also *next*. `nextWaypoint.distance` etc.

rob42:
To clarify I have two types of points in this design. A route is made up of legs, each between two waypoints. The track between the waypoints is defined by a GeoJson lineString, so its not necessarily straight, it may have many turn points. 

These turning points are what Ive call 'marks'. eg
Route>WaypointA(mark0)>mark1>mark2>WaypointB(mark3)>WaypointC(mark0)>WaypointD(mark1)=Destination

rob42:
It allows complex routes and converting tracks to routes without creating hundreds of waypoints.
Plus you can adjust the leg marks without rewriting routes or waypoints

rob42:
So the simplified structure is now:
```
navigation.course.lastWaypoint.* (position and arrivalTime)
navigation.course.nextWaypoint.* (position and misc)
navigation.course.geometry the GeoJson linestring, by default two points [last,next] waypoint lon/lat, but could be many.
navigation.course.lastMark.* data on last mark passed
navigation.course.nextMark.* data to next mark, bearings, distance, vmg,xte etc
navigation.course.nextMark.greatCircle.* same by great circle calcs
```

tkurki:
Sorry if I'm a bit slow, but how and when is Waypoint different from Mark?

sumps:
Much as I think SK should embrace new ways of doing things, I am a little worried that this is too big a change to the industry standard way of navigating routes and waypoints. How will this GeoJSON polyline work in practice and in particular when calculating XTE for driving an autopilot ?

For V1 I would prefer to see us support a more traditional route and WP navigation structure but use GeoJSON to store and display the vessels track, so that we can look at adding route to track functionality in a later version.

rob42:
Waypoints are same as we currently understand. They are recorded and saved in the model etc. Marks = the position points in the GeoJson linestring and are recorded only as a geojson string.

The simplest path between two waypoints is a straight line, hence the geojson is a straight line, [mark0, mark1]. start waypoint = mark0, endWaypoint=mark1. eta, xte etc all work exactly as with current waypoint to waypoint nav.

But I have a number of routes with many turns, eg our harbour entrance, so in traditional systems I have to record 6 waypoints. You end up with hundreds of useless waypoints. In the design above I just record a geoJson linestring with 6 marks. I can edit the linestring (marks 1-5) to suit without moving waypoints.

sumps:
I think GeoJSON would be a great way to store tracks and a whole host of "Mariners Objects" like anchorage areas, nogo areas, reefs, difficult approaches, etc. in fact and GIS type data that boat owners might want in the future. However, the use of GeoJSON should be reserved for storage of this type of data in the Resources group and not for storing the current navigation data to steer to the next waypoint, which from a safety point of view needs to be stored and processed in a particular way so that a simple autopilot can steer a course to it or perhaps an even simpler helmsmen can interpret the data to hand steer to the next waypoint.

I would love to provide boaters with a means to steer in to a difficult anchorage, by storing the track and creating a route from it, but the reality is that this needs very careful handling as you cannot just feed a polyline in to an autopilot or present it to a user on their mast head display or chart plotter and safely follow it.

Since the first Decca/Loran/GPS units came on the market marine electronics has focused on steering from one WP (Origin WP) to the next (Next/Destination WP), maintaining the vessels position as close as possible to the ideal straight line track (BOD) between the waypoints (XTE). GPS units, Chart Plotters, Autopilots, Instruments, Navigation Repeaters, Radars, etc. and N0183 and N2000 have all been developed based on this assumption.

I am happy for the SK schema to support GeoJSON, with the view of implementing new and interesting ways to steer and navigate a boat, but to depart completely from this industry accepted method of WP navigation is too big a step to make in one go, in my opinion.

rob42:
The change is not that big really. If you never add a lineString by default you get a straight line from waypoint to waypoint as you do now. If there is a lineString, then to convert back to current systems you simply save each point in the linestring as an additional waypoint, and steer straight lines between. Think of the linestring marks as a poormans waypoint with just the most basic data [lon,lat]

The linestring basically avoids storing lots of trivial waypoints for every turn. and its really useful to convert a track to a route as they are both linestrings

sumps:
@rob42 there must be a way to implement both the traditional WP to WP (straight line) method and the geoJSON method, so that new systems can adopt the new and exciting geoJSON whilst still supporting the safe and proven WP-WP so that we can work with current autopilot and navigation systems. I would like to use the following for the traditional methods as we are coding them this weekend (!) and if you can think how to add what is required for geoJSON, then we can approve and close the PR next week.

image

sumps:
geoJSON would be a great way to support Great Circle routes

rob42:
The traditional method is supported by default. If you setup a course from waypoint x to waypoint y, and dont provide a geoJSON linestring then the default geoJson line string is mark0 to mark1, eg in this case mark0 is alias for waypoint x, and mark1 is waypoint y. All the traditional xte etc is just recorded on the mark0-mark1 level rather than on waypoint level.

rob42:
An example of a traditional waypoint to waypoint entry:
```javascript
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.lastWaypoint.position.altitude
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.lastWaypoint.position.latitude
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.lastWaypoint.position.longitude

# lastMark.index =0, is an alias for lastWaypoint
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.lastMark.index.value

vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.nextWaypoint.position.altitude
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.nextWaypoint.position.latitude
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.nextWaypoint.position.longitude
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.nextWaypoint.timeToGo.value

# nextMark.index =1, is an alias for nextWaypoint
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.nextMark.index.value
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.nextMark.timeToGo.value
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.nextMark.vmg.value
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.nextMark.xte.value
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.nextMark.bearingDirectMagnetic.value
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.nextMark.bearingDirectTrue.value
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.nextMark.bearingMagnetic.value
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.nextMark.bearingTrue.value
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.nextMark.courseRequiredMagnetic.value
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.nextMark.courseRequiredTrue.value
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.nextMark.distance.value
#if we are using great circle populate here
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.nextMark.greatCircle.bearingDirectMagnetic.value
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.nextMark.greatCircle.bearingDirectTrue.value
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.nextMark.greatCircle.bearingMagnetic.value
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.nextMark.greatCircle.bearingTrue.value
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.nextMark.greatCircle.courseRequiredMagnetic.value
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.nextMark.greatCircle.courseRequiredTrue.value
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.nextMark.greatCircle.distance.value
```

rob42:
oops timeToGo is duplicated.

So in this PR Ive renamed activeRoute > course, which is defined as the segment between the two waypoints. That might be part of a greater route of many waypoints, or just from 'here'  to some temporary chartplotter 'goto'. 

If we use the geoJson capabilities, we add a geoJson linestring as follows:

```
vessels.urn:mrn:signalk:uuid:e09e5a1c-150f-42ea-9110-63b37b64ef83.navigation.course.geometry
```
Assuming the linestring has multiple points, or marks, then the lastMark/nextMark iterates through tem till it reaches the lastWaypoint. eg for a linestring with  3 points we get:
```
lastWaypoint = mark.index=0
    mark.index0 > mark.index=1
    mark.index=1 > mark.index=2 = nextWaypoint
```

rob42:
Hmm, just a thought but in 2016 why not make great circle the default? Even an RPi has enough CPU to do it easily, and for small distances the error is tiny anyway, so its possible to just skip the calc for distances below say 120NM for efficiency...that would tidy up the whole section quite a bit, and avoid lots of if/else stuff in code all over the server and gui?

timmathews:
I'm confused as to why we care? Signal K is a _transport_ mechanism. What does it matter to us if routing software is generating routes with rhumb lines or great circles?

rob42:
Signalk is not just a transport mechanism. It also defines the meaning of the data that it transports, hence the use of SI units, and WGS84 grid refs. So if we send a `distance` we specify it will be in SI units (meters), but if its a navigational distance then its on the face of a sphere. Then we have direct (rhumbline) or great circle calculations, and the need to specify which.

Elsewhere we have avoided this by just using one definition, so I think we should apply that here too, and always use great circle. 

However rhumbline and great circle distances and bearings dont really diverge until 100NM+, so it is still pretty simple for stuff like crossing the bay, or anchor watch as the error is below the general error of GPS or any practical use.

sumps:
I am happy with the activeRoute = Course and I think using great circle throughout is a good idea as long as we make this very clear, perhaps a Signal K Navigation Overview white paper, which I could put together to cover datums, use of WGS84, different keys and what they refer to.

The bit I am still not happy with is how we implement the geoJSON. I want to include geoJSON in Signal K but it has to be supported alongside a traditional next waypoint method otherwise Signal K will never be able to confidently drive an autopilot.

tkurki:
There seems to be a bit of misunderstanding here - or possibly I'm misunderstanding something.

A GeoJSON LineString is _an array of two or more positions_. It is just another way to store a route: instead of an ordered collection of waypoint ids (the traditional method) it stores just the coordinates, without creating standalone waypoints in the data model.

So whichever way you store the route something will take as input that data, direction you want to navigate the route and current position and calculate as output the required course.

tkurki:
How about "steer to true wind angle" functionality, now courseRequired* are under mark, should they be one level up under course?

sumps:
From the server/gateway point of view, there is no point having a "Steer to wind angle" setting as this cannot be communicated to the autopilot and it will be up to the user to manually select "Steer to Wind" on the autopilot. It is apparent Wind that the autopilot uses and as long as this is being output by the server/gateway, the user will be able to select "Steer to Wind" on the pilot.

When sailing in "Nav Mode" the autopilot needs bearing from last (origin) to next waypoint, this is often referred to as BOD, which dates back to one of the earlier NMEA0183 sentences, XTE and bearing from current position to next waypoint (BTW). Also it needs to know when the next waypoint has been reached arrival radius alarm or arrival perpendicular alarm.

The autopilot will constantly try and steer the boat along the BOD course as this has been entered by the user as the safe course from A to B. When it arrives at the next waypoint by entering the set arrival radius or crossing the imaginary perpendicular line, it will set the arrival alarm bits in the NMEA message for a short period and start outputting data for the next waypoint.

The autopilot should indicate that the waypoint has been arrived and either ask the user if it is safe to make the course change or some pilots will automatically make the course change if the change is less than a certain number of degrees.

My concern is that in order to get the autopilot to steer to a more complex geoJSON track, we will have many more changes of BOD and Waypoints and this will potentially not be a good user experience if they have to confirm every geoJSON Mark course change. 

The fact is that no one knows how every autopilot will behave if presented with geoJSON derived navigation data and we would have to do a lot of testing to find out. For this reason, I think we should focus on the more traditional method of WP to WP navigation for V1 but look at supporting geoJSON within Signal K and doing some autopilot tests with interested end users.

tkurki:
Ah silly of me, of course steer to wind just uses wind.

tkurki:
How complex the desired course is is not related to it being stored/transmitted as a collection of waypoints or as GeoJSON LineString.

sumps:
This issue has stalled and we really need to get our heads together and get this important PR merged. We need to find a way to have the traditional Waypoint navigation and GeoJSON side by side in the V1 Release.

@rob42 what if we turn this around and have all of the traditional ETA, VMG, XTE, BRG, DST, etc. related to the Next Waypoint, so that we retain the traditional and accepted navigational methods and terms, but then we have the GeoJSON Line String with its Mark1, Mark2, Mark3, etc. and for any server/complex consumer that wishes to support GeoJSON, they simple read the GeoJSON Line String and use this to drive the Last and Next Waypoint objects i.e copy current position to lastWaypoint and Mark1 to nextWaypoint, then when Mark1 is reached, it becomes lastWaypoint and Mark2 is copied to nextWaypoint, then when Mark2 is reached it becomes lastWaypoint and Mark3 becomes nextWaypoint, etc. etc.

This seems to me to be a good way to retain traditional and well proven methods that we know work with the current and past generation of marine electronics, but at the same time introduce the new and exciting GeoJSON feature.

We also need to have the new Arrival Radius and Perpendicular fields to tell an autopilot that we have arrived at the next waypoint.

sumps:
Thinking more about this, it makes sense to store tracks and complex routes in the "Resources" group as GeoJSON Line Strings rather than in the dynamic "Navigation" group. The GeoJSON Line String can then be selected and loaded by the server/consumer and passed into the "Navigation" group a leg at a time.

fabdrol:
I had a chance today to read up on this issue. To be frank, I don't see what the problem is apart from language used? When you navigate from one waypoint to another you want to send start and end coordinates to your tiller pilot. Whether those coordinates come from a list of waypoints or GeoJSON coordinates ("marks", in Rob's post), doesn't really matter IMHO.

That said, I think GeoJSON is a good way to encode this data in the model. A more traditional approach would cost flexibility and wouldn't gain anything from the p.o.v. of the data model. When it comes to generating data for auto pilots a translation is necessary anyway (e.g. NMEA0183, SeaTalk or N2K). Encoding a GeoJSON coordinate (or "mark" in Rob's terminology) as a "waypoint" would during this translation step is trivial. 

To sum up, +1 on this PR.

sumps:
Maybe I am not explaining myself very well, so let me start again. Different users have different ways of navigating via an autopilot; 

1)  Manually bring the boat on to the required course and hit AUTO on autopilot
2)  Select one point on the chart, hit GOTO button on chart plotter and then NAV button on autopilot
3)  Build a route from a list of Waypoints that have been built-up and used over a long period of time, select the nearest one, hit GOTO button on chart plotter and then NAV button on autopilot
4)  Enter a route on the chart plotter, select first waypoint, hit GOTO button on chart plotter and then then NAV button on autopilot 

Focusing on methods 2-4, the autopilot needs to know the ideal track (bearing from origin to next waypoint), how far off course the boat is (cross track error) and the current bearing to the next waypoint.

autopilot apb

The autopilot also needs to know when the next waypoint has been reached (Arrival Radius or Perpendicular Alarm) so that when the Navigator (GPS, Chart Plotter, etc.) switches to the new waypoint, the autopilot can ask the user to confirm the course change, which in many cases requires a manual button press.

As well as the autopilot, many other pieces of equipment such as 20:20 Mast displays, MFDs, Radars, Instrument Repeaters, etc. all repeat the next waypoint type data and we need to be sure that the data we send out from a Signal K system will be received and handled correctly by these devices. 

Maybe not for V1, but very soon you will be wanting Signal K to drive an autopilot and I have concerns that this new GeoJSON method will immediately work with the traditional equipment that is already onboard boats.

Let me try and illustrate my concerns....the diagram below shows a traditional route on the left and the same route in GeoJSON on the right....

geojson v trad
  
I like the way that the GeoJSON allows us to encode complex routes, created from tracks or to add extra waypoints to an existing route, but what happens when you start sending this data to an autopilot. If the resolution of the GeoJSON line string is higher than the Arrival Radius, then you will potentially have two or more waypoints within the Arrival Circle and I have no idea how to handle this and I am sure the autopilot will have no better idea. You could artificially change/stretch the resolution of the GeoJSON to create a series of longer straighter legs, but you will then be taking the boat off the original track and possibly in to dangerous waters. 

Some autopilots will automatically change course to the next waypoint if the course change is less than a certain number of degrees, I think latest Navico units will do an auto change if <20º, but I do not know how every other autopilot will behave and if I was a user, I would soon get hacked off pressing the confirm course change button every minute or less.

I think we should go with the more traditional approach and wording of the Signal K navigation.course objects but implement GeoJSON in the Resources Group so at least servers and consumers can start playing with these new objects, even doing calculations and passing data to the navigation group. So I am not saying don't use GeoJSON, just don't jump 100% into using it as the only or preferred method in V1, until we have more real life experience.

tkurki:
@sumps: 

The number or closeness of the waypoints has nothing to do with the format they are stored in. 

The route  on the left in your illustration can be stored as GeoJSON and the route on the right as waypoints.  

The problems that you point out regarding autopilots are equally valid for waypoint-based routes.

fabdrol:
@sumps: your concern about wether it will work with existing devices won't be an issue: for existing devices a SK server has to translate the routing data - regardless of the internal format - into a traditional format anyway (N2K or 0183). 

Regarding the other issue: I'm not sure that would be an issue at all. GeoJSON doesn't require a complex route like the one on the right in your picture (as Teppo points out, I was too slow!). What @rob42 proposes, the way I read it, is not that different; i.e. the image on the left can be encoded that way perfectly fine in a GeoJSON LineString. The only difference is that in Rob's proposed version this route could be created whilst only saving WP1 and WP4 as waypoints, whilst saving WP2 and WP3 as marks. 
Think of it this way: WP1 would be my marina and WP4 could be my destination. The points in between are not actual geographical places - they're just "virtual" points on a map that I need to get to WP4. Having two categories of "waypoints" kinda makes sense to me.

sumps:
Yes I agree, but traditional routes/waypoints are entered manually by the user, who has also entered the Arrival Radius and generally has used the least number of waypoints possible (longest legs) because we are human. 

One of the main reasons put forward for using GeoJSON was to allow recorded tracks to be converted to complex routes. As soon as you have this automatic creation of routes, I start to have concerns.

A typical track consists of recorded points, that are either based on distance or time. Typical values will be 0.01NM or every 10secs and so the resolution can be pretty high, which is what you want if you are recording a difficult approach. 

Like I keep saying, I want to support GeoJSON in Signal K, I just have concerns over this aspect of autopilot navigation, the naming of the navigation.course objects and that  Line Strings cannot store user entered waypoint names, so we will be limited to simple R01-001, R01-002, etc. names.

fabdrol:
I don't dispute that there might be issues when one is converting a track to a route. However, that's a separate issue and not related to how we store the route internally (i.e. even if we'd store route points as simple waypoints, translating a route into a course would require the software to pick and choose out of all the tracked points). 

Regarding naming: if I understand correctly, @rob42's proposed solution works like this: a waypoint has a "mark" (i.e., it's position). A "mark", on the other hand, doesn't need to have a waypoint. When sending route data to the autopilot, it will receive a list of "waypoints", some of which will have user defined names (the waypoints) and some will have generated names (the "marks"). 

Regarding "arrival radius".. To be honest (in my limited experience) I have never had to input that - seems very old-fashioned to me. Seems to me like something the software could determine very easily without bugging the user, or am I missing something obvious?

sumps:
@fabdrol the whole reason I am concerned, is because I am in the middle of designing a Signal K to NMEA gateway. You cannot expect the gateway/server to control an autopilot correctly if we have not included all of the necessary data in the right format within the Signal K data model. 

GeoJSON is a storage mechanism for geographical data and should not change the way we organise the navigation.source objects. These should be chosen to safely communicate with NMEA devices and to present users with clear and unambiguous data that they can understand and make decisions on.

I don't want to mix up marks and waypoints and use new and non-standard terms and methods just because that fits in with GeoJSON. Make GeoJSON work for Signal K and safe navigation, not the other way round.

fabdrol:
@sumps let's talk about the missing data: which data points required for auto-pilots are missing in the proposed format?

canboat:
I’m with Paul here.

fabdrol:
@rob42 aside from this discussion, I do have a few points regarding the actual contents of the file: there are a few abbreviations in there (`vmg`, `xte`, `eta` etc). Shouldn't these be written out as per conventions?

tkurki:
Based on discussions on Slack: The real point is being able to map traditional autopilot related data such as APB and RMB sentences to the Signal K model and being able to generate them from data in the Signal K data mode, generated by a "navigator entity". The discussion has been about "to GeoJSON or not", which is related but not directly connected to having these data items in the SK model.

The data we need to represent is
- heading to steer true/magnetic
- bearing to next waypoint
- bearing origin/last waypoint to next waypoint 
- perpendicular passed at waypoint
- arrival circle entered => needs arrival circle radius
- XTE
- direction to steer
- destination closing velocity (velocity made good)
- range to destination
- next waypoint position
- previous/next waypoint id

Please see https://github.com/SignalK/specification/blob/626bc5de00cc765e141c150f552a6d6903c023dc/test/data/course-delta.json for a sample of what this PR actually has.

todo
- [ ] review the mapping between the list above and the schema, add missing data (like arrival radius)
- [ ] review the terms: discussion above prefers waypoint over mark, something else needs changing?
- [ ] remove abbreviations

rob42:
@sumps I see what you mean about a simple track conversion, where the track has a point every 10M or so. I hadnt imagined that as the main use case for an autopilot

In my mind the there were two main use cases:
1) I start at my waypoint outside Nelson  harbour and set a goal of French Pass. Both are waypoints. But to travel between them I need a couple of kinks in my course to avoid shallows, rocks, etc. Using geojson I can set a simple course between start and end, and drag out the kinks to suit without random waypoints cluttering my system. 
2) I have a successful track into some harbour, and I want to save it as a route. Mainly so I can follow it next time, and exit in poor visibility if needed. Its not going to be by autopilot, it will be followed manually. However I do expect to simplify the track down to major segments, not keep every one. I already use a simplify algorithm in the java server to avoid 20,000 points on a nearly straight line.

As the others have pointed out the problem of numerous very close waypoints is not a geojson problem. Its actually related to the autopilot control having a problem with arrival circle < distance to next waypoint.

The other issue is what to call the current leg, waypoint-waypoint, or mark-mark. I mean, are there multiple marks between waypoints, or do we copy a mark to a waypoint and next mark to nextWaypoint.

From a program point of view its just a name, but in navigation a waypoint is a well known location with name, and other attributes. Its stored and listed and reused. If we copy geojson marks to waypoints and navigate between waypoints then we mix that up, we have lots of temporary waypoints, and some permanent ones. 

Hence I used the idea of marks. Waypoints are permanent, and marks are temporary, literally just pencil marks on the chart.

rob42:
Added arrivalRadius and removed all greatCircle keys as all values are great circle in signalk now.
@fabdrol I didnt extend eta, xte, and vmg as they are such well known terms, not sure they need changing.

fabdrol:
The "alarms" for arriving wether the vessel is in the arrivalRadius (flags in the 0183 sentences) should be in the group `notifications`, as these will only be present for a short period (until the vessel sails on) and as "notifications" should be acted upon according to their type. I'm working on an example that should work for the mentioned use cases. Will post that in the course of the day (or night, for those far away!)

sumps:
RED FLAG - @rob42 the decision to make Signal K Great Circle only will have massive affects on all App developers and Gateway/Server developers. In my opinion this is too big a jump and despite the fact that most modern systems have the floating point capability to do these calculations without too much of a performance hit, for existing apps that want to support Signal K, we have just forced them to do a lot of extra code development, app restructuring, etc. Also for iKommunicate, CanBoat, Node Server, Java Server and other future gateway/servers we have now forced these devices and libraries to support Great Circle.

As an example, pretty much the majority of NMEA data is in Rhumbline and a gateway like iKommunicate will have to convert all bearings and distances it receives in to Great Circle which will definitely have a performance impact to say nothing of the extra work involved at this eleventh hour. 

I would please request that we support RL and GC bearings and distances in the Signal K specification.

rob42:
It has a negligible effect on any distance short of 200NM, and only as you go beyond 40deg lat. 

From way back above:
"Just did some maths, greatest variation is travelling W/E, no variation N/S. No difference at the equator. At 50deg N, travelling W, its about 100M shorter between 2 points 200NM apart. Ideal direction is a curve starting 88deg and ending at 92deg, rather than direct at 90deg.

Even Hobart-Nelson (1200NM ENE) its only 5NM difference."

So basically any distance less than 200NM the difference will be so small that it can be ignored and straight line calcs used.

sumps:
I understand the accuracy implications but can we really mix values like this ? If a gateway or server receives a RL bearing should it still copy it in to the Signal K bearing ? We really do not want to be checking distances and deciding whether it is safe to use a RL bearing instead of a GC bearing, would it not just be a better solution to have fields for RL values and fields for GC values ?

canboat:
I agree with Paul here.

Met vriendelijke groeten,
Kees

tkurki:
Superceded by #211 and #218, #219.

PR #169, opened 2016-02-09
Inconsistency in schema and example

joabakk:


tkurki:
Inconsistency resolved by 3e0f448f91924b1f5ad54b44a3a72b026b3796cb

joabakk:
Although this is well hidden for the average boater I would prefer rpm to revolutions, as this i a fairly common standard, but understand if that discussion has been closed. Mocha probably fails now as I used .rpm in the nmea conversion, but I will check. Thanks again Teppo

tkurki:
It is revolutions because the unit is not revolutions per minute but Herz or revolutions per second, an SI unit.

... and I just realised that RPM sentence conversion is not doing the conversion to Hz.

tkurki:
https://github.com/SignalK/signalk-parser-nmea0183/commit/f721eef65723f2afb2b6c7701c876ff7129a9eaa

PR #175, opened 2016-02-16
Added agnostic target speed and target angle

joabakk:
Some instruments and equipment will give an NMEA value as target boat speed. Meaning if you are sailing too close hauled, you have lost speed and need to fall off and vice versa. This is taken from the polar diagram, but does not distinguish between beat or gybe.

### DIFF
--- a/schemas/groups/performance.json
+++ b/schemas/groups/performance.json
@@ -64,6 +64,18 @@
       "$ref": "../definitions.json#/definitions/numberValue",
       "units": "m/s"
     },
+    
+    "targetAngle": {
+      "description": "The true wind gybe or beat angle for the best velocity made good downwind based on current polar and trueWindSpeed.",
+      "$ref": "../definitions.json#/definitions/numberValue",
+      "units": "rad"
+    },    
+    
+    "targetSpeed": {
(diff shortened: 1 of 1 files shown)


schemas/groups/performance.json
```diff
@@ -64,6 +64,18 @@
       "$ref": "../definitions.json#/definitions/numberValue",
       "units": "m/s"
     },
+    
+    "targetAngle": {
+      "description": "The true wind gybe or beat angle for the best velocity made good downwind based on current polar and trueWindSpeed.",
```
tkurki: A bit inconsistent, should probably say down/upwind and not just downwind.

And isn't current polar really a function of trueWindSpeed, should the polar mention be omitted?

joabakk:
Absolutely, it should be up or downwind. What I meant by polar was the polar diagram. The polar diagram will give you the polar speed, ie the best speed for your current heading, and the target speed (tack/beat or reach/gybe best vmg angle and speed)

tkurki:
Could you please update the description, otherwise I'm good with merging this.

joabakk:
Something like this?

PR #178, opened 2016-02-18
Highlighted Altitude in 3D Position Fix

sumps:
The specification mentioned a 3D position but did not mention Altitude units.

Need to make a decision on Datums and Geoidal Separation (Altitude adjustment) for Signal K.


tkurki:
All units are in the schema explicitly in `units`, not in description.

```
          "altitude": {
            "type": "number",
            "description": "Altitude",
            "units": "m"
          }
```

PR #180, opened 2016-02-21
add unitOfMeasure to sensors schema

damianham:
There should probably be a unit of measure for the sensor data so that the sensorData element can contain a simple value e.g. thinking about an XDR sentence

$YXXDR,C,7,C*54

C == sensorType (temperature)
7 ==  sensorData
C == unitOfMeasure (Celcius)

### DIFF
--- a/schemas/groups/sensors.json
+++ b/schemas/groups/sensors.json
@@ -19,6 +19,11 @@
       "type": "string",
       "description": "The data of the sensor data. FIXME - need to ref the definitions of sensor types"
     },
+		
+		"unitOfMeasure": {
+			"type": "string",
+			"description": "The unit of measurement for the sensor data"
+		},
 
     "fromBow": {
       "$ref": "../definitions.json#/definitions/numberValue",


tkurki:
The way to make sensor data from XDR or for example I2C connected sensors useful in the SK universe is to map & convert the data to the normal Signal K data model. In the SK model it must have the correct unit. In this way the unit of the sensor data is meaningless - a client should have very little interest in the original unit that the sensor outputs, like it doesn't care if the sensor is outputting NMEA or raw binary data.

Driving the conversion is another matter. If we think of the sensors section as the configuration data for XDR conversion what do we need?
- id of the sensor that the server can use to match to input: "nmea1.YXX" 
- SK path: propulsion.x.oilTemperature
- sensor units: C(entigrade)
and with this we could configure the XDR conversion code.

Ping @joabakk @rob42.

tkurki:
To be clear: I'm not against adding a unit field, I'd just like to understand its utility a bit more.

damianham:
Ah ok, agreed. Ignore that pull request then

PR #184, opened 2016-02-26
Update Vessel Schema to re-structure the Electrical Group

sumps:
Have one Electrical Group (Map) that categorises Electrical Devices by type. References a /group/electrical.json that needs to be created in another PR


tkurki:
PRs should contribute a meaningful whole - making nontrivial chacnges piecemeal with a PR for each change is not good practice.

Issue #183, opened 2016-02-26
Single Electrical Object


sumps:
The Electrical Group is probably the group with the most amount of work still to do for V1 of the Signal K Specification release. Currently it has two objects DC and AC, but the reality is that we can probably abstract everything down to one Signal K Electrical Object. There are many electrical objects on the boat, some AC, some DC and some that have both types i.e. Inverter or Battery Charger. We simply populate the fields that the particular electrical object has.

If the Electrical Group was a map like Propulsion, Tanks, etc. we could have the following structure;
-  UUID (regex)
-  Type (Enum; Battery, Solar Panel, Wind Generator, Water Generator, Inverter, Battery Charger, Converter, Generator, Bus, Mains (Shore), Combi, etc.
-  DC Voltage
-  DC Current
-  Case Temperature
-  Ripple Voltage
-  State of Charge
-  State of Health
-  Time Remaining
-  Amp Hours
-  AC Phase A
  > Voltage L-N
  > Voltage L-L
  > Current
  > Power
  > Frequency
+  AC Phase B
+  AC Phase C  

We have to finalise the Electrical Group ASAP, so comments please.....

sumps:
OK going to raise a PR on Electrical Group but will be slightly different to the above. After discussions with Teppo, I will categorise the Electrical Devices slightly, so electrical/battery/uuid/voltage or electrical/generator/uuid/acPhaseCurrent

PR #185, opened 2016-02-29
Reorganise electrical: remove ac/dc distinction, branches for equipment types

tkurki:
This PR organises electrical schemas so that
- there is just one electrical schema, no ac/dc separation
- schema has definitions for [dcQuantities](https://github.com/SignalK/specification/blob/88499feef8da0ca334980f582b0ba3f553373028/schemas/groups/electrical.json#L52) and [acQuantities](https://github.com/SignalK/specification/blob/88499feef8da0ca334980f582b0ba3f553373028/schemas/groups/electrical.json#L180)
- equipment types are branches with keyed instances
- equipment types are composed by adding ac, dc or both quantities and adding any additional properties such as [battery chemistry](https://github.com/SignalK/specification/blob/88499feef8da0ca334980f582b0ba3f553373028/schemas/groups/electrical.json#L262)

See also #183


tkurki:
There is stuff to do like
- [ ] add charger
- [ ] add solar panel
but should this be merged and then additional equipment added as separate, smaller PRs - are we happy with this organisation or should it be stressed with more sample configurations, the one I came up with is still pretty simplistic?

schemas/groups/electrical.json
```diff
@@ -177,32 +177,73 @@
...
+        },
+        "reactivePower": {
+          "$ref": "../definitions.json#/definitions/numberValue",
+          "description": "Reactive power",
+          "units": "W"
+        },
+        "powerFactor": {
+          "$ref": "../definitions.json#/definitions/numberValue",
+          "description": "Power factor",
+          "unit": "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX fraction"
```
tkurki: @canboat what do we use for unit here?

canboat: We have `ratio` with description 'Relative value compared to reference or normal value. 0 = 0%, 1 = 100%, 1e-3 = 1 ppt' which fits perfectly for this.

tkurki:
I think this is ripe for merging, additional equipment can be added as needed.

sumps:
In these Mapped objects types like Electrical, Tanks, etc. is there any reason why you would not call the object /battery.1 rather than as you have it at the moment /battery/1 ? From a REST API point of view would it not be better to call /electrical/ and get a list returned of battery.1, battery.2, inverter, etc rather than have an extra branch for each object ?

timmathews:
I've made a few changes, pretty minor stuff. The biggest change was using plain "number" as the type for the temperature and current limit values

Issue #188, opened 2016-03-05
Add Reef State to Sails Group


joabakk:
great work so far!

I have a use case:
I want to make the most out of my polar diagrams, to improve racing speed. Say I have sailed in similar conditions before, I want to know what sails I had set, and if they were reduced (furled or reefed)

In propulsion, we have each engine witth a state. This could also apply for each sail ( available, set and damaged/unavailable). Also, the sail area, the reduction by reefing or furling, and perhaps ven the shape, in a few years time, we might all be doing this on the cheap http://www.sailingworld.com/gear/new-science-sail-trim

tkurki:
See https://github.com/SignalK/specification/blob/master/schemas/groups/sails.json#L43

No way to set reefing status though.

joabakk:
Thanks. If this is something we want, this can either be done as number of reefs set or a ratio set (includes furling also), or simply boolean (reduced=true or false)

- number of reefs set is simple and user friendly, but does nothing for furling
- ratio is more correct, but requires a system for measuring luffs or the like, and would need a lookup table for reefable sails
- reduced boolean is simple and user friendly, and indicates that the resulting polar diagrams will be less trustworthy

PR #191, opened 2016-03-07
#188

joabakk:
Here's an attempt to model issue #188 .

### DIFF
--- a/schemas/groups/sails.json
+++ b/schemas/groups/sails.json
@@ -14,9 +14,29 @@
           "$ref": "../definitions.json#/definitions/sail" 
         }
       }
+	  "value": {
+		"type": "string",
+		"enum": [
+			"available",
+			"set",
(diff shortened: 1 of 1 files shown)


tkurki:
There is already individual sail parameters  in https://github.com/SignalK/specification/blob/master/schemas/definitions.json#L233-L299, this is overlapping.

How about going document-first - creating a sample document describing a real world situation with all the necessary data and then seeing what's missing? A sample document would demonstrate how the schema works better than the schema definition files.

Schema alterations should anyway have an accompanying test that exercises the relevant parts of the schema. A test would have also  shown that the json is missing a comma before the addition and the fact that the additions are one hierarchy level too high, they should be under patternProperties. As it is you have area of one sail at the same hierarchy level as the total area of set sail.

How about adding the reefing state to the existing sail definition?

PR #196, opened 2016-03-17
There are too many levels in Navigation

sumps:
With the changes to Active Route there was a new layer Waypoint created which is a layer too far. There were some values like Bearing to next Waypoint in the Active Route level and other values like XTE in the Waypoint level which did not make any sense. Need all of the next waypoint navigation data in the Active Route level and to get rid of the Waypoint level.

### DIFF
--- a/schemas/groups/navigation.json
+++ b/schemas/groups/navigation.json
@@ -272,6 +272,16 @@
           "$ref": "../definitions.json#/definitions/numberValue",
           "description": "Positional Dilution of Precision"
         },
+        
+        "verticalDilution": {
+          "$ref": "../definitions.json#/definitions/numberValue",
+          "description": "Vertical Dilution of Precision"
+        },
+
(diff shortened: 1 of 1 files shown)


sumps:
Just one outstanding issue in Active Route section - Arrival Alarms; radius and perpendicular. Need to decide how to do these as they are effectively logical True/False values and should trigger an alarm.

rob42:
#168 is still outstanding, which will affect this. Pls review and comment

tkurki:
This PR has some unrelated stuff (DOPs) and otherwise is based on outdated master. Please review any changes not present in current master & take action if needed.

PS. It is in general a Bad Idea to include several unrelated issues in one PR. PRs often need a review round and changes and some stuff that might be otherwise mergeable right away stall for no real reason. It is technically possible to cherry pick the changes, but it is more work for maintainers and makes PRs messy.

PR #197, opened 2016-03-18
Include data from PGN 126996 and 126998

sumps:
This is all of the Product and Configuration information.

### DIFF
--- a/schemas/groups/sources.json
+++ b/schemas/groups/sources.json
@@ -86,7 +86,7 @@
         },
         "manufacturerInfo": {
           "type": "string"
-        },
+        }
       }
     }
   }
--- a/schemas/groups/sources.json
+++ b/schemas/groups/sources.json
@@ -53,7 +53,40 @@
         },
         "type": {
           "type": "string"
-        }
+        },
+        "manufacturer": {
+          "type": "string"
+        },
+        "productFunction": {
+          "type": "number"
+        },
+        "productClass": {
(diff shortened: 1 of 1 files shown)


timmathews:
These fields could use some descriptions. How does `manufacturerInfo` differ from `manufacturer` for instance? Also, `serialNo` and `manufacturerInfo` should be spelled out. And finally, `productFunction` and `productClass` should be strings, not numbers. The numbers aren't really useful, but perhaps the actual function and class names they represent may be.

Is the version of [Appendix B.6](http://www.nmea.org/Assets/20120726%20nmea%202000%20class%20&%20function%20codes%20v%202.00.pdf) on NMEA's site current?

tkurki:
An example document with these fields filled in would be great! I can do that if somebody coughs up some real world data.

sumps:
Tim the Appendix B6 on the NMEA site is still current and it is these Function and Class codes that are extracted as numbers from the PGN so they are very important in allowing the automatic naming of a Signal K source.

Teppo, here is a screen shot from the Actisense NMEA Reader program that shows the typical info that can be extracted from these PGNs - one shows the Actisense NGT-1 and the other an AIS Transponder.

ngt-1 whown in actisense
cobalt shown in actisense

tkurki:
I refactored the fields a bit and added a sample document with associated test at #202. 

Closing this, let's continue there.

PR #198, opened 2016-03-18
Added Age of Variation Calculation

sumps:
Magnetic variation varies over time, so it is important to know the date that the magnetic variation calculation was made.


tkurki:
Unit should be seconds. Implemented in 2a551f42e76f32762b77cf16a809bbe088599b1e

PR #199, opened 2016-03-18
Tidying up Propulsion Group

sumps:
Too many temperatures in the propulsion group as Coolant and Exhaust Temp will be handled by the new Environment Temp object. Only include the specific temperatures where there are dedicated Propulsion PGN Fields; Engine Temp, Oil Temp and Trans Oil Temp, 

Fuel Rate was also repeated when it is included very well in the new Fuel Object level.

### DIFF
--- a/schemas/groups/propulsion.json
+++ b/schemas/groups/propulsion.json
@@ -35,7 +35,7 @@
       "units": "Hz"
     },
 
-    "temperature": {
+    "engineTemperature": {
       "description": "Engine temperature",
       "$ref": "../definitions.json#/definitions/numberValue",
       "units": "K"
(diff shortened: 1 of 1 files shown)


timmathews:
I think `exhaustTemperature`, `coolantTemperature` and `oilTemperature` should stay. However, I agree `waterTemperature` is a bit redundant.

sumps:
Hi Tim, it is not a case of getting rid of them, it is just a case of them not being in the propulsion group and being in the environment group under the new temperatures section instead.

The problem is that there is no dedicated PGN for Exhaust Temperature and so this sensor would be outputting the standard generic temperature PGN which appears in the Environment Group with Bait Box, Galley, Jacuzzi, etc.

sumps:
Coolant, water and engine temperature are the same thing - there is just one sensor in your average engine that measures this. The other "engine" temperature sensors are Engine Oil Temp and Transmission Oil Temp.

timmathews:
Where they come from in n2k is irrelevant. They don't belong in the environment group, they're related to the engine and belong in the engine group.

tkurki:
See https://groups.google.com/forum/#!searchin/signalk/engine$20temperature/signalk/7DCpiWnpDNM/vpqfWt45EAAJ for an example that actually goes into even more detail than the current schema. We could add in/out values as per this example.

We need not and should not limit SK data model to what N2K has, but we do need to find explicit mappings for most N2K data. I don't see extra fields as a liability, as long as there is no or little duplication.

Coolant and exhaust temperatures are not under environment.

Deleted duplicated fuelRate a1c63d61d74eb6173fb47d69882d8080e1d19159

PR #200, opened 2016-03-18
Added State of Health to Batteries

sumps:
State of Health is the overall status of the battery based on the life of the battery - 0-100%

### DIFF
--- a/schemas/groups/electrical.json
+++ b/schemas/groups/electrical.json
@@ -328,6 +328,12 @@
                     "units": "ratio"
                   },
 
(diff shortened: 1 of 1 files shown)


tkurki:
What does a value say .2 signify? Where does this value come from? Just curious.

sumps:
That the battery has 20% of life left I guess, I suspect that much under 0.5 (50%) means your battery is pretty bad. I am not sure how this value is calculated and I am sure different sensors will calculate it differently, but it is a value supported in the PGN so SK should also support it.

Issue #177, opened 2016-02-18
Datum (was: 3D Position but no Altitude?)


sumps:
We say in the spec that Position is in 3D but in all the examples I have seen it is just Lat/Lon....

"values": [
          {
            "path": "navigation.position",
            "value": {
              "source": {}, 
              "timestamp": "2016-02-17T14:1:19.000Z",  
              "longitude": 24.933841705322266,
              "latitude": 60.20331954956055 

I assume to add altitude you just put **"altitude": 123** and that the value is in meters but it would be good to confirm this and also confirm that the Altitude value stored is adjusted for Geoidal Separation or not.

tkurki:
Schema at https://github.com/SignalK/specification/blob/master/schemas/definitions.json states altitude units as meters. Technically none of triplet properties are required, latlon probably should be.

Please annotate description and post PR?

tkurki:
@timmathews where should we document the Geoidal Separation stuff in your opinion?

sumps:
And even more importantly documenting the WGS84 datum issue as this has a much greater affect on safe navigation than the altitude being wrong.

timmathews:
Have a look at this, and let me know what you all think: http://signalk.org/api/gnss-field-guide.html.
That was something I started on very early on in the SK project before we had started using JSON-schema. I don't know why that stuff never made it in to the schema. I guess I ought to add it.

As for identifying the datum, I don't think it needs to be explicitly stated. If we know what GNSS satellites we're talking to, we know what reference frame is being used.

* GPS uses WGS84
* GLONASS uses PZ-90
* Beidou uses CGCS2000
* Galileo uses GTRF

However, if you tell me that there are GNSS receivers which automatically do the conversion internally, then my opinions are subject to change.

That said, if there _are_ receivers which do this, how do they communicate the fact that they've done so? As far as I know there isn't anything in either NMEA spec which holds that information. At this point, we're probably talking about supporting manufacturer proprietary sentences or their binary interfaces instead. Not that I'm opposed to that, often you get better info from a GPS receiver if you can communicate with it via some protocol other than a NMEA one.

Where we absolutely _do_ need to record the datum is in the section of the schema where we're storing routes and waypoints. In addition to the coordinates for each waypoint, we need to store the datum that those coordinates use. And there are way more than the four I referenced above. A stupid huge number in fact.

[...]

PR #208, opened 2016-04-08
Add Datum object

sumps:
Although all Signal K positions stored in the schema should be WGS84, if information about the datum being used by the GNSS system and the Local datum are provided by an NMEA sentence or PGN, this data will be stored in this new Datum object.


tkurki:
https://github.com/SignalK/specification/issues/177#issuecomment-207372400

Issue #89, opened 2015-09-16
History / time series API


tkurki:

tkurki:
We need an API that provides ways to 
- discover what time series are available => present a list of for example dates / time ranges for which we have data
- a way to query &  summarize the data, for example

*Soundings for a date*
- September 15
- 5 minute resolution
- position (average), depth (min/max/avg)

*Depth data for an area*
-- all data within a geofence 
-- position+depth, all data

*True wind history*
- last 2 hrs
- true wind direction and speed, 1 minute average

There are undoubtedly countless others.

Here the schema is less important than the query language. For example the current delta format in an array or something similar would work pretty nicely - a collection of data items with timestamps.

rob42:
From slack discussion:
Basically we can store a list of `{"timestamp":"123456", "path":"y.z.x", "value":"nnn"}` in a persistent db. That would provide a place for history outside the signalk model, which is really the 'current' data.

Then I would propose `/signalk/v1/history/[context]/[path]?start=xxxxx&end=yyyyyy`

So as similar as possible to the REST api, but with its own url. Then we can advertise it if the server supports it. And an equivalent signalk message for other transports:

`{"context":"vessels.self","history":{"start":nnnnnnn, "end":yyyyyyyy, "path":"x.y.z"}}`

The reply should be either a series of update messages, or a list of the stored objects  
`{"timestamp":"123456", "path":"y.z.x", "value":"nnn"}`
Seems the updates would be more verbose but handy for replays etc.

tkurki:
Your `{"timestamp":"123456", "path":"y.z.x", "value":"nnn"}` look awfully lot like a delta message. I think there is value in using one format for both delta stream and history query result. Also a co moremplex use case like _position and speed for the racing fleet, 5 minute average_ would fit pretty well if not optimally to the result being a list of deltas.

I would not put context and path into the url path. How do you query for multiple values? It implies a hierarchy, which is not there. They are also somewhat orthogonal: context is part of the `where` clause and `path` a part of the `select` clause in sql terms.

tkurki:
Food for thought: http://eagleio.readthedocs.org/en/latest/reference/historic/jts.html

tkurki:
Food for thought: http://eagleio.readthedocs.org/en/latest/reference/historic/jts.html

https://github.com/james-hu/cjtsd-js/wiki/Compact-JSON-Time-Series-Data

tkurki:
[...]

PR #212, opened 2016-04-20
add time reference and make $source optional

vyacht:
1. Make $source optional 

It can be empty anyway. Not every system provides the source which also contains a lot of overhead.

2. Allowing to use a time reference to fetch "historic" data with the full format. 

There would be two options
i) use an array "historic": [{"value":3.7, "timestamp":"..."}]
ii) make a path-able reference as in this commit, this will allow for mostly everything discussed in wiki's RFC  [time reference model](https://github.com/SignalK/specification/wiki/RFC---time-reference-model)

There might be more elegant JSON schema ways to express the same thing.

I also propose to use parameters ?startAfter=<timestamp>&endBefore=<timestamp>&limit=100 in order to request historic data. This would work for GET and ws. 

The delta format has the possibility for delivering historic data since values are delivered in an array.

### DIFF
--- a/schemas/definitions.json
+++ b/schemas/definitions.json
@@ -276,9 +276,43 @@
                 "$ref": "#/definitions/valuesNumberValue"
               }
             }
+          },
+          "historic": {
+              "$ref": "#/definitions/valuesTimeValue"
+          },
+          "last24h": {
+              "$ref": "#/definitions/valuesTimeValue"
           }
         }
       }]
+    },
+
+    "valuesTimeValue": {
+      "type": "object",
+      "patternProperties": {
+        "description": "ISO-8601 (UTC) string representing date and time.",
+        ".*": {
+          "properties": {
+            "value": {
+              "type": "number"
+            }
+          }
+        },
+        "max": {
+          "properties": {
+            "value": {
+              "type": "number"
+            }
+          }
+        },
+        "min": {
+          "properties": {
+            "value": {
+              "type": "number"
+            }
+          }
+        }
+      }
     },
 
     "valuesNumberValue": {
--- a/schemas/definitions.json
+++ b/schemas/definitions.json
@@ -233,7 +233,7 @@
 
     "commonValueFields": {
       "type": "object",
-      "required": ["timestamp", "$source"],
+      "required": ["timestamp"],
       "properties": {
         "timestamp": {
           "$ref": "#/definitions/timestamp"


tkurki:
So we have two issues here: time series api (something pretty complex) and whether a field is required or not.

I think that time series api warrants an issue of its own and I created one in #89. Please bear in mind that we need not extend one single schema to cover everything: if we define the api as `.../navigation/courseOverGround/history?source=...&begin=..&end=..&period=...` the response can be an array of values & timestamps with practically no relation to the main SK schema. So let's define the api & schema separately there.

We have had discussions before about whether some things should be enums or not (closed or open set of possible values) or should some fields be required or not. The idea of the $source field is to provide a pointer to an object under `sources` that can hold more metadata. I would like to keep it as a required field to be able to validate the different SK implementations coming up. More constraints, less leeway on the how servers operate.

In what case would you see it absolutely positively not useful and required at all? 

PS. Pull Requests with several unrelated fixes place a lot more work on the maintainers of the software project. For example in this case if one commit would be ready to merge and the other needs more work/consideration a committer would need to hand pick the individual commits instead of merging them with a push of a button or the one ready for merging would have to wait for the other to be completed.

vyacht:
Agreement on your "PS". Probably need to push, pull request, push, pull request.
Agreement on array. Complete agreement on separate API. Actually I had both in a first prototype: a separate API based on the full format with a timestamp, value array. Then re-read "Arrays are Evil". While I don't agree with the sentence - the arguments were valid in the sk context. 

`server:3000/signalk/v1/vessel/244670354/navigation/headingTrue/historic/2016-04-19-23:21:65.123/value`

is not too bad either as a path. Maybe I'll push another proposal based on the full format later. I need history urgently.

Its just that I don't like the delta format. A variant might be useful but I can't come up with a good one for history:

An array `"history":[{"timestamp":"123456", "path":"y.z.x", "value":"nnn"}]` is a bit of an overkill. 

`"history":[{"path":"y.z.x", "values":{"timestamp":"123456", "value":"nnn"}}]}`
is a not really aesthetic due to the double "value(s)" - a first indicator that something might be wrong with it.

`[{"path":"y.z.x", "history":{"timestamp":"123456", "value":"nnn"}}]}`

I am in disagreement about "$source". I don't keep track of the source. It adds overhead and I don't see the use case in the front end either. Sounds like something for hunters & collectors :)

tkurki:
See the stuff I added to #89. Arrays are totally ok here I think, where we present an ordered set of values.

PR #214, opened 2016-04-22
Temperatures were still not consistent

sumps:
The use of a sub-group Air and having some Temperatures not as Objects made the whole temperature sub-group inconsistent and messy. These changes aim to tidy everything up and make them consistent in terms of a location and its environment.

### DIFF
--- a/schemas/groups/environment.json
+++ b/schemas/groups/environment.json
@@ -17,100 +17,123 @@
     }
   },
   "properties": {
-    "air": {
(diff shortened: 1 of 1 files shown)


tkurki:
Tried to fix things in https://github.com/SignalK/specification/blob/a403e39be21891cf0d1e90f2edf00b7208b2e6f5/test/data/temperatures.json in #217

PR #216, opened 2016-04-22
Update temperatures.json

sumps:
To match the changes in....
https://github.com/SignalK/specification/pull/214

### DIFF
--- a/test/data/temperatures.json
+++ b/test/data/temperatures.json
@@ -11,66 +11,46 @@
             "pgn": 130312
           }
         },
-        "air": {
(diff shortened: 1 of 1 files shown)


tkurki:
The json has invalid json structure and liveWell and baitWell are inside.liveWell, which isn't quite right.

tkurki:
Tried to fix things in https://github.com/SignalK/specification/blob/a403e39be21891cf0d1e90f2edf00b7208b2e6f5/test/data/temperatures.json in #217

PR #217, opened 2016-04-24
Reorganise temperatures a bit

tkurki:
- remove air hierarchy level
- drop well hierarchy level
- move baitWell and liveWell under water


tkurki:
Supercedes #216 and #214.

PR #221, opened 2016-05-15
Add GC/RL distinction

### LINKED ISSUE #219: Great circle course data
Current course information has only "default" (rhumbline) figures. Great circle versions should be added.

tkurki:
Create two course subtrees, one for Great Circle calculated values and
the other for Rhumbline calculated ones.

Fixes #219


tkurki:
Merged.

PR #225, opened 2016-05-18
Add defaults overlay

timmathews:
Do we still need this branch to support defaults overlay? If so, can we merge it?


tkurki:
I added this to support https://github.com/SignalK/instrumentpanel/pull/21 : to support a way to populate the full tree's meta elements. Zones is still work in progress, but I suppose this could be merged.

rob42:
@timmathews - do we still need this? If so can you fix the conflicts and I will merge.

Issue #227, opened 2016-06-04
keyswithmetadata.json with not only numeric values


sailoog:
Hi,
I need to create menus in OpenPlotter with all the signal k data items to be selected and bounded with sensors and triggers in the monitoring system. keyswithmetadata.json do the work for sensors but for triggers I would need that on keyswithmetadata.json are all the possible signal k data items, not only numeric values and units or timestamp but also string values and strings values with enum data.
I have been trying to edit extractKeysAndMeta.js but unsuccessfully :( 
any help? this could be useful for other apps too.

PR #228, opened 2016-06-04
Include enums in keyswithmetadata.json

tkurki:
@sailoog is this what you're after in #227 ?


sailoog:
Yessssssss
Thanks a lot @tkurki 
The new keys don't have description, could be possible to add it? This file will be really useful for apps to build menus.
Thanks again.

PR #230, opened 2016-06-10
Added examples, and tidied descriptions

rob42:

PR #232, opened 2016-06-28
Added missing message types

rob42:
Added subscribe, unsubscribe, get, put, list message types

### DIFF
--- a/schemas/definitions.json
+++ b/schemas/definitions.json
@@ -13,41 +13,50 @@
     },
     "source": {
       "type": "object",
-      "description": "Source of data. An object containing at least the properties defined in 'properties', but can contain anything beyond that.",
+      "description": "Source of data, a record of where the data was received from. An object containing at least the properties defined in 'properties', but can contain anything beyond that.",
       "required": [
         "label",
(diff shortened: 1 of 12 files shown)


rob42:
Aargh..this pull has inherited #230 which has not yet been merged...

rob42:
Merging as there are no comments or objections

rob42:
Re subscriptions - oops - should be a $ref. I will fix the duplication.

The need for the `get` and `put` over websockets came from my recent work on freeboard-sk and xmpp. When using an async message transfer (like xmpp or any pub/sub protocol) a REST call doesnt work anymore. It becomes necessary to store and forward the message, but it was not valid (or defined anywhere else) in the current delta.json. 

The `get` has no equivalent except to `subscribe` and get value, and `unsubscribe` for devices without http  capabilities. The `put` could be done with an `update` but the use case is different, so separating them allows for finer-grained control. `list` is not really as important, I included it for completeness.

I dont want to remove `get` or `put` as I use these now.
I will add to the docs for these.

rob42:
See  #239

PR #233, opened 2016-07-18
Added chart scale, and some new charts types

rob42:
Ive been working on my web chartplotter and I can store and retrieve charts nicely, but I cant easily derive the scale of the chart. This means I cant arrange the layer stack on screen and hi-res charts hide under low-res charts, etc

Adding scale fixes this.


rob42:
Merging as there are no comments or objections

PR #238, opened 2016-08-03
Source handling for NMEA and non-NMEA sources

tkurki:
This is issue is for discussing source handling for all kinds of sources, NMEA and non-NMEA alike.

### DIFF
--- a/schemas/delta.json
+++ b/schemas/delta.json
@@ -17,18 +17,15 @@
       "items": {
         "type": "object",
         "required": ["values"],
+        "dependencies": {
+					"not":{"$source":"source"},
+					"not":{"$source":"source"}
+					},
         "properties": {
-          "source": {
-            "allOf": [{
-              "$ref": "./definitions.json#/definitions/source"
-            }, {
-              "properties": {
-                "instance": {
-                  "type": "string"
-                }
-              }
-            }]
-          },
+        	
+						"$source": {"$ref": "./definitions.json#/definitions/sourceRef"},
+						"source": {"$ref": "./definitions.json#/definitions/source"}
+					,
           "timestamp": {
             "$ref": "./definitions.json#/definitions/timestamp"
           },
--- a/schemas/delta.json
+++ b/schemas/delta.json
@@ -16,16 +16,17 @@
       "description": "The updates",
       "items": {
         "type": "object",
-        "required": ["values"],
-        "dependencies": {
-					"not":{"$source":"source"},
-					"not":{"$source":"source"}
-					},
+       	"required": ["values"],
+				
+       	"not":{"allOf": [ 
+					{"required": ["source"]},
+					{"required": ["$source"]}
+				]}
+				,
+            
         "properties": {
-        	
-						"$source": {"$ref": "./definitions.json#/definitions/sourceRef"},
-						"source": {"$ref": "./definitions.json#/definitions/source"}
-					,
+        	"$source": {"$ref": "./definitions.json#/definitions/sourceRef"},
+        	"source": {"$ref": "./definitions.json#/definitions/source"},
           "timestamp": {
             "$ref": "./definitions.json#/definitions/timestamp"
           },
--- a/test/sources.js
+++ b/test/sources.js
@@ -54,7 +54,7 @@ var deltaWithMiscSources = {
       "path": "electrical.batteries.house.current",
       "value": -0.837
     }]
-  },{
+  }, {
     "timestamp": "2016-08-03T07:55:57.000Z",
     "values": [{
       "path": "navigation.headingTrue",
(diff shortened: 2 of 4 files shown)


tkurki:
See 417b185 for a comprehensive example of proposed source handling in delta messages for different kinds of current and future sources.

The basic idea is that there can be well known sources with predefined attributes, like `talker` and ? `sentence` or `src` and `pgn`, or hierarhical sources with grouping/hierarchy encoded in a single `$source` string, period-separated. These result in hierarchy under `/sources` in the full data model.

tkurki:
@sailoog how does this look to you?

sailoog:
Perfect!

rob42:
I favour this scheme. A general rule of `sources.protocol.bus.device.data` is suitable for most cases we see at present, but there is no reason not to use a complete freeform path if required. The path is referenced in the `$source` entry anyway so it can always be found back. 

That said we should use standard patterns for common protocols like n2k and 0183 as these are very common and will aid apps that want to inventory all n2k devices etc.

So that suggests _well-known_ paths like `sources.0183.ttyUSB0.GP.RMC` and allows for `sources.i2c.0.0x48.volts`

rob42:
```
{
	"sources": {
		"0183": {
			"ttyUSB0": {
			  "GP":{
					"RMC": {
						"value": "$ECRMC,194310,A,3748.788,N,12226.836,W,8.3,316.0,140708,015.0,E,A*10",
						"timestamp": "2016-08-05T01:59:33.890Z"
					},
					"GSV": {
						"value": "$GPGSV,3,2,12,08,08,243,24,04,13,287,23,06,08,102,29,16,38,058,44*75",
						"timestamp": "2016-08-05T01:59:39.890Z"
					}
				}
			}
		},
		"ais": {
			"366988490": {
				"StaticCommon": {
					"value": "!AIVDM,2,1,9,B,55Mw<jP09sNIL@?;7?H@Di@60<51@4Tp0000000l0`E2540000000000,0*2F\r\n!AIVDM,2,2,9,B,000000000000000,2*2E",
					"timestamp": "1969-12-31T23:59:59.999Z"
				},
				"Position": {
					"value": "!AIVDM,1,1,,B,15Mw<jPP01G@9KhEWGA8FgvP089:,0*71",
					"timestamp": "1969-12-31T23:59:59.999Z"
				}
			}
		}
	}
}

And references in the vessels.* tree:
"$source":"0183.ttyUSB0.GP.RMC"
"$source":"0183.ttyUSB0.GP.GSV"
"$source":"ais.366988490.StaticCommon"
"$source":"ais.366988490.Position"
```

rob42:
We also need to make $source valid in delta messages, as it reduces clutter hugely, and the source is available when required anyway.

Maybe a subscription option `includeSource=true` or similar

tkurki:
I think our views on sources differ still a little. To me ais is not a source: consider a n2k connected ais receiver: it would just show up as a source within the n2k bus, producing ais pgns. Likewise for say a multiplexed 0183 input with data from ais, wind and gps.

rob42:
Oh yes, you are right. I have been getting my ais from a demo file, and using source=ais. In fact it should be under either `sources.n2k...` if it comes from an n2k decoder. If its serial over NMEA0183 then `sources.0183.1.AI.VDM` etc. But it does show how the scheme incorporates arbitrary paths quite nicely :-)

tkurki:
The test in the first commit is about $source being valid in delta. Next step is to alter the schema to make the test pass.

Then add a way to assert that $source s and the /source hierarchy match in a full model. That can be used as part of the testclient.

rob42:
Ive altered the schemas to allow either `source` or `$source` but not both in delta messages. Also removed the mandatory `type` so only `label` is required.

test/sources.js
```diff
@@ -7,3 +7,82 @@ describe('Sources in the full tree', function() {
...
+      "value": 0.2231
+    }]
+  },]
+}
+
+
+describe('Bad sources in delta', function() {
+  it("are not valid", function() {
+    !deltaWithBadSources.should.be.validSignalKDelta;
+  });
```
tkurki: I don't think this works, the assert triggers before the negation. 

Try should.not.be.

schemas/delta.json
```diff
@@ -17,18 +17,15 @@
       "items": {
         "type": "object",
         "required": ["values"],
+        "dependencies": {
+					"not":{"$source":"source"},
+					"not":{"$source":"source"}
```
tkurki: Duplicate line?

tkurki:
Add a test for missing delta?

tkurki:
I would be ready to merge this. The changes are 100% backwards compatible afaik.

Next step would be to implement this in the servers, at least in node server for Open Plotter usage.

tkurki:
On second thought: how about a regex for `$source` and documentation  that `.` is a delimiter in `$source` that refers to the hierarchy under `/sources`?

rob42:
There is documentation on $source in definitions.

rob42:
The java server already behaves roughly this way, but Ive added some mods to better follow the `sources.protocol.bus.device.data` pattern

sailoog:
@tkurki @rob42 Thanks a lot!

rob42:
This is ready to merge - any dissenters? Say now or....
I will merge tomorrow.

PR #239, opened 2016-08-15
Move delta.json to /messages and split into updates.json and getPutLi…

rob42:
…st.json


rob42:
While looking into comments on the #232 it seemed that the previous  delta.json was actually related to `updates` only. We already had a `subscribe` and `unsubscribe` message in`schemas /messages` and the formats didnt allow `$refs` etc to reuse them. 

So I moved the old delta.json to messages/updates.json, and added a new schema for get, put, and list.

Not ready yet as there are no tests on `getPutList`

tkurki:
We have used the word "delta" for quite a long time to refer to the message with `updates`, it is in mailing list archives, Github issues and wiki.

I think we should continue to do so and renaming the schema file for that message  to updates.json is just confusing.

+1 on having that schema file contain just the `updates`.

tkurki:
Renamed updates back to json. 

@rob42 isn't the `"reply-to"` misplaced in the delta.json, it is not relevant in delta?

rob42:
No its correct. In my trials with MQTT/STOMP/XMPP ist not possible to reply to a specific connection as they all come from the same message broker, hence session/IP. The `reply-to` key allows a message or correlation id to be attached to the update, so that the broker can then route the message back to the proper target. 
eg MQTT device `1234567` subscribes for `navigation.*`, by sending a message to the MQTT broker, and listens for replies on a temporary queue. Periodic updates are sent to the MQTT broker, and the broker then knows they are  to be routed to the temporary queue in `reply-to`.

rob42:
Old and defunct now I think.

Issue #46, opened 2014-12-12
Human readable specification document usable in offline format


tkurki:
To better promote Signal K we should create a nicely formatted, printable and downloadable version of the specification.

woobagooba:
+1 to that.  Minimize the effort required to read/eval the spec.  Expectation when clicking through your site to Specification is ... well ... the specification is readily available.  It is not.  Cheers.

fabdrol:
@woobagooba would you prefer a paper document over (better) online documentation?

tkurki:
offline != paper. 

Current, clickable experience is not usable without a net connection, not random browsable and not searchable.

woobagooba:
What Teppo said

PDF works

Thanks for checking

fabdrol:
In that case, lets release a pdf upon every stable release - but not in between. Saves a lot of work and ensures that people with old PDFs develop against a certain stable version others are using too

tkurki:
Fixed by #255, see http://signalk.org/specification/master/

PR #240, opened 2016-08-19
Standard js api to look up metadata for a key

tkurki:
@sumps brought up the need for a standard (js) api to look up metadata, such as units and description and example, about a key.

This PR adds one with JavaScript implementation. I think this should be exported in https://github.com/SignalK/signalk-js-client/ but the implementation is really schema dependent so should live close to it.

TODO
- [x] support for foo.*.bar keys
- [ ] ~~proper implementation based on schema files, not on the generated keyswithmetadata.json~~
- [ ] documentation (in signalk-js-client ?)


tkurki:
I realised that if we produce a keyswithmetadata.json for people to use as such and as the fodder for `getMetadata` implementation the same data would work pretty well as schema keys documentation, including descriptions, units and examples.

I am not sure if it is a great idea to use hand crafted logic like in this PR to pull the information out of the schema files, but I don't see a viable alternative. Both @rob42's keys listing in java and @fabdrol's efforts do essentially the same, figure out a way to extract a selection of stuff from the schema files.

Anyway, take a look at doc.md in this PR or run `npm run build` to generate it for yourself.

The file does not have the intermediate levels but nor does it [have too much information going through your head](https://open.spotify.com/track/34duNcSyV9qP1Q0kEYn2C9).

The .md idea I credit to @rob42.

So essentially this is a step towards #46.

sailoog:
Really useful. I miss defined values in doc.md. 
ex. 
design.keel.type: "enum": [
      "long",
      "fin",
      "flare",
      "bulb",
      "wing",
      "centerboard",
      "kanting",
      "lifting",
      "daggerboard"
]

tkurki:
- [ ] enums to doc

rob42:
You missed the best of all `foiling` :-)

tkurki:
Depends on #290.

tkurki:
@sailoog enums have been in the doc for some while now: http://signalk.org/specification/master/keys/#vesselsregexpdesignkeeltype

tkurki:
Superceded by #332

PR #243, opened 2016-08-27
Fix notifications to require method and state

rob42:
Fix schema to reqiuire `state` and `method` and update samples to work

### DIFF
--- a/schemas/groups/notifications.json
+++ b/schemas/groups/notifications.json
@@ -11,9 +11,12 @@
     "(^method$)": {
       "description": "Method to use to raise notifications",
       "type": "array",
-      "items": {
-        "$ref": "../definitions.json#/definitions/alarmMethodEnum"
-      }
+			"title": "Alarm Method",
+			"description": "The method to use to raise the alarm. An alarm requires immediate attention, eg no oil pressure",
+			"default": ["visual", "sound"],
+			"items": {
+				"$ref": "../definitions.json#/definitions/alarmMethodEnum"
+			}
     },
 
     "(^state$)": {
--- a/schemas/groups/notifications.json
+++ b/schemas/groups/notifications.json
@@ -5,7 +5,7 @@
   "title": "notifications",
   "description": "Notifications, their state, and actions. The notification limits are set in any Signal K key.meta.zones array.",
   "dependencies": {
-  	"message": ["method","state"]
+  	"message": ["state"]
   	},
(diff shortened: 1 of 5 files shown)


tkurki:
I am good on mandatory `state`, but I don't think `method` should be required.

The sender of a notification can have an opinion on the severity/state of the notification, but how can a sender know what capabilities the receiver has and how the receiver should or could take action?

Example:  I have some fancy gadget around my wrist that first vibrates, then adds a flashing strobe and then upgrades to electric shocks until the notification is acknowledged - and my whole setup doesn't have audible alarms, because I am deaf.

rob42:
The `method` is taken from an enum `[ "visual", "sound" ]. We can add more as required, but they need to be well-defined as they become constants. `method` also needs a default behaviour especially if its not mandatory. 
I didnt know you were deaf :-) but custom behaviour on notification belongs to the app, not the protocol. In your case your wristwatch needs to compensate with flashing>shocks for `visual`, and immediately go to shocks for `sound`.

Maybe the `method` options can have a default with `state`, eg `alarm` or above has `sound`, below has `visual`?

tkurki:
As you _almost_ said yourself: behaviour on notification belongs to the app, not the protocol. 

Having a field for specifying the action to take is good, making it mandatory too strict and will result in either non-compliant software or just stuffing some value in there, making the field meaningless.

schemas/groups/notifications.json
```diff
@@ -4,14 +4,19 @@
   "id": "https://signalk.github.io/specification/schemas/groups/notifications.json#",
   "title": "notifications",
   "description": "Notifications, their state, and actions. The notification limits are set in any Signal K key.meta.zones array.",
-
+  "dependencies": {
+  	"message": ["state"]
```
tkurki: Could you elaborate on this - none of the properties are now required, but if there is `message` then `state` must also be there.

rob42: Because the depth of the hierarchy is arbitrary I couldnt use `required` as it then applies to all levels of the path. By using `dependencies` we force the use of `state` whereever `message` is used, which is the best compromise I could find.

tkurki:
I would like to close this:
- Having a field for specifying the action to take is good, making it mandatory too strict and will result in either non-compliant software or just stuffing some value in there, making the field meaningless
- Introduces a complicated schema structure with very little benefit

PR #245, opened 2016-08-31
Make tank capacity numberValue

tkurki:
Tank capacity should be a Signal K numberValue (with value/values/timestamp/source info) instead of a regular number, because it can be updated from for example PGN 127505. 

See also https://github.com/SignalK/n2k-signalk/pull/32


tkurki:
- [ ] convert to m3

PR #249, opened 2016-09-13
Gitbook documentation

tkurki:
This branch adds a way to create Gitbook documentation from Markdown source files and publish them at http://signalk.org/specification/master/ eg. version specific subdirectories.

The Gitbook toolchain can produce also pdf, epub and mobi versions of the document, fixing #46.

It also includes a set of documentation files, including a data model keys reference document.


tkurki:
I think this is ready for merging, we can continue incremental improvement off master.

The one TODO item is per key detailed doc page, but I'm happy with @fabdrol 's solution (not too hacky) once we exclude the bullshit keys (pgn, timestamp etc under all the real keys).

tkurki:
Replaced by #255, attack keys.json  and per key doc pages separately off master.

PR #250, opened 2016-09-15
rporated Gitbook docgen

fabdrol:
Incorporated fabdrol/signalk-documentation-generator in the gitbook branch. There are still some issues with detail pages for individual keys due to the way gitbook works (or rather, doesn't work): 

- gitbook won't parse documents that are not referenced in `SUMMARY.md`. Adding a reference to every single key detail page in the summary doesn't work (I tried, believe me), as the gitbook CLI hangs
- gitbook has no support for nested SUMMARY.md files. This issue was mentioned on the gitbook repo in april 2016 (and referenced a couple more times afterwards) but nobody has replied to it (in fact, gitbook doesn't seem very active at fixing issues with >500 open issues on github!): https://github.com/GitbookIO/gitbook/issues/1251
- there's no way with gitbook's markdown parser to force a link to open in a new window. As such, the only way to make sure the links open is to use absolute URLs. 

The solution to the problem of detail pages for the keys was very, very hacky: fabdrol/signalk-documentation-generator parses HTML versions of every detail page. In the index, links to the Signal K documentation side are generated (essentially hardcoded). This will provide problems once we try things like a version of the docs for each version of the spec. 

In addition to the problems mentioned above, are the generated detail pages not as nice as the rest of the docs. A solution should be found, but for now this will do the trick.


tkurki:
Integrated into #249

Issue #251, opened 2016-09-16
Steering group reorganizing


joabakk:
While looking for autopilot control from Signal K data, I discovered some discrepancies between the way the specification groups values and the way most autopilots use these values. 
If I use SignalK to control my autopilot, and select wind as source, then the description of steering.autopilot.targetHeadingNorth  or steering.autopilot.targetHeadingMagnetic is misleading, since I would input an angle relative to wind, where + is port tack and - is starboard. The description currently relates to north, which is true if you steer from GPS or compass.
From @tkurki Would it make more sense to set/describe the autopilot steering with a combination of  angle and reference, reference being an enum of wind | northMagnetic | northTrue?

From a brief glance at autopilot manuals (Raymarine, Simrad and B&G), it appears that the most common controls are:

1. Magnetic heading
2. GPS heading with additional nodrift mode
3. Wind relative angle, with additional tack/gybe commands
4. Standby
5. Depth contour following mode
6. Nav mode (follow course changes from route)
7. Direct rudder control

In addition to these, there are settings parameters that may be required to fully control the autopilot. 

As these are control parameters, an acknowledgment is probably used back from the autopilot. This may have to be implemented as well, and separate patches to communicate with each system.

I will start on a PR for this when time permits, and appreciate any input and help. I would suggest making a separate branch to work this out.

joabakk:
The spec updates are now implemented, and I will close this now.

PR #252, opened 2016-09-17

joabakk:
As discussed in #251, here is a proposal for reorganizing the steering group to align with autopilot logics.


joabakk:
I will get my code straight and try again...

PR #253, opened 2016-09-17

joabakk:
In line with #251 

Propose a separate branch to work this out. Any input appreciated.


joabakk:
Created from separate private branch

PR #254, opened 2016-09-17
gitbook - with nested key details 

fabdrol:
In this PR the keys docgen generates a detail page for each key. Since gitbook doesn't parse files unless they are in a `SUMMARY.md` file, this version takes the `_SUMMARY.md` file in the gitbook-docs folder, appends the generated details files and writes it to disk as `SUMMARY.md`. 

The downside (and reason this is in a separate branch) to this approach is that the gitbook command generating the HTML, PDF, EPUB and MOBI files now takes over 20 minutes on a fast Macbook Pro (2.2ghz i7, 16gb RAM).


tkurki:
I don't think this approach is feasible - all the keys included in SUMMARY.md blows the left side index of the html version up.

Another stab at this would be to add Appendix B: Schema as snippets to the Summary and generate a file like keys/index.md but with the schema snippets included inline.

Still another would be to include the snippets in keys/index.md, but create a Gitbook plugin that collapses them by default. This way the epub versions would have a single Appendix A with the snippets inline.

PR #257, opened 2016-09-19

joabakk:
Created a separate branch on my side, ref#251

 Changes to be committed:
	modified:   i18n/en.json
	modified:   schemas/groups/steering.json
	new file:   test/data/steering.json
	new file:   test/steering.js

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -802,11 +802,11 @@
     "shortName": "Mode",
     "longName": "Operational Mode of Autopilot"
   },
-  "steering.autopilot.targetAngle": {
+  "steering.autopilot.target.angle": {
     "shortName": "Target (relative) heading",
     "longName": "Target heading for autopilot, relative to magnetic North or Apparent wind +port -starboard"
   },
-  "steering.autopilot.headingReference": {
+  "steering.autopilot.target.reference": {
     "shortName": "Heading Reference",
     "longName": "Current Reference of Heading"
   },
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -804,7 +804,7 @@
   },
   "steering.autopilot.targetAngle": {
     "shortName": "Target (relative) heading",
-    "longName": "TTarget heading for autopilot, relative to magnetic North or Apparent wind +port -starboard"
+    "longName": "Target heading for autopilot, relative to magnetic North or Apparent wind +port -starboard"
   },
   "steering.autopilot.headingReference": {
     "shortName": "Heading Reference",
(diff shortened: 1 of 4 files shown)


i18n/en.json
```diff
@@ -802,21 +802,13 @@
     "shortName": "Mode",
     "longName": "Operational Mode of Autopilot"
   },
-  "steering.autopilot.targetHeadingNorth": {
-    "shortName": "Target Heading North",
-    "longName": "Target Heading North in Degrees True"
+  "steering.autopilot.targetAngle": {
+    "shortName": "Target (relative) heading",
+    "longName": "TTarget heading for autopilot, relative to magnetic North or Apparent wind +port -starboard"
```
timmathews: Ideally, `shortName` should be really short. Like "HTS" for Heading to Steer. `longName` could then be "Target Heading (relative)". What's in `longName` currently better fits in the `description` field below, IMO.

But honestly this is probably a bigger discussion than this PR. The i18n stuff still needs work. Perhaps in addition to `shortName` and `longName` we need `abbreviatedName` as well?

If nothing else, there's an extra 'T' in `longName` that should be removed :smiley:

schemas/groups/steering.json
```diff
@@ -6,12 +6,14 @@
   "title": "steering",
   "properties": {
     "rudderAngle": {
+      "type": "number",
```
timmathews: These aren't necessary, the `$ref` handles this.

schemas/groups/steering.json
```diff
@@ -97,12 +90,6 @@
                 "gps"
               ]
             },
-
-            "source": {
-              "description": "Source of this data",
-              "$ref": "../definitions.json#/definitions/source"
-            },
-
```
timmathews: Curious why you're removing `source` for the enums?

i18n/en.json
joabakk: Removed the extra T in longname. Since as you say this is not specific to this PR, perhaps we can agree and fix all shortnames in a separate PR?

schemas/groups/steering.json
joabakk: Removed now, thanks

schemas/groups/steering.json
joabakk: Oh that was a quick fix/destruct to get the test to pass. Not sure why it doesn't pass with these in place... Help needed

joabakk:
You are right @tkurki . I can commit a change when time allows

joabakk:
Hi and sorry for all the commits. I would squash them if I were on my own computer

PR #261, opened 2016-09-24
Clean ups

thomasonw:
o Generalized dcQuantities away from 'battery' to more generic usage.
o Removed in corrected Celsius reference
o Used common wordage for device limits resulting in disable/disconnect.


timmathews:
Looks good to me, I think this wording makes more sense.

PR #269, opened 2016-10-11
RFC0004 :Replace JsonPath with wildcard in subscription paths

tkurki:
# Summary
[summary]: #summary

This RFC proposes replacing the [JsonPath](http://goessner.net/articles/JsonPath/) expression specified in the subscription protocol with simple wildcards.

# Motivation
[motivation]: #motivation

Current subscription protocol specifies that the path expression can be JsonPath expressions.

I believe that JsonPath is needlessly complex here.

Subscriptions can be expressed sufficiently with the use of a simple wildcard, making server implementation considerably easier.

# Detailed design
[design]: #detailed-design

See diff in this pull request. 

The specification change is mainly in the documentation. Subscription message schema could be enhanced with a regexp.

# Drawbacks

We lose the expressiveness of JsonPath.

# Alternatives
[alternatives]: #alternatives

We could specify the JsonPath support as optional and wildcard support as part of minimal subscription support, but then we would have to deal with complexity of informing the client about which parts the server supports.

# Unresolved questions
[unresolved]: #unresolved-questions

### DIFF
--- a/gitbook-docs/subscription_protocol.md
+++ b/gitbook-docs/subscription_protocol.md
@@ -69,8 +69,8 @@ To subscribe to the required criteria send a suitable subscribe message:
 }
 ```
 
-* `path=[path.to.key]` is appended to the context to specify subsets of the context. The path value can use jsonPath
-  syntax.
+* `path=[path.to.key]` is appended to the context to specify subsets of the context.
+The path value can use the wildcard `*`. A wildcard in the middle of a path (`propulsion/*/oilTemperature`) allows any value for that part and a wildcard at the end (`propulsion/port/*`) matches all paths beginning with the specified prefix.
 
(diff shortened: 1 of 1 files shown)


rob42:
JsonPath provides a standard and powerful way to query the signalk tree, but does involve complexity in implementation. 
We should leave JsonPath as the intended query method so that implementations that provide complex queries use a std protocol rather than a custom method, but we only require simple wildcard support in the spec.
 eg MAY provide query by JsonPath, MUST (SHOULD?) provide query by wildcard

tkurki:
Does java server support JsonPath? If there is no existing implementation I would rather wait for the use case & need for implementation rather than specify stuff before the actual need & use.

rob42:
No. I was using it when I had a json based model, but now use a regex implementation against the hashmap  keys

PR #272, opened 2016-10-15
Update multiple values documentation

tkurki:
The doc is outdated and doesn't match the schema nor the specified logic related to having multiple values in the full model, linked to sources.

TODO:
- [x]  replace temperetures with device oriented examples
- [ ] rewrite with `values` structure
- [ ] add `sources` handling

### DIFF
--- a/gitbook-docs/data_model_multiple_values.md
+++ b/gitbook-docs/data_model_multiple_values.md
@@ -1,95 +1,24 @@
 # Multiple Values for a Key
 
 There are two use cases for multiple data:
-* Multiple versions of a common device - eg two engines
-* Multiple devices providing duplicate data - multiple values for the same signalk key from different sensors, eg COG from both compass and gps
+* Multiple instances of a common device - eg two engines or multiple batteries
+* Multiple devices providing duplicate data - multiple values for the same Signal K key from different sensors, eg COG from both compass and gps or multiple depth sounders
 
-###Multiple versions of a common device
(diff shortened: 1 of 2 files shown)


tkurki:
Ping @rob42 @thomasonw

thomasonw:
@tkurki - nice addition.  Question:  In the overall vision for Signal-K is there an intention at some time to do value-add processing?   Examples might include:

1. Qualifying of multiple position (lat/long) data values, selecting the 'preferred' one to commonly pass on to all apps requesting 'position'.  (Still allowing for a way to access data from all sources)
2. Monitoring of data values vs. limit values, proactively invoking an 'alarm' process of some type?

Just wondering.  Many value-add processes I think can live well at the app level (ala, example 2 above), but some seems like being the man-on-the-middle would be useful for a few (example 1 above).

tkurki:
See https://github.com/SignalK/signalk-server-node/pull/105 for a plugin mechanism for server side value-add processing.

thomasonw:
@tkurki Very nice!

rob42:
@thomasonw - Every key can have a `meta` tag which includes ranges and alarms that apply to them. The server raises alarms by monitoring the value against the range and placing an entry in the `notifications` structure. See http://signalk.org/specification/master/data_model_metadata.html

In the java server background processes are used to derive true/apparent wind, periodically calculate magnetic deviation, anchor watch, etc.

thomasonw:
You know, I will apologize for my lack of knowledge of what is already in SignalK - do appreciate the patience to 'educate me'!

tkurki:
Merging, as this is better than the current version.

PR #274, opened 2016-10-18
Resolve relative references in the same file

### LINKED ISSUE #259: processSchemaFiles does not handle file-internal references
References in json schema files can reference stuff from the same schema file's `definitions` 

```
"$ref": "#/definitions/dcQuantities"
```
or from another file/schema with a relative ref:
```
"$ref": "../definitions.json#/definitions/timestamp",
```

but `processSchemaFiles` only handles the latter:

https://github.com/SignalK/specification/blob/master/scripts/processSchemaFiles.js#L472-L483

joux3:
Falls back to `definitions.json` if not found in the same file.

Fixes #259 
The diff in keys.md after fix: https://gist.github.com/joux3/14ad4d2654e00f556f3dd1563c9f9cf8

### DIFF
--- a/scripts/processSchemaFiles.js
+++ b/scripts/processSchemaFiles.js
@@ -413,7 +413,8 @@ class Parser {
         }
 
         if (typeof this.tree[`${prefix}/${key}`] !== 'undefined' && typeof this.tree[`${prefix}/${key}`].allOf !== 'undefined') {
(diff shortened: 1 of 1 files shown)


joux3:
This could probably use some work before merging. Right now references properties overwrite keys that are already defined. For example the fields `title` and `description` in `batteries` get overwritten by definitions in `dcQuantities`.

joux3:
Latest commit passes base properties to `allOf` reducer and stops `allOf` from overwriting already existing fields.

PR #276, opened 2016-10-19
Parse local schema in processSchemaFiles.js

joux3:
Previously the code fetched the schema available under http://signalk.org/specification/schemas/. This however makes developing locally harder. The schemas online also seem to be out of sync with the repository.

On a related note: should the online schemas be updated and by who and when?

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,6 @@
   "devDependencies": {
     "gitbook-cli": "^2.3.0",
     "infuse.js": "^2.0.2",
-    "json-schema-load-tree": "^0.2.0",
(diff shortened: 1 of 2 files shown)


joux3:
On a more closer inspection I don't believe writing this from scratch makes sense. Switching from `json-schema-load-tree` to for example [json-schema-ref-parser](https://www.npmjs.com/package/json-schema-ref-parser) would allow reading and schema reference resolving in one go. This would reduce the amount of code needed.

Closing this for now. I will open a new pull request if/when I have done more work regarding that.

PR #278, opened 2016-10-23
Validate missing schema references & fix missing references

joux3:
These changes add a test for missing schema references. The already existing test actually didn't try to resolve the reference key paths, it was only concerned about files. The changeset also includes fixes for missing schema references.

The fixes include some trivial changes and one that is largely in progress. The trivial ones are:
1. accidentally referring to `$source` instead of `sourceRef`
2. lacking `../definitions.json`

The more WIP one is the handling of `lineString` and `polygon` references in `groups/resources.json`. Do you have any input @rob42? Right now I just reference the `external/geojson/geometry.json` using relative links. This however leads to certain issues as the ID of the `geometry.json` schema is `http://json-schema.org/geojson/geometry.json` (which BTW is a 404). One way would be to bring the external schemas under Signal K related ID.

### DIFF
--- a/index.js
+++ b/index.js
@@ -46,6 +46,12 @@ function getTv4() {
   for (var schema in subSchemas) {
     tv4.addSchema('https://signalk.github.io/specification/schemas/groups/' + schema + '.json', subSchemas[schema]);
   }
+
+  // HACK! two different IDs should not point to the same schema
+  var externalGeometry = require('./schemas/external/geojson/geometry.json');
+  tv4.addSchema('https://signalk.github.io/specification/schemas/external/geojson/geometry.json', externalGeometry);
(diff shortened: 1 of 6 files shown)


tkurki:
Ping @rob42

rob42:
Geojson allows us to describe complex geometric shapes so I imagine we will eventually use much more than the few elements we do now. Seems like a good re-use of existing functionality
I originally included the geojson schemas in the external dir as I needed them to do the docs (old docson all in one page) if I remember correctly. Ideally we should lookup the geojson from the geojson.org site but that didnt seem to work. 
I dont want to bring them under the signalk namespace as I am trying to reuse geojson, rather than copy it into signalk and this solution worked at the time.  Jsonschema  seemed to be less developed in resolving  references to remote schemas than xml in this respect.

Better ideas welcome.

PR #280, opened 2016-10-27
Added a separate field for pilot wind angle

sbender9:


tkurki:
As I understand the schema here there should be no need for a separate field, instead autopilot steering to the wind should have the target angle in `angle` and `reference` set to `wind`.

@joabakk you expanded the autopilot schema and @timmathews merged it - comments?

joabakk:
I'm open to any logic that captures the combinations without redundancy in info. I understand that Scott's autopilot spits out both wind angle and heading angle when in wind vane mode. It would be a shame to forego this, but I'm not sure this PR does the trick. I'll have a closer look at the original logic.

sbender9:
Yeah, the problem is that Raymarine sends out heading and wind angle. Since there's no state in the node server, we don't know which one to use.

tkurki:
Let's fix this at n2k-signalk.

PR #283, opened 2016-11-01
Add Transverse Water Speed and Leeway Angle

sumps:
A new generation of electro magnetic transducers that can measure the boat's speed through the water both longitudinal and transverse will provide a view of the boat's "Leeway" that was previously difficult to accurately calculate. Signal K needs to support these new transducers which will output the VBW sentence or PGN 130578.

### DIFF
--- a/schemas/groups/navigation.json
+++ b/schemas/groups/navigation.json
@@ -432,17 +432,17 @@
     },
     "speedThroughWater": {
       "$ref": "../definitions.json#/definitions/numberValue",
-      "description": "Vessel speed through the water",
+      "description": "Vessel speed through the water (longitudinal)",
       "units": "m/s"
     },
-    "speedTransverseWater": {
+    "speedThroughWaterTransverse": {
       "$ref": "../definitions.json#/definitions/numberValue",
       "description": "Transverse speed through the water (Leeway)",
       "units": "m/s"
     },
     "leewayAngle": {
       "$ref": "../definitions.json#/definitions/numberValue",
-      "description": "Leeway Angle derived from transverse speed through the water",
+      "description": "Leeway Angle derived from the longitudinal and transverse speeds through the water",
       "units": "rad"
     },
     "log": {
--- a/schemas/groups/navigation.json
+++ b/schemas/groups/navigation.json
@@ -435,6 +435,16 @@
       "description": "Vessel speed through the water",
       "units": "m/s"
     },
+    "speedTransverseWater": {
+      "$ref": "../definitions.json#/definitions/numberValue",
+      "description": "Transverse speed through the water (Leeway)",
+      "units": "m/s"
+    },
+    "leewayAngle": {
+      "$ref": "../definitions.json#/definitions/numberValue",
+      "description": "Leeway Angle derived from transverse speed through the water",
+      "units": "rad"
+    },
     "log": {
       "$ref": "../definitions.json#/definitions/numberValue",
       "description": "Log value",
--- a/schemas/groups/navigation.json
+++ b/schemas/groups/navigation.json
@@ -432,12 +432,17 @@
     },
     "speedThroughWater": {
       "$ref": "../definitions.json#/definitions/numberValue",
-      "description": "Vessel speed through the water (longitudinal)",
+      "description": "Vessel speed through the water",
       "units": "m/s"
     },
     "speedThroughWaterTransverse": {
       "$ref": "../definitions.json#/definitions/numberValue",
       "description": "Transverse speed through the water (Leeway)",
+      "units": "m/s"
+    },
+     "speedThroughWaterLongitudinal": {
+      "$ref": "../definitions.json#/definitions/numberValue",
+      "description": "Longitudinal speed through the water (outside boundary layer)",
       "units": "m/s"
     },
     "leewayAngle": {


sumps:
To calculate the Leeway Angle (degrees), you need both the longitudinal water velocity (knots) and the transverse water velocity (knots). The calculation is as follows (assuming that the sensor alignment is correct and the sensor has been field calibrated):

Leeway Angle (degrees) =(180/Pi) (X) atan (Leeway Velocity/Forward velocity)

tkurki:
Should it be speedThroughWaterTransverse to be in line with speedThroughWater?

What is positive/negative direction for speed? angle?

Does the way the angle is calculated need to be specified in the schema, wouldn't it be better just to say what is the meaning of the key?

sumps:
I am happy with speedThroughWaterTransverse.

A negative value of Transverse Water Speed should indicate Port and Positive indicate Starboard.

In fact these transducers can measure negative speedThroughWater, not something that was possible with a paddle wheel transducer, so we should include in the documentation for speedThroughWater that a negative value means the boat is moving astern (reverse).

The calculation was just included for information and I think at some stage the Signal K community would benefit from a White Paper on derived values, with all of the necessary calculations included.

tkurki:
My point was that

>Leeway Angle derived from transverse speed through the water

describes one method for deriving the angle and doesn't really explain what the value means. 

Theoretically an underwater "windvane" would be able to measure leewayAngle and one could derive transverse speed from the angle.

How do we go about updating the addition? If you could please make the changes to https://github.com/sumps/specification/tree/patch-1 this Pull Request would reflect the changes automatically.

pod909:
The sensor in question does not give pure leeway as it doesn't account for cross flow and will still need detailed calibration. It gives "longitudinal speed of water passing the boat" and "transverse speed of water passing the boat", where as a paddle wheel just gives you "probably longitudinal speed of water passing the boat"

As with a paddle wheel these are only converted to boatspeed through the calibration. The direction with which boatspeed is given is entirely dependent on the calibration protocol.

To properly record all this you need to recognize the difference between the following:

1. raw longitudinal water speed (typical paddle wheel derived boat speed)
2. raw cross flow/transverse water speed
3. calibrated boatspeed (longitudinal)
4. calibrated boatspeed (direction of travel through the water, ie. accounting for leeway angle)
5. calibrated boatspeed (transverse)

I need the model to support them all as they are all recorded by the d10.

The apparent wind also needs to consider the direction of travel of the MHU 

Equally for heading there is:
The direction the boat is pointing - magnetic
The direction the boat is pointing - true
Direction of travel through water - true
...
and COG

sumps:
@pod909 I think we should get the raw data in to the model first.

In the majority of systems, you cannot tell if a value is calibrated or not and I think it makes for easier and better system design for the client or display system to receive raw data and do the calibration and calculation of derived values.

pod909:
Would prefer crossFlow for the raw value as you can get it with out the boat actually moving sideways through the water outside the boundary layer.... but if we have speedThroughWaterTransverse can we introduce speedThroughWaterLongitudinal at the same time?

All the systems out there follow the same model. The raw values are sent from sensors, processing is done on the server and (often very badly) calibrated values are sent to the display. Both the raw and calibrated values exist on the server and network at the same time so there's a need to be differentiate them.

For any half decent true wind calculation if you're going to use the boatspeed provide by another system you need to know the frame of reference. Has it been calibrated to give the longitudinal component of boaspeed through the water or boat speed through the water its self. For systems that care about such things Signal K needs to be able to differentiate between them to be useful.

tkurki:
@pod909 could you come up with concrete additions to the SK data model: new paths with descriptions and possibly revisions to the descriptions of the current paths?

Or are you happy with the changes in this pull request, we just need to add a few more?

pod909:
I'm happy with what's there (although speedTransversWater seems a little clumsy)

Since we're talking about a device that measures longitudinal flow if we could add 1 more at the same time -- speedThroughWaterLongitudinal or speedLongitudinalWater -- that would be fine.

That would help greatly with the other problem I discussed in terms of systems being precise about what they provide. It's possible to calibrate a paddle wheel so it gives speed in the direction of leeway for instance.

...

Personally I'd be calling them longitudinalFlow, crossFlow, angleOfAttack and speedThroughWater but there we go.

sumps:
I've changed the key to speedThroughWaterTransverse for better consistancy and added a note to the speedThroughWater description to say that it is longitudinal boat speed but we cannot change the key name to speedThroughWaterLongitudinal without breaking the existing schema.

sumps:
@pod909 just to clarify, you are saying that we should have a speedThroughWaterLongitudinal object for these electro magentic logs because they measure a different boat speed than a conventional paddle wheel. 

I can accept this, as they measure outside the boundary layer and are less affected by the hull shape and location.

pod909:
Yep
+
also you can calibrate a paddle wheel so that it gives you speedThroughWater or speedThroughWaterLongitudinal and the difference is material when you try to calculate a worthy true wind.

e.g.

calibrate based on time taken to sail a distance on GPS upwind and you've actually calibrated the boatspeed to speed in the direction of travel

calibrate by sailing a known distance parallel to the center line of the boat and you've calibrated to speedThroughWaterLongitudinal

tkurki:
So what paths / measurements  do you suggest should be added? Baseline being the current schema.

pod909:
speedThroughWaterTransverse
speedThroughWaterLongitudinal

leave speedThroughWater description as is, no mention of longitudinal

sumps:
OK added Longitudinal value, ready for merging as far as I am concerned.

sumps:
Are we all happy for this to be committed ?

@canboat have you anything to add, after your investigation of these new transducers ?

canboat:
I'm happy except for the description on speedThroughWaterTransverse, it should not mention the boundary layer i.m.o. -- some sensors will measure beyond the boundary layer by design, and others need calibration.

Also we now have leewayAngle (LA), speedThroughWater (STW) and speedThroughWaterLongitudinal + Transverse (STWL + STWT), so is the idea that if you have one of these super sensors that STWL + STWT will give the length of the speed vector components whilst STW will be the length of the full vector with LA being the angle of that vector compared to straight ahead?

Some sensors may actually sense STW and LA, others might sense STWL and STWT directly and compute the other two, I guess.

PR #284, opened 2016-11-02
Steering group in line with autopilot communication

joabakk:
As discussed with @tkurki and @sbender9, this tweak allows the co-existence of wind and heading angles.


tkurki:
The wind angle is not a _heading_, it is the _Target wind angle from the bow_, could you please update the names and the description.

joabakk:
I beleive the latest commit should cover your comment?

Issue #147, opened 2016-01-05
Tcp service - raw json over a tcp connection


rob42:
In #146 a new signalk service for raw jsom messaging over tcp is added `signalk-tcp._tcp`. This is pretty simple and easy to define, we just make a tcp connection, and tx/rx signalk messages as json text. There are a couple of issues that need definition:

1) the json text MUST be UTF-8.
2) session handling. Server should assign a session id per connection, implementation specific.
3) security + authentication - should do https, could do certs, what about user/login - how is that managed
4) message delimiter - do we need a message delimiter or is it implicit in the json format. Ive used a state engine in arduino - it tracks where in a streaming message it currently is, so it knows where start and end is. Quite simple and fast. But for a currupted message or poorly formatted json we may need a RESET message?

pod909:
On 4) to enable SignalK delta to coexist on a stream with NMEA0183 how about using \r\n as a record delimiter? Slightly limiting in terms of JSON but anything else and it would break it completely.

rob42:
Simple \r\n is what we use currently, since signalk json doesnt use \r\n in values (so far anyway) it works quite well. It also assumes that the json is not pretty printed, eg all on one line.  In theory we could also have \r\n in descriptions or some other fields, which would break the current scheme.
Tracking the json format (first { to last }) would fix that, but seems quite complex.
I am more inclined to make the start/end of a message quite distinct, for easy parsing. eg
`!{{....}}` or similar. Also easy to wrap/unwrap, could add a checksum in too !{{....}0aBf}

rob42:
...only for serial BTW...no need for `{{..}}` on packet or message based transports.

tkurki:
>Simple \r\n is what we use currently
Do we have that documented or where is it stated?

Control characters are not allowed in JSON in unescaped form so they are not a problem. Pretty printing is not something to be considered here.

I think this issue should result in an addition to the Signal K documentation: a retitled http://signalk.org/specification/master/streaming_api.html with tcp transport added, as well as additions to http://signalk.org/specification/master/connection.html.

tkurki:
The doc sources are at https://github.com/SignalK/specification/tree/master/gitbook-docs

pod909:
TCP already handles checksum and rebroadcast so no need to add.

... Serial is really the territory of a low bandwidth binary protocol (?).
Tend to feel that the first step there is to not get into the transport
argument as there is no ubiquitous standard emerging for iot yet. A lot can
[...]

PR #286, opened 2016-11-05
Introduction of raw TCP rules

pod909:
Added rules for supporting mixed message TCP sockets.

Each message must begin {
Each message must end \r\n

Added source to hello message
Added description of how to use hello timestamp to increase the accuracy of subsequent delta messages

### DIFF
--- a/gitbook-docs/connection.md
+++ b/gitbook-docs/connection.md
@@ -1,3 +1,4 @@
+
 # Discovery and Connection Establishment
 
 #### Service Discovery
@@ -56,6 +57,7 @@ Using the information above a web client or http capable device can discover and
             "version": "1.1.2",
             "signalk-http": "http://192.168.1.2/signalk/v1/api/",
             "signalk-ws": "ws://192.168.1.2:34567/signalk/v1/stream"
+            "signalk-tcp": "tcp://192.168.1.2:3858"
         }
      }
  }
--- a/gitbook-docs/connection.md
+++ b/gitbook-docs/connection.md
@@ -1,4 +1,3 @@
-
 # Discovery and Connection Establishment
 
 #### Service Discovery
@@ -56,7 +55,7 @@ Using the information above a web client or http capable device can discover and
         "v1": {
             "version": "1.1.2",
             "signalk-http": "http://192.168.1.2/signalk/v1/api/",
-            "signalk-ws": "ws://192.168.1.2:34567/signalk/v1/stream"
+            "signalk-ws": "ws://192.168.1.2:34567/signalk/v1/stream",
             "signalk-tcp": "tcp://192.168.1.2:3858"
         }
      }
--- a/gitbook-docs/connection.md
+++ b/gitbook-docs/connection.md
@@ -6,6 +6,7 @@ A Signal K server SHOULD advertise its services over mDNS/Bonjour. The server MU
 
 * `_signalk-http._tcp` for http API
 * `_signalk-ws._tcp` for WebSocket
+* `_signalk-tcp._tcp` for raw TCP socket
 * `_signalk-https._tcp` for HTTPS API
 * `_signalk-wss._tcp` for secure WebSocket
 
--- a/gitbook-docs/streaming_api.md
+++ b/gitbook-docs/streaming_api.md
@@ -23,3 +23,17 @@ Upon connection a 'hello' message is sent as follows:
   "self": "123456789"
 }
 ```
+
+#### Streaming: raw TCP sockets
+
+A signalk server MAY stream signalk-delta over a raw TCP socket.
+
+When signalk-delta is streamed over a raw TCP socket:
+
+1. The first character of each signalk-delta message MUST be the openning curly bracke (char(123), '{').
+2. Each signalk-delta message MUST be terminated by a carriage return and new line (char(10) + char(13), '\r\n') sent after the final closing curly bracket (char(125), '}').
+3. The server SHOULD send signalk-delta in a compact manner with any unnessesary white space removed.
+4. The server MAY also send sentences for other protocals (e.g. NMEA0183) interleaved with the signalk-delta messages provided that sentences of the other protocal:
+  a) Can not start with a curly bracket first character and 
+  b) Are terminated by a carriage retrun and new line (char(10) + char(13), '\r\n')
+5. The client MUST not assume that only signalk-delta will be sent over the socket.
--- a/gitbook-docs/streaming_api.md
+++ b/gitbook-docs/streaming_api.md
@@ -26,10 +26,12 @@ Upon connection a 'hello' message is sent as follows:
 
 #### Streaming: TCP sockets
 
-A Signal K server MAY stream signalk-delta over a TCP socket.
+A network node MAY stream signalk-delta over a TCP socket.
 
-When signalk-delta is streamed over a TCP socket:
+When a hello or signalk-delta message is sent over a TCP socket:
 
-1. Each signalk-delta message MUST be terminated by a carriage return and new line (char(10) + char(13), '\r\n') sent immediatly after the final closing curly bracket (char(125), '}'). Note: It is common for the APIs for Ethernet and WiFi devices to close a TCP packet when a line feed ( char(13) ) is received.
-2. The first character of each signalk-delta message SHOULD be the openning curly bracket (char(123), '{').
-3. The server SHOULD send signalk-delta in a compact manner with any unnessesary white space removed.
+1. Each message MUST be terminated by a carriage return and new line (char(10) + char(13), '\r\n') sent immediatly after the final closing curly bracket (char(125), '}').
+2. The first character of each message SHOULD be the openning curly bracket (char(123), '{').
+3. The message SHOULD be sent with any unnessesary white space removed.
+
+Note on hardware implementation: The serial APIs for ethernet and wifi modules often close a TCP or UDP packet on receipt of a new line ( char(13) ). Please check the data sheet and ensure the new line character is still sent.
--- a/gitbook-docs/streaming_api.md
+++ b/gitbook-docs/streaming_api.md
@@ -1,4 +1,4 @@
-#### Streaming WebSocket API: /signalk/v1/stream
+ #### Streaming WebSocket API: /signalk/v1/stream
 
 Initiates a WebSocket connection that will start streaming the server's updates as Signal K delta messages. You can specify the contents of the stream by using a specific URL:
 
@@ -30,6 +30,6 @@ A Signal K server MAY stream signalk-delta over a TCP socket.
 
 When signalk-delta is streamed over a TCP socket:
 
-1. Each signalk-delta message MUST be terminated by a carriage return and new line (char(10) + char(13), '\r\n') sent immediatly after the final closing curly bracket (char(125), '}').
+1. Each signalk-delta message MUST be terminated by a carriage return and new line (char(10) + char(13), '\r\n') sent immediatly after the final closing curly bracket (char(125), '}'). Note: It is common for the APIs for Ethernet and WiFi devices to close a TCP packet when a line feed ( char(13) ) is received.
 2. The first character of each signalk-delta message SHOULD be the openning curly bracket (char(123), '{').
 3. The server SHOULD send signalk-delta in a compact manner with any unnessesary white space removed.
--- a/gitbook-docs/streaming_api.md
+++ b/gitbook-docs/streaming_api.md
@@ -1,4 +1,4 @@
- #### Streaming WebSocket API: /signalk/v1/stream
+3 #### Streaming WebSocket API: /signalk/v1/stream
 
 Initiates a WebSocket connection that will start streaming the server's updates as Signal K delta messages. You can specify the contents of the stream by using a specific URL:
 
@@ -30,8 +30,8 @@ A network node MAY stream signalk-delta over a TCP socket.
 
 When a hello or signalk-delta message is sent over a TCP socket:
 
-1. Each message MUST be terminated by a carriage return and new line (char(10) + char(13), '\r\n') sent immediatly after the final closing curly bracket (char(125), '}').
+1. Each message MUST be terminated by a carriage return and new line (char(13) + char(10), '\r\n') sent immediatly after the final closing curly bracket (char(125), '}').
 2. The first character of each message SHOULD be the openning curly bracket (char(123), '{').
 3. The message SHOULD be sent with any unnessesary white space removed.
 
-Note on hardware implementation: The serial APIs for ethernet and wifi modules often close a TCP or UDP packet on receipt of a new line ( char(13) ). Please check the data sheet and ensure the new line character is still sent.
+Note on hardware implementation: The serial APIs for ethernet and wifi modules often close a TCP or UDP packet on receipt of a new line ( char(10) ). Please check the data sheet and ensure the new line character is still sent.
--- a/gitbook-docs/streaming_api.md
+++ b/gitbook-docs/streaming_api.md
@@ -26,14 +26,10 @@ Upon connection a 'hello' message is sent as follows:
 
 #### Streaming: TCP sockets
 
-A signalk server MAY stream signalk-delta over a TCP socket.
+A Signal K server MAY stream signalk-delta over a TCP socket.
 
 When signalk-delta is streamed over a TCP socket:
 
-1. The first character of each signalk-delta message MUST be the openning curly bracket (char(123), '{').
-2. Each signalk-delta message MUST be terminated by a carriage return and new line (char(10) + char(13), '\r\n') sent after the final closing curly bracket (char(125), '}').
+1. Each signalk-delta message MUST be terminated by a carriage return and new line (char(10) + char(13), '\r\n') sent immediatly after the final closing curly bracket (char(125), '}').
+2. The first character of each signalk-delta message SHOULD be the openning curly bracket (char(123), '{').
 3. The server SHOULD send signalk-delta in a compact manner with any unnessesary white space removed.
-4. The server MAY also send sentences for other protocals (e.g. NMEA0183) interleaved with the signalk-delta messages provided that sentences of the other protocal:
-  a) Can not start with a curly bracket first character and 
-  b) Are terminated by a carriage retrun and new line (char(10) + char(13), '\r\n')
-5. The client MUST not assume that only signalk-delta will be sent over the socket.
--- a/gitbook-docs/streaming_api.md
+++ b/gitbook-docs/streaming_api.md
@@ -24,13 +24,13 @@ Upon connection a 'hello' message is sent as follows:
 }
 ```
 
-#### Streaming: raw TCP sockets
+#### Streaming: TCP sockets
 
-A signalk server MAY stream signalk-delta over a raw TCP socket.
+A signalk server MAY stream signalk-delta over a TCP socket.
 
-When signalk-delta is streamed over a raw TCP socket:
+When signalk-delta is streamed over a TCP socket:
 
-1. The first character of each signalk-delta message MUST be the openning curly bracke (char(123), '{').
+1. The first character of each signalk-delta message MUST be the openning curly bracket (char(123), '{').
 2. Each signalk-delta message MUST be terminated by a carriage return and new line (char(10) + char(13), '\r\n') sent after the final closing curly bracket (char(125), '}').
 3. The server SHOULD send signalk-delta in a compact manner with any unnessesary white space removed.
 4. The server MAY also send sentences for other protocals (e.g. NMEA0183) interleaved with the signalk-delta messages provided that sentences of the other protocal:
--- a/gitbook-docs/urls_etc.md
+++ b/gitbook-docs/urls_etc.md
@@ -70,7 +70,7 @@ See [Subscription Protocol](subscription_protocol.html) for more details.
 
 #### Streaming Connection "Hello"
 
-Upon accepting a connection a server MUST send a 'hello' message before any other communicaiton is sent. The format of the 'hello' message is as follows:
+Before a node sends any signalk-delta messages over a streaming connection it MUST send a 'hello' message. The format of the 'hello' message is as follows:
 
 ```json
 {
@@ -80,9 +80,9 @@ Upon accepting a connection a server MUST send a 'hello' message before any othe
 }
 ```
 
-"version" - The 'hello' message MUST contain a "version" property which MUST have a string value identifying the lowest version of signalk-delta that defines the format of the messages that will be sent by the server.
+"version" - The 'hello' message MUST contain a "version" property which MUST have a string value identifying the lowest version of signalk-delta that defines the format of the messages that will follow.
 
-"timestamp" - The 'hello' message SHOULD contain a "timestamp" property which MUST have a string value representing the date and time at the server that the 'hello' mesage was sent at. The client SHOULD compare the timestamp to the local time at the client at which the 'hello' is received and use the delta to correct the timestamp of subsequent signalk-delta messages recieved from the server.
+"timestamp" - The 'hello' message SHOULD contain a "timestamp" property which MUST have a string value representing the date and time at the node that node that sent the 'hello' mesage. The timestamp SHOULD be synchronised with UTC. If the timestamp is not synchronised with UTC then the receiving node SHOULD compare the timestamp with the time at the receiving node at which the 'hello' is received and use the delta to correct the timestamp of subsequent signalk-delta messages recieved from the sending node.
 
 "self" - The 'hello' massege MAY contain a "self" property which MUST have a string value that identifies the vessle the message pertains to.
 
--- a/gitbook-docs/urls_etc.md
+++ b/gitbook-docs/urls_etc.md
@@ -75,21 +75,14 @@ Upon accepting a connection a server MUST send a 'hello' message before any othe
 ```json
 {
   "version": "1.1.2",
-  "source": "actisense.ngw-1.1234567",
   "timestamp": "2015-04-13T01:13:50.524Z",
   "self": "123456789"
 }
 ```
 
-"version" - The 'hello' message MUST contain a "version" element which MUST have a string value identifying the lowest version of signalk-delta that defines the format of the messages that will be sent by the server.
+"version" - The 'hello' message MUST contain a "version" property which MUST have a string value identifying the lowest version of signalk-delta that defines the format of the messages that will be sent by the server.
 
-"source" - The 'hello' message MUST contain a "source" element which MUST have a string value that identifies the server as follows: manufacturer/publisher.device/application.uuid, where:
-* maufacturer is the name of a manufacturer or publisher in the list maintained in this repository
-* device is the name of a device or software application made by the manufacturer or publisher in the list maintained in this repository
-* uuid is a uneque identifier attributed to the physical device or installation of the application by the manufacturer/publisher
-When a source is not provided in a signalk-delta message the client SHOULD take the source from the 'hello' as a default.
+"timestamp" - The 'hello' message SHOULD contain a "timestamp" property which MUST have a string value representing the date and time at the server that the 'hello' mesage was sent at. The client SHOULD compare the timestamp to the local time at the client at which the 'hello' is received and use the delta to correct the timestamp of subsequent signalk-delta messages recieved from the server.
 
-"timestamp" - The 'hello' message SHOULD contain a "timestamp" element which MUST have a string value representing the date and time at the server that the 'hello' mesage was sent at. The client SHOULD compare the timestamp to the local time at the client at which the 'hello' is received and use the delta to correct the timestamp of subsequent signalk-delta messages recieved from the server.
-
-"self" - The 'hello' massege MAY contain a "self" element whcih MUST have a string value that identifies the vessle the message pertains to.
+"self" - The 'hello' massege MAY contain a "self" property which MUST have a string value that identifies the vessle the message pertains to.
 
--- a/gitbook-docs/urls_etc.md
+++ b/gitbook-docs/urls_etc.md
@@ -65,7 +65,7 @@ See [Subscription Protocol](subscription_protocol.html) for more details.
 
 #### Streaming Connection "Hello"
 
-Before a node sends any signalk-delta messages over a streaming connection it MUST send a 'hello' message. The format of the 'hello' message is as follows:
+Before a node sends any signalk-delta messages over a Web Socket or TCP connection it MUST send a 'hello' message. The format of the 'hello' message is as follows:
 
 ```json
 {
--- a/gitbook-docs/urls_etc.md
+++ b/gitbook-docs/urls_etc.md
@@ -3,7 +3,7 @@
 ### Short Names
 
 - `self` refers to the current vessel. Normally used in `vessels.self...`.
--### Ports
+### Ports
 
 The Signal K HTTP and WebSocket services SHOULD be found on the usual HTTP/S ports (80 or 443). The services SHOULD be
 found on the same port, but may be configured for independent ports and MAY be configured for ports other than HTTP/S.
@@ -13,7 +13,7 @@ A Signal K server MAY offer Signal K over TCP or UDP, these services SHOULD be o
 If an alternate port is needed it SHOULD be an arbitrary high port in the range 49152&ndash;65535[[2]](#fn_2).
 
 ### URL Prefix
-Tihe Signal K applications start from the `/signalk` root. This provides some protection against name collisions with
+aTihe Signal K applications start from the `/signalk` root. This provides some protection against name collisions with
 other applications on the same server. Therefore the Signal K entry point will always be found by loading
 `http(s)://«host»:«port»/signalk`.
 
@@ -23,8 +23,7 @@ The version(s) of the Signal K API that a server supports SHALL be available as 
 ```json
 {
     "endpoints": {        "v1": {
-            "version": "1.1.2",            "signalk-http": "http://192.168.1.2/signalk/v1/api/",
-            "signalk-ws": "ws://192.168.1.2:34567/signalk/v1/stream"
+            "version": "1.1.2",            "signalk-http": "http://192.168.1.2/signalk/v1/api/",            "signalk-ws": "ws://192.168.1.2:34567/signalk/v1/stream"
         },
         "v3": {
             "version": "3.0",
@@ -80,5 +79,5 @@ Before a node sends any signalk-delta messages over a streaming connection it MU
 
 "timestamp" - The 'hello' message SHOULD contain a "timestamp" property which MUST have a string value representing the date and time at the node that sent the 'hello' mesage. The timestamp SHOULD be synchronised with UTC. If the timestamp is not synchronised with UTC then the receiving node SHOULD compare the timestamp with the time at the receiving node at which the 'hello' is received and use the delta to correct the timestamp of subsequent signalk-delta messages received from the sending node.
 
-"self" - The 'hello' massege MAY contain a "self" property which MUST have a string value that identifies the vessle the message pertains to.
+"self" - The 'hello' massage MAY contain a "self" property which MUST have a string value that identifies the vessel the message pertains to.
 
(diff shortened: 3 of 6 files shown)


gitbook-docs/streaming_api.md
```diff
@@ -23,3 +23,17 @@ Upon connection a 'hello' message is sent as follows:
...
+#### Streaming: raw TCP sockets
+
+A signalk server MAY stream signalk-delta over a raw TCP socket.
+
+When signalk-delta is streamed over a raw TCP socket:
+
+1. The first character of each signalk-delta message MUST be the openning curly bracke (char(123), '{').
+2. Each signalk-delta message MUST be terminated by a carriage return and new line (char(10) + char(13), '\r\n') sent after the final closing curly bracket (char(125), '}').
+3. The server SHOULD send signalk-delta in a compact manner with any unnessesary white space removed.
+4. The server MAY also send sentences for other protocals (e.g. NMEA0183) interleaved with the signalk-delta messages provided that sentences of the other protocal:
```
jboynes: Mixing messages from multiple protocols adds a lot of ambiguity to the format and I'm not clear on the benefit.
Can we make this just a stream of JSON objects separated by /r/n where each object is a signalk-delta.

gitbook-docs/streaming_api.md
```diff
@@ -23,3 +23,17 @@ Upon connection a 'hello' message is sent as follows:
...
 }
 ```
+
+#### Streaming: raw TCP sockets
+
+A signalk server MAY stream signalk-delta over a raw TCP socket.
+
+When signalk-delta is streamed over a raw TCP socket:
+
+1. The first character of each signalk-delta message MUST be the openning curly bracke (char(123), '{').
```
jboynes: typo: opening curly bracket

The wording here is a bit ambiguous on whether the JSON object is surrounded by {} or not. Suggest:

The stream MUST comprise a set of compact JSON objects separated by a carriage return, new line sequence (char(13) followed by char(10), "\r\n"). Each JSON object SHOULD conform to the JSON schema for a signalk-delta message.

gitbook-docs/urls_etc.md
```diff
@@ -69,14 +68,28 @@ If a server does not support some streaming options listed in here it must respo
 
 See [Subscription Protocol](subscription_protocol.html) for more details.
 
-##### Connection Hello
+#### Streaming Connection "Hello"
 
-Upon connection a 'hello' message is sent as follows:
+Upon accepting a connection a server MUST send a 'hello' message before any other communicaiton is sent. The format of the 'hello' message is as follows:
```
jboynes: This works when there is specific connection setup (e.g. after opening a TCP connection) but won't allow a listener to tap into a stream in progress (e.g. if they start listening to a RS422 bus or a multicast session).

Instead, suppose we add a sync message to the delta stream where the server sends all or part of the signalk tree. A minimal set would be very similar to this but by sending the tree the server can provide the context to which subsequent deltas would apply and re-synchronize clients that are new listeners or whose version of the model may have drifted.

tkurki:
Terminology: is the word `raw` correct in this context? _Raw socket_ is not a TCP socket: https://en.wikipedia.org/wiki/Raw_socket. Is _raw TCP socket_ an oxymoron?

tkurki:
While it is good to consider similar transports we might want to concentrate on getting TCP done.  We can add stuff like periodic sync messages separately.

tkurki:
I first found the rule about beginning the line with { a bit odd but on second thought it sounded like a good idea. With very little effort we can add support for coexistence on the same port with NMEA0183 data.

The fact that the payload is Signal K delta in JSON format already states that SK data is in JSON format. However many JSON parsers allow whitespace before the first opening curly brace (just tried JS and Python) and in my opinion stating that the first character must be the opning curly brace and one can tell if it is NMEA0183 or SK delta from that is a good idea.

tkurki:
Formating: 

I always write Signal K as either 
- capitalised and separate: Signal K
- abbreviated: SK

Unless there are good reasons for not doing so I think we should use the Signal K format in the documentation.

gitbook-docs/connection.md
```diff
@@ -55,6 +57,7 @@ Using the information above a web client or http capable device can discover and
             "version": "1.1.2",
             "signalk-http": "http://192.168.1.2/signalk/v1/api/",
             "signalk-ws": "ws://192.168.1.2:34567/signalk/v1/stream"
```
tkurki: Missing trailing comma

gitbook-docs/urls_etc.md
```diff
@@ -69,14 +68,28 @@ If a server does not support some streaming options listed in here it must respo
...
 ```json
 {
   "version": "1.1.2",
+  "source": "actisense.ngw-1.1234567",
   "timestamp": "2015-04-13T01:13:50.524Z",
   "self": "123456789"
 }
 ```
+
+"version" - The 'hello' message MUST contain a "version" element which MUST have a string value identifying the lowest version of signalk-delta that defines the format of the messages that will be sent by the server.
```
tkurki: As JSON is JavaScript Object Notation and JavaScript objects have properties it might be prudent to refer to `properties` instaed of `elements`.

gitbook-docs/urls_etc.md
```diff
@@ -69,14 +68,28 @@ If a server does not support some streaming options listed in here it must respo
...
+
+"source" - The 'hello' message MUST contain a "source" element which MUST have a string value that identifies the server as follows: manufacturer/publisher.device/application.uuid, where:
+* maufacturer is the name of a manufacturer or publisher in the list maintained in this repository
```
tkurki: Do we want to start maintaining such a list?

gitbook-docs/urls_etc.md
```diff
@@ -69,14 +68,28 @@ If a server does not support some streaming options listed in here it must respo
...
   "version": "1.1.2",
+  "source": "actisense.ngw-1.1234567",
   "timestamp": "2015-04-13T01:13:50.524Z",
   "self": "123456789"
 }
 ```
+
+"version" - The 'hello' message MUST contain a "version" element which MUST have a string value identifying the lowest version of signalk-delta that defines the format of the messages that will be sent by the server.
+
+"source" - The 'hello' message MUST contain a "source" element which MUST have a string value that identifies the server as follows: manufacturer/publisher.device/application.uuid, where:
```
tkurki: This is not the same source as in definitions.

While I think that adding something like a `source` property may be a good idea I don't think it should be added as part of addition of TCP streaming, as it is not coupled to this issue.

gitbook-docs/urls_etc.md
```diff
@@ -69,14 +68,28 @@ If a server does not support some streaming options listed in here it must respo
...
+* maufacturer is the name of a manufacturer or publisher in the list maintained in this repository
+* device is the name of a device or software application made by the manufacturer or publisher in the list maintained in this repository
+* uuid is a uneque identifier attributed to the physical device or installation of the application by the manufacturer/publisher
+When a source is not provided in a signalk-delta message the client SHOULD take the source from the 'hello' as a default.
+
+"timestamp" - The 'hello' message SHOULD contain a "timestamp" element which MUST have a string value representing the date and time at the server that the 'hello' mesage was sent at. The client SHOULD compare the timestamp to the local time at the client at which the 'hello' is received and use the delta to correct the timestamp of subsequent signalk-delta messages recieved from the server.
```
tkurki: Format and UTC need to be specified.  

Maybe the wording should be that _The client MAY compare..._, as this greatly depends on the client.

gitbook-docs/urls_etc.md
jboynes: The source of the data here is the SK server itself. The type in definitions can represent that, I don't think we need/should add anything new.

jboynes:
Mixing SK messages with others seems confusing. An existing '183 client may be confused by the unexpected SK data, especially where the JSON contains '183 sentences (e.g in source properties). I don't see the advantage for a new implementation of a client as the SK feed can represent more data anyway. If a client wants both SK data and a '183 stream it can simply open two connections and not worry about demultiplexing (IP does that for you).

It's like mixing the raw binary output from an NGT-1 with '183 data: do-able but weird.

jboynes:
On whitespace, it would be simpler to either allow it everywhere or ban it entirely. Allowing it makes things more readable and fits naturally with JSON. The CR-LF delimiter is not needed to parse a stream of objects as the token sequence "}" "{" can't legitimately appear inside an object.

tkurki:
Anything but a simple delimiter that can not appear in valid JSON strings and we exlude from the message leads to unnecessary complexity.

Like do you allow whitespace between "}" and "{"? What constitutes whitespace? Multiple whitespace chars? All this requires a considerably more complex state machine than a delimiter. Complex state machines lead to errors and differences in implementations.

gitbook-docs/urls_etc.md
jboynes: To be a blt clearer, the 'hello' message comes from whatever is initiating the SK stream. This could be a client sending to a server, or a server responding to a client. In either case, the sender wants to identify themselves as the source of the information which is what the type in /definitions is for.

For example, a battery monitoring device discovers a SK server on its network, connects, sends a 'hello' identifying itself and then starts sending readings. It doesn't have a RTC so it sends a base timestamp of '0' and then time deltas in ms with the 'update' messages. The SK server includes information from the source value in its model and in delta messages it sends out.

A client discovers the SK server, connects to it and subscribes to a stream. The SK server identifies itself in the 'hello' and then forwards updates as it receives messages over SK, N2K or '183. The client uses the source information from the battery monitor to distinguish updates for the house bank from those from the engine bank.

A intermediate SK server can act as a forwarding proxy for SK delta without needing be configured with what every device it.

jboynes:
JSON parsers already have to handle whitespace, which is why I specifically said "token sequence" and not characters. Parsers for "concatenated JSON" already exist and handle this automatically. They tend to be event based, which would be needed on a small-footprint device anyway as it would not have the resources to buffer a potentially long line of text between CR/LF delimiters.

In terms of avoiding unnecessary complexity, the simplest definition would be to define the byte stream to be a simple sequence of JSON objects encoded using UTF-8 and avoid all the new-line delimited, multi-protocol stuff.

pod909:
That's beyond what is currently specified for the streaming interface and a
more complex issue than the scope of this ticket. Currently the stream is
defined as a pull interface with the source device is defined as a server;
with the node that wants the information connecting to it and receiving
deltas from it.

This ticket is concerned with the basics of using a raw TCP socket as a
transport instead of a websocket. Pushing deltas from one node to another
is a bigger subject and a separate issue should be raised. A proper
discussion would be needed on managing subscriptions etc. and there are
existing protocols to deal with that sort of thing.

e.g. How does a node know another node actually wants to be sent the info
it can produce?

...

The existence of the timestamp in the hello addresses a separate concern I
had with identifying the age of deltas sent by a node that doesn't have an
RTC.

When an SK aggregator connects to a node to get deltas it can compare the
hello timestamp to it's RTC time at the point the hello is received. That
delta can then be added to the timestamp in subsequent deltas to rebase
line the deltas to the local time on the aggregator. The aggregator should
not blindly apply a timestamp based on the time of receipt. That would be
highly inaccurate for trivial examples like boat speed from a paddle wheel
and apparent wind from a mast head unit.

pod909:
One mans unnecessary is another mans necessary. Supporting multiple
protocols on a single socket vs writing multi threaded firmware to provide
separate ports for different protocols....

gitbook-docs/streaming_api.md
pod909: typo corrected

gitbook-docs/urls_etc.md
pod909: Which ISO standard is used else where and I'll add.

The timestamp cannot be restricted to UTC. For devices that do not have an RTC it needs to be left open as the local time maintained by the node. I think it's stronger than MAY. The chances of all the nodes in a network being time synchronized and there being not latency in the communication in all instances is slim. Nearly suggested is a MUST in fact.

gitbook-docs/urls_etc.md
pod909: Source removed form this update

Will raise a separate issue as I view this as vital.

gitbook-docs/urls_etc.md
pod909: Most nodes dealing with signal-delta wont have the full tree

gitbook-docs/connection.md
pod909: corrected

gitbook-docs/urls_etc.md
pod909: corrected

gitbook-docs/urls_etc.md
pod909: Should be no overhead to maintaining the list. Manufacturers/publishers would just add to it them selves under version control here. Documented as a JSON object it would provide a useful resource for validating messages.

gitbook-docs/urls_etc.md
pod909: What ISo standard is used and I'll update.

UTC can't be mandated as some nodes -- in particular low function node from which values originate -- do not have an RTC. They should all be able to provide a relative time stamp/"local time at the node" for each sk-delta though. Nearly suggested the client making a timestamp adjustment as a MUST, so SHOULD is a reasonable compromise I think.

pod909:
Happy to change the delimiter to }<<white space>>{ if agreed on

tkurki:
While I really appreciate the discussion I really wish that people would have voiced their thoughts earlier in #147 which was raised already early this year.

@jboynes I think your point about mutual / two way hello is valid, thanks for raising it. However that is not the most crucial point in TCP streaming and applies to WebSockets as well, so I think the source handling should be handled separately.

jboynes:
@tkurki the mixed protocols and source parts of this PR weren't mentioned in #147. I'll open a new discussion thread for the source portion now that @pod909 has removed it from this one.

The current [streaming spec](http://signalk.org/specification/master/streaming_api.html) simply says the server sends a stream of SK delta messages and the [model spec](http://signalk.org/specification/master/data_model.html#encodingdecoding) talks about resource limited devices using streaming parsers. For those reasons, my feedback here is not to add the requirement for using the newline format and the additional support for mixed protocols. If we see a need for the latter, let's specify it properly and have a larger discussion about how to distinguish between the different messages.

pod909:
Support for mixed protocols came up in the discussion on a delimited.
Makes no odds really. \r\n is white space, although the "junk" of the alternative message wouldn't' be.

sumps:
Can anyone provide a technical reason, benefit or use case for a TCP socket transmitting multiple protocols? I am struggling and therefore finding it very difficult to give this any priority.

That said I am very interested in a solution for Signal K data over a serial connection and it would be great if serial could share the same format as TCP, which should be possible.

sumps:
I like the idea of the Timestamp being in the hello message as even more powerful devices like iKommunicate do not have an RTC and if it is not connected to a GPS cannot generate its own correct time/date timestamp. This would allow iKommunicate to sync its timestamp with an SK Server.

If we had a two way hello message method it would allow a client running on a mobile device that has an accurate date/time to sync the iKommunicate timestamp.

pod909:
Low power/function device that only wants to maintain a single socket but
provide 0183 and SK-delta. Don't feel strongly enough about it to argue it.

Maintaining data integrity is an issue when communicating over a serial
connection. You'd want to add a checksum. Compression (LZW) would probably
need to be looked at as well to get decent through put out of 115k

pod909:
Shouldn't need to sync the times on the 2 devices provided the client
adjusts the timestamp.

e.g. A paddlewheel (server) sending delta to iKomm (client)...

Neither has an RTC, but they both have clocks.

1. Paddlewheel receives a connection request from iKomm 10 mins after it
booted up
2. Paddlewheel sends a hello with "00/00/00T00:10:00"
3. iKomm receives the hello at "01/02/2016T10:23:45" and stores the
difference between the timestamp in the hello and the time it thinks it
received it.
4. First delta arrives from the paddle wheel with a timestamp of
"00/00/00T00:10:01"
5. iKomm uses the difference found at 3 to update the timestamp to
"01/02/2016T10:23:46"

The same adjustment would happen when the sk-delta was passed on tot he
next node.

Mention was made of devices that sleep. They would just need to resend the
hello when they woke up.

sumps:
Yes the checksum would need to be after the last } and before the \r\n. Don't think LZW is a good solution, think it is better to naturally limit the use of serial SK by the available bandwidth at 115k.

For some applications it will be a suitable transfer mechanism and for others TCP or WS will be the best method.

pod909:
That's going to invalidate using '} white space {' as a delimiter as the
there are instances that the checksum would look a like a comma.

For the checksum there's no need for one for WS and TCP and I don't think
one should be specified when it's not needed.

For serial and UDP a CRC-32 looks the most commonly implemented option.

---

For TCP I guess it makes implementation of senders that rely on standard
JSON libraries more difficult so may be the \r\n is a SHOULD or MAY rather
than a MUST... but then why bother at all.

tkurki:
>Parsers for "concatenated JSON" already exist and handle this automatically

@jboynes That argument could be used for a multitude of formats. Furthermore it requires you to use a parser / more complex state machine, if whitespace use is free. I think that "parse each line delimited with \r\n as JSON" is easier to communicate, understand and implement. It still allows you to use an evented parser. 

Having said that I admit that the thought that `} {`could be used to break the stream to messages never occurred to me and is viable. One possibility is that we skip the coexistence with 0183 data and just state that messages are JSON objects (not JSON arrays or JSON values) with \r\n between individual messages? This would effectively make `}\r\n{` the message boundary and allow free whitespace use within the messages.

The drawback is that the same format would not work as easily for serial with checksums in the way discusses in #147. For serial there needs to be a way to hop on board in the middle of the stream and things get more complicated with a checksum. A simple solution to that again is to make the message single line, making serial different from TCP streaming in this respect. There is value in making different streaming formats similar. And since WebSockets is message based it is by definition different from real streaming.

Also this is not totally about parsing but overall ease of use. I find line oriented JSON straightforward to process with different tools, such as `grep` and `less` (or Word and Excel). Processing a  stream/file of pretty printed JSON objects requires using a real parser.

@sumps The one theoretical case I can think for streaming different formats over the same channel is when you want to maintain order over all messages, but I don't think it applies here.

So while there is value in allowing whitespace use within the JSON I find more value in the more simplistic _single line \r\n message delimiter_ solution which can be extended for serial.

In what use case would `} {` be superior?

pod909:
I think the key thing is that a \r\n is white space and would not prevent
the use of "concatenated JSON parser"

pod909:
Opening { as first char changed form MUST to SHOULD
Requirement for ability to handle mixed protocol streams removed.

pod909:
Separate issue raised for serial connections with a view to closing this PR for TCP.

jboynes:
@tkurki In order to process JSON safely you need a parser to handle Unicode character decoding (from UTF-8 or whatever), escape sequences, arbitrary property ordering etc. Line oriented text tools like grep, awk, vi or Word are great for debugging but error prone for production use.

I agree with suggesting that the sender emit newline separated objects to make debugging easier but don't think we should require it. 

I don't understand what you mean when you say WebSockets is message based. I suggest we continue the discussion on checksums in the serial thread.

tkurki:
>Line oriented text tools like grep, awk, vi or Word are great for debugging but error prone for production use.

I was talking about developer productivity.

tkurki:
https://en.wikipedia.org/wiki/WebSocket
>WebSocket enables streams of messages on top of TCP. TCP alone deals with streams of bytes with no inherent concept of a message.

I was thinking out loud why what we are discussing here doesn't apply to WebSockets.

tkurki:
I am happy with this PR and would be willing to merge. 

Something that we have discussed informally is that all specification additions should have at least some implementation, preferably more, so that we excercise the addition a bit and not just add more ivory tower stuff with little or no connection to reality. Towards this end I modified the existing node server tcp interface to comply with this spec, see https://github.com/SignalK/signalk-server-node/pull/124

pod909:
Final call for comments.
This PR will be merged to the main branch on the 27th November.

sumps:
Sorry for my ignorance but in JSON, are \r \n sent as ASCII control characters 10 and 13 or as four ASCII characters?

The reason I ask is that many Serial to Ethernet converters have a function to close a TCP or UDP packet when they receive a particular character such as <LF> (ASCII Character 10). This means that you get a complete sentence or in this case Delta message in one network packet rather than split across multiple packets which can create serious problems in UDP mode.

If \r \n are sent as those four ASCII characters then this will be impossible for the serial to Ethernet adaptor to detect amongst the other data.

tkurki:
This has nothing to do with what goes into the JSON, just the delimiter between the JSON messages.

Two bytes:
- \r means carriage return (decimal 13, hex 0D)
- \n means linefeed (decimal 10, 0A)

\r\n is just another way to say <CR><LF> - the same message delimiter as in NMEA0183

Another way to say this is
- one delta per line
- line delimiter \r\n
- nothing but the line delimiter between deltas

pod909:
will make sure this is clearer in the spec

just writing a reference implementation for a sensor with an Arduino HAL at
the mo

sumps:
OK I am happy with this and I would make the \r\n mandatory along with an implementation note about closing TCP/UDP packets with the Linefeed character as it will avoid problems with the majority of serial to Ethernet adaptors in the market today.

This is the only problem I have seems with data over TCP/UDP.

jboynes:
@sump Ethernet, TCP and UDP are all binary protocols that do nothing special with line break characters. You may be thinking of serial port emulators that mimic an old-fashioned TTY on top of these to give a command-line like interface for humans.

@tkurki NMEA0183 has other delimiters as well that allow sending sentences over an unreliable serial link (the !$ introduction, use of * to denote checksum). Its use of a line break makes it easier to recover from lost characters. We do not have have any of those problems here.

I prefer to leave this unchanged from the current stream format where sending these characters is optional.

pod909:
@sumps, are you finding that the \n is still sent?

pod909:
.. it's still sent with the WiFi SOC I'm using. The delimiter can be
configured or completely disabled as well.

tkurki:
>unchanged from the current stream format 

The current stream format, deltas over WebSockets, has no use for message delimiters, because each delta is an individual message.

While not strictly necessary using \r\n makes the data easier to work with.

pod909:
Having a consistent approach to message format across transports makes
implementation a lot more consistent, even if that means sending a couple
of theoretically unnecessary but harmless chars over WS and TCP. So far
that just leaves a checksum for serial communication as the only oddity.

tkurki:
@pod909 I think that the addition in 2732fd9d4e7409f7e3fe2897068002e75dfeab73 adds more confusion than it clears things up.

Like, if I am writing a SK TCP client or server how does it affect my work? As far as I can see in no way.

pod909:
Updated again.

Trying to find language that avoids "Server" terminology and allows for
streaming in either direction.

gitbook-docs/streaming_api.md
```diff
@@ -23,3 +23,15 @@ Upon connection a 'hello' message is sent as follows:
...
+
+A network node MAY stream signalk-delta over a TCP socket.
+
+When a hello or signalk-delta message is sent over a TCP socket:
+
+1. Each message MUST be terminated by a carriage return and new line (char(10) + char(13), '\r\n') sent immediatly after the final closing curly bracket (char(125), '}').
+2. The first character of each message SHOULD be the openning curly bracket (char(123), '{').
+3. The message SHOULD be sent with any unnessesary white space removed.
+
+Note on hardware implementation: The serial APIs for ethernet and wifi modules often close a TCP or UDP packet on receipt of a new line ( char(13) ). Please check the data sheet and ensure the new line character is still sent.
```
jboynes: Do you have an example of this? That's not how TCP and UDP work normally.

gitbook-docs/streaming_api.md
```diff
@@ -23,3 +23,15 @@ Upon connection a 'hello' message is sent as follows:
...
 }
 ```
+
+#### Streaming: TCP sockets
+
+A network node MAY stream signalk-delta over a TCP socket.
+
+When a hello or signalk-delta message is sent over a TCP socket:
+
+1. Each message MUST be terminated by a carriage return and new line (char(10) + char(13), '\r\n') sent immediatly after the final closing curly bracket (char(125), '}').
```
jboynes: s/char(10) + char(13)/char(13) + char(10)/

gitbook-docs/streaming_api.md
```diff
@@ -23,3 +23,15 @@ Upon connection a 'hello' message is sent as follows:
...
+2. The first character of each message SHOULD be the openning curly bracket (char(123), '{').
```
jboynes: s/openning/opening/

gitbook-docs/urls_etc.md
```diff
@@ -14,19 +13,16 @@ A Signal K server MAY offer Signal K over TCP or UDP, these services SHOULD be o
...
 
 ### Versioning
 
 The version(s) of the Signal K API that a server supports SHALL be available as a JSON object available at `/signalk`:
-
 ```json
 {
-    "endpoints": {
-        "v1": {
+    "endpoints": {        "v1": {
```
jboynes: Typo?

gitbook-docs/urls_etc.md
```diff
@@ -80,3 +76,10 @@ Upon connection a 'hello' message is sent as follows:
   "self": "123456789"
 }
 ```
+
+"version" - The 'hello' message MUST contain a "version" property which MUST have a string value identifying the lowest version of signalk-delta that defines the format of the messages that will follow.
+
+"timestamp" - The 'hello' message SHOULD contain a "timestamp" property which MUST have a string value representing the date and time at the node that sent the 'hello' mesage. The timestamp SHOULD be synchronised with UTC. If the timestamp is not synchronised with UTC then the receiving node SHOULD compare the timestamp with the time at the receiving node at which the 'hello' is received and use the delta to correct the timestamp of subsequent signalk-delta messages recieved from the sending node.
```
jboynes: s/recieved/received/

gitbook-docs/urls_etc.md
```diff
@@ -80,3 +76,10 @@ Upon connection a 'hello' message is sent as follows:
...
+
+"self" - The 'hello' massege MAY contain a "self" property which MUST have a string value that identifies the vessle the message pertains to.
```
jboynes: s/massege/message/
s/vessle/vessel/

jboynes:
With a WS connection, the client can specify what messages it wishes to be sent in the WS URL using the `subscribe` query parameter. No similar mechanism exists when opening a simple TCP connection? What is the default behaviour? For example, 
- all changes
- everything applying to `vessels.{vesselId}`
- no changes until a subscription message is sent upstream
- something else?

I would recommend "no changes until subscription" as that prevents the risk of flooding the TCP connection with unneeded data, especially when bandwidth or client resources may be limited. NOte this is _different_ to the WS protocol which defaults to "everything for the vessel" but which avoids the problem by allowing the client to specify the initial subscription.

gitbook-docs/urls_etc.md
pod909: not part of this change so can't comment

gitbook-docs/urls_etc.md
pod909: corrected (thanks for the review)

gitbook-docs/urls_etc.md
pod909: corrected

pod909:
@jboynes For this change "none of the above". No mechanism is suggested for deciding what should be sent as part of this change. As a result the receiver must handle anything the sender sends. That may be values across the whole domain, it may not.

Certainly worthy of a separate discussion. I'm not sure a straight TCP connection is the best option if throttling is required or if its best done at the messaging level if it is.

sumps:
@jboynes no confusion on my side, I am talking about a specific but very common use case where a serial data stream is converted to TCP or UDP packets and the line feed character is used to tell the adaptor to close and transmit the network packet. If you use the other two mechanisms; number of characters or time period you run the risk of sentences or massages being spread across multiple packets and in the case of UDP packets can (and I have send this first hand) be received out of order with bad results.

sumps:
I would also request that a Hello message is not mandated but recommended as it is only possible with TCP and not UDP.

pod909:
Could get into a discussion on "sends" vs "receives", but for now I've added clarification that the hello is only mandated for WS and TCP.

UDP isn't covered in this update. It probably needs separate discussion to make sure everything is captured before it's included.

jboynes:
@sumps That was what I meant by "serial port emulators" that are trying to mimic the behaviour of a TTY (such as waiting for a CR to denote end-of-user-input or to scroll a display up) on top of a data stream.

At the TCP or UDP level **none of that applies**. When [send(buffer)](https://linux.die.net/man/2/send) is called the data will be sent[1], irrespective of any newline characters in the buffer. TCP guarantees the bytes in the stream will be delivered to the other end, reliably, and in order, no matter how the stream gets fragmented into packets for transmission. 

UDP is a datagram protocol and not relevant to this PR, see #289. UDP datagrams do get fragmented at the IP or Ethernet layer but UDP guarantees that its datagram gets reassembled correctly. It does not guarantee that it will be delivered, or the order in which different datagrams get delivered.

I think crux of the matter is whether we are talking about a TTY-style protocol here or **raw TCP.**

[1] Subject to TCP_NODELAY and other socket options but that's not particularly relevant to how TCP does not care about newlines in data

jboynes:
I hacked out a [prototype](https://github.com/jboynes/SKgateway/blob/master/server/src/test/java/org/skgateway/server/ServerTest.java) using the low-level socket interfaces which shows that the newline stuff is not _needed_. Tracing with Wireshark & netcat shows the TCP segments being sent when the application calls write() even though there are no newlines.

IOW, the concerns about needing newlines to make TCP work as expected are not valid.

I think everyone here thinks the sender _should_ send out compact messages separated by newlines to make the raw stream easier for humans to read e.g. with netcat.

I think the difference is I that I think receivers should not be able to rely on that (with the implication that senders _must_ add newlines and not emit whitespace). I think there's more flexibility if receivers must only rely on receiving a stream of JSON objects with no special handling for whitespace (per the JSON spec).

sumps:
@jboynes we are still not on the same page and perhaps I have not made myself clear.

Use case - a simple and cheap SK sensor outputs SK over serial with a checksum and \r\n (<CR><LF>) end of message characters. Then someone wants to connect the serial data to a serial to ethernet adaptor (wired or wireless) and what I want to make sure of is that we do not make this impossible by insisting on a mandatory "Hello" message and not allowing a TCP or UDP SK message to have a checksum or  \r\n characters.

pod909:
The hello is currently mandatory so the question @tkurki is why? The only mandatory content is the version.

Checksum. Makes it easier for the dumb bridge but more complicated for what ever receives the delta from TCP/WS as it's not going to get valid JSON anymore. The \r\n is harmless as it's just seen as white space. If we have to allow for a checksum everywhere to cater for this then I'm definitely on the side of a decimal.

pod909:
@sumps  not sure that's actually a valid use case. A device that simple could well end up putting rubbish onto TCP/WS. It needs as a minium to check the checksum, why bother with it otherwise.

sumps:
This is a very common use case, with most of the current generation of wireless nmea devices using this approach. The DIY user would be even more likely to do this as you do not have to write any code, just buy a module that does the serial to ethernet conversion.

pod909:
Well "you might be getting crap" is a bit of a default position with NMEA0183  and the checksum was designed in with it being a serial protocol in essence. Not the same with here. Serial support is the edge case.

Doing the checksum even remotely properly requires the messages to be buffered, so stripping the checksum is no effort compaired to the other work that should be done. Once it's on to TCP mandatory buffering and checksum validation at every step, just in case crap was sent out from a serial connection? That's a serious imposition.

sumps:
I don't want to get in to an arguement over the benefits of one solution over another, I am just highlighting a popular use case that allows for cost effective data transmission, that we should not make impossible by making certain things mandatory.

tkurki:
Hello is mandatory in ws because we thought it would be a good idea to define a message that conveys some information about the server that the client is connecting to and that can be expected and is distinguishable from deltas.

jboynes:
@sumps We have #287 to cover sending SK over serial which deals with the additional packetization and framing a low level data link like that needs. The serial-ethernet-converter application is about transporting that over an IP network and the applications should treat it as such; for example, it can't rely on TCP's guarantees as data can be lost on the serial segment.

The case I think we are reviewing here is where the device talks TCP natively and hence is able to rely on the properties of a TCP connection in terms of reliability and ordering. That makes using the connection simpler as TCP is doing all the work of framing, checksumming and recovering from lost packets.

As a DIY user it's easy and cheap to pick up a [controller](https://www.adafruit.com/products/2821) able to talk Ethernet or Wifi directly, with TCP support built in. For example, the Arduino EthernetClient makes it [about as easy](https://www.arduino.cc/en/Reference/ClientConnect?action=sourceblock&num=1) as using the Serial library.

pod909:
@tkurki is it the node that is due to send or receive the deltas that's supposed to send the hello??

tkurki:
I don't really get why you are trying to avoid the words _server_ and _client_: the tcp connection is initiated by the client by connecting to the socket that the server is already listening to. There is clearly a server, that waits for incoming connections. The parties are not symmetrical in terms of the TCP connection, even if the tcp and ws connections are duplex.

The server sends the hello and informs what SK version it uses. The client then has the option to disconnect, if it can not support the version that the server speaks. 

So far the usual arrangement has been that there is a server, that provides one or more vessel's data when connected to. 

I believe java server accepts deltas from a ws client, but I know node server doesn't. If needed that capability can be added easily. 

With the duplex capability we can devise different setups, like
- a display connects to the boat's SK server and starts streaming updates that the server receives from different inputs
- a sensor connects to the boat's SK server and starts streaming updates to the server, doesn't care about the server's data (so maybe would like to tell the server that)

Both of these case assume a "master" server on the boat.

pod909:
Everything gets flipped on its head when you talk about sensor providers and Servers consumers, rather than Servers providers and browser consumers.

The following use cases have been suggested:
1. The sensor is a _TCP server_ and the Signal K Server will connect to it as a _TCP client_ to pull messages from it (this is the model described by the WS standard at the moment)
2. The sensor is a _TCP client_ and connects to a _TCP server_ at the Signal K Server in order to push messages to it.
3. Bi directional traffic

So some times clients are providers; some times servers are providers. I also wanted to avoid using the word "server" as the reference aggregator/broker has been named Signal K Server.

I'd assumed (apologies) that the _provider_ would send a "hello" and then the _consumer_ would decide whether it wanted to read what followed or disconnect and refuse future connection if not. If the _consumer_ found the _provider_ through discovery it now knows not to bother connecting again. This only requires a mono directional connection and the "hello" can carry more interesting stuff like local timestamp and default context, device and source information.

If the _consumer_ says hello that assumes bi directional communication and requires the _provider_ to handle receiving the "hello".

pod909:
So...

It looks like we have 2 different services here.

_signalk_stream_serial._tcp which must satisfy the conditions for a serial connection with limitations on data integrity that consumers must cater for.

And

_signalk_stream_tcp._tcp which must satisfy the conditions here.

In the absence of discovery a consumer will be able to differentiate between streets based on receipt of a hello as the first message.

I'm also going to expand the hello to include a list of paths where a subscription request is required to start receiving a stream.

pod909:
I think it would also be wise to retract this PR and document a proposal separate from the current WS standard.

timmathews:
We're definitely deep into RFC territory here. I agree that it makes sense to close this PR, write up the current state of the proposed TCP layer and submit that as an RFC.

That said, for clarity of communication going forward, here are some definitions I think we can use:

- *client*: a client is a device, program, etc. which initiates a connection to a server. Clients do not expose a signal k service on a well defined port, but rather open an ephemeral port for the duration of the connection
- *server*: a server is a device, program, etc. which responds to a connection request, and more importantly has a signal k process listening on a well defined port

This is all TCP/IP 101, and @tkurki said as much above, but I think it bears repeating.

I do not see the value in having two different TCP/IP services, it will lead to confusion and compatibility issues down the road. As if having a WebSockets service, a TCP/IP service, a UDP/IP service, and a message broker (MQTT?) service isn't already fragmented enough. We need to pick one and move forward with it.

pod909:
No problem Tim. Agree and will put forward an RTC for a consistent approach
to streamed updates based on the discussion here.

I'm going to leave a number of options open on the basis that adoption may
be the best arbiter.

I'll also commit to produce an MIT licenced reference sensor implementation
with an Arduino HAL. Assistance most welcome.

pod909:
RFC 0005 created: closing this PR

PR #290, opened 2016-11-09
Make processSchemaFiles produce keyswithmetadata.json

tkurki:
Produce keyswithmetadata with the same logic as keys.md is produced.


tkurki:
@sailoog This change will make all the paths in keyswithmetadata contain the full path, not just the path starting at a vessel's root.

Sorry about the accidental notifications commit in this branch, please ignore it.

PR #295, opened 2016-11-23
Added commonValueFields to ativeRoute

sbender9:


tkurki:
Fixed by #313 afai can see.

PR #299, opened 2016-12-03
Generate documentation for object types

### LINKED ISSUE #273: Object valued properties structure in docs keys list
Examples:

http://signalk.org/specification/master/keys/#vesselsregexpnavigationposition
http://signalk.org/specification/master/keys/#vesselsregexpenvironmentcurrent

There should be information about the structure and property units.

joux3:
This pull request tries to fix #273.

Keys documentation in master: [click here](https://www.niksula.hut.fi/~ristela1/signalk-documentation-master/keys/)
Keys documentation generated by this PR: [click here](https://www.niksula.hut.fi/~ristela1/signalk-documentation-with-object-documentation/keys/)
See for example `/vessels/<RegExp>/navigation/position`.

There are actually multiple parts to my changes but laziness & explorative coding led to just a single commit. The parts are:
1. replace `json-load-tree` with `json-schema-ref-parser` to
   - load schema files locally
   - let a library handle references instead of the previous `resolveReference`
2. fix `allOf` handling to properly handle more difficult cases
3. change Markdown generator to generate documentation for object types (+ fix a couple of already existing bugs)

I didn't verify if the `keyswithmetadata.json` generator still works. @tkurki? Bugs/missing documentation fields would also be nice to be discussed before merging. The leaf logic is probably also still off.

### DIFF
--- a/keyswithmetadata.json
+++ b/keyswithmetadata.json
@@ -643,23 +643,8 @@
   "/vessels/*/electrical/batteries/RegExp": {
     "description": "Batteries, one or many, within the vessel"
   },
-  "/vessels/*/electrical/batteries/RegExp/associatedBus": {
-    "description": "Name of BUS device is associated with"
-  },
-  "/vessels/*/electrical/batteries/RegExp/voltage": {
-    "units": "V",
-    "description": "Data should be of type number."
-  },
-  "/vessels/*/electrical/batteries/RegExp/voltage/ripple": {
-    "units": "V",
-    "description": "Ripple voltage"
-  },
-  "/vessels/*/electrical/batteries/RegExp/current": {
-    "units": "A",
-    "description": "Data should be of type number."
-  },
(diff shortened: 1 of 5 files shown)


tkurki:
keyswithmetadata.json is in git - please generate it with the code in PR and add it here => Github will provide a diff.

tkurki:
Diff between old and new https://gist.github.com/tkurki/2663e91cd365529f2c23e162e01270e7

Some obvious degradations and naturally marked improvements.

I think some of these are caused by description from definitions overriding the description in context.

joux3:
Now the earliest definition for string fields (for example `description` and `title`) should stay. This also means that in schema the most useful values for these fields should be defined first (see for example `/vessels/<RegExp>/electrical/batteries/<RegExp>/temperature`).

I also updated [the generated documentation](https://www.niksula.hut.fi/~ristela1/signalk-documentation-master/keys/).

PR #300, opened 2016-12-04
Expand on the Discovery Section

timmathews:
Clarify mDNS/Bonjour and explain that the process being used is DNS-SD
via mDNS. This should give anyone who needs to implement this on a
platform without a ready-to-hand mDNS client library to get started in
the right direction.

Describe the fields or attributes which MUST, SHOULD and MAY be present
in the TXT record of the service advertisements and update the examples
to match.

Add a link to the section of the documentation which explains the
endpoints API and remove the example endpoint JSON as it is already
described in the other section.

### DIFF
--- a/gitbook-docs/connection.md
+++ b/gitbook-docs/connection.md
@@ -2,34 +2,29 @@
 
 ## Service Discovery
 
-A Signal K server SHOULD advertise its services using DNS Service Discovery (DNS-SD) via Mutlicast DNS (mDNS); also
-known as Bonjour. The server MUST provide DNS [Service (SRV) Records](https://en.wikipedia.org/wiki/SRV_record) and
-[Text (TXT) Records](https://en.wikipedia.org/wiki/TXT_record) describing the Signal K interfaces it provides. These
-service identifiers are:
+A Signal K server SHOULD advertise its services using [DNS Service Discovery
+(DNS-SD)](https://tools.ietf.org/html/rfc6763) via Mutlicast DNS (mDNS); also known as Bonjour. The server MUST provide
+DNS [Service (SRV) Records](https://en.wikipedia.org/wiki/SRV_record) and [Text (TXT)
+Records](https://en.wikipedia.org/wiki/TXT_record) describing the Signal K interfaces it provides. These service
+identifiers are:
 
 * `_http._tcp` and/or `_https._tcp` for the server's web interface
 * `_signalk-http._tcp` and/or `_signalk-https._tcp` for the Signal K REST API
 * `_signalk-ws._tcp` and/or `_signalk-wss._tcp` for the WebSocket data stream
 
-One of the following parameters MUST be set, both MAY be set:
+If a Signal K server is using DNS-SD, it MUST provide the following parameters (key/value pairs) in the TXT record
+portion of the DNS-SD advertisement:
 
-* `vessel_uuid` is the Signal K UUID which uniquely identifies the vessel
-* `vessel_mmsi` is the MMSI assigned to the vessel
+* `txtvers` is a US-ASCII decimal number identifying the version of the DNS-SD record. Currently, this MUST have a value
+  of 1
+* `roles` specifies which roles the server is capable of providing. See [Roles](#roles) below for details
 
-A vessel may have one or both of these values. One or both MUST be available when advertising the Signal K server.
+The server MAY provide the following values:
 
-The following parameters SHOULD be set in the TXT record:
-
+* `self` is the unique identifier of the vessel using the URN format specified for the `uuid` field in the Signal K
+  schema
 * `server` is the name of the Signal K server software, e.g. signalk-server-node
 * `version` is the version of the Signal K server software
-
-These parameters identify the server and its version. They are purely informational.
-
-The following parameters MAY be set in the TXT record:
-
-* `vessel_name` is the registered name of the vessel
-* `vessel_brand` is the name of the manufacturer of the vessel
-* `vessel_type` is the model of the vessel
 
 An example DNS-SD record set might look something like this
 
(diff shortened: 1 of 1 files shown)


gitbook-docs/connection.md
```diff
@@ -1,63 +1,76 @@
...
+        'vessel_name=Volare',
+        'vessel_brand=Friendship',
+        'vessel_type=22',
+        'vessel_mmsi=503123456',
         'vessel_uuid=urn:mrn:signalk:uuid:6b0e776f-811a-4b35-980e-b93405371bc5',
-        'version=v1.0.0',
-        'vessel_name=urn:mrn:signalk:uuid:6b0e776f-811a-4b35-980e-b93405371bc5',
-        'vessel_mmsi=urn:mrn:signalk:uuid:6b0e776f-811a-4b35-980e-b93405371bc5',
-        'server=signalk-server',
-        'path=/signalk'
```
timmathews: The node server doesn't include `path=/signalk`, so I removed it from the example. Should the server be providing it?

gitbook-docs/connection.md
```diff
@@ -1,63 +1,76 @@
...
+        'version=0.1.23',
+        'server=signalk-server'
```
timmathews: Updated the list of attributes here to match what the node server currently publishes

timmathews:
Having read through the RFC for DNS-SD, I've learned a few things about how we should use the protocol.

1. The key SHOULD be no more than nine characters long (the `vessel_*` keys violate this)
2. The total size of a typical DNS-SD TXT record is intended to be small -- 200 bytes or less. (we're pretty close to this)
3. A `txtvers` key is recommended to be included to indicate the version of the TXT record. It should be the first key in the TXT record.
4. For a well-designed application protocol, even if there is no information at all in the TXT record, it should be possible, knowing only the host name, port number, and protocol being used, to communicate with that listening process and then perform version- or feature-negotiation to determine any further options or capabilities of the service instance.

So, based on this, I think we should drop the `vessel_*` keys, they're kind of redundant with `self` and this information can be retrieved by clients using the REST API. And we should add a `txtvers` field specified to  always have a value of `1` for now.

A complete record should look like:

```
txtvers=1
roles=master,main
self=urn:mrn:signalk:uuid:c0d79334-4e25-4245-8892-54e8ccc8021d
server=signalk-server
version=0.1.23
```

this makes for a much more compact 128 bytes, comfortably fitting inside one DNS packet. The order above would be the preferred order of keys, specified from most to least significant.

PS. Regarding point 3, using the mdns Node package, can we specify the order of the keys? Since they're generated from an object, there is no guarantee that they will be in a specific order (as I understand JavaScript).

tkurki:
I believe Node/v8 honors insertion order for non numeric property names.

sumps:
Can we change server to name as other SK devices will use Discovery not just servers and if it is the network name it means we can use the same settng for both

timmathews:
Per the DNS-SD RFC 6763 Section 6.3 Paragraph 3, host name and port number must not be in the TXT record (they already exist in the SRV record).

However, I agree that `server` isn't the clearest. Along with `version` it should identify the software running on the device. How useful is that? I have no idea.

If we choose to keep those parameters, `swvers` and `swname` make more sense as they're less ambiguous than the current params.

pod909:
The current service names are a little open ended with some assumptions made about the mix of protocol and transport. For instance it's reasonable to send delta messages over http. Can we be a little more explicit with out requiring an http implementation as well as dns-sd?

I'm thinking something like:
_signalk-rest-http._tcp
_signalk-stream-ws._tcp

pod909:
Should vessels_uuid and vessels_music not be a single key called 'context'?

gitbook-docs/connection.md
```diff
@@ -1,63 +1,97 @@
...
+portion of the DNS-SD advertisement:
 
-A sample Bonjour record output, dumped using avahi-discover:
+* `txtvers` is a US-ASCII decimal number identifying the version of the DNS-SD record. Currently, this MUST have a value
+  of 1
+* `roles` specifies which roles the server is capable of providing. See [Roles](#roles) below for details
+
+The server MAY provide the following values:
+
+* `self` is the unique identifier of the vessel using the URN format specified for the `uuid` field in the Signal K
```
tkurki: Self can be mmsi or url as well.

gitbook-docs/connection.md
```diff
@@ -1,63 +1,97 @@
...
+(DNS-SD)](https://tools.ietf.org/html/rfc6763) via Mutlicast DNS (mDNS); also known as Bonjour. The server MUST provide
+DNS [Service (SRV) Records](https://en.wikipedia.org/wiki/SRV_record) and [Text (TXT)
+Records](https://en.wikipedia.org/wiki/TXT_record) describing the Signal K interfaces it provides. These service
+identifiers are:
 
-* `_signalk-http._tcp` for http API
-* `_signalk-ws._tcp` for WebSocket
-* `_signalk-https._tcp` for HTTPS API
-* `_signalk-wss._tcp` for secure WebSocket
+* `_http._tcp` and/or `_https._tcp` for the server's web interface
```
tkurki: http://www.dns-sd.org/serviceTypes.html

>Rather than having an entirely different DNS-SD service type for https, we recommend using the same redirection mechanism: advertise a plain "http" service, which consists of nothing except an HTTP redirection to the desired "https" URL.

tkurki:
@timmathews I chose to comment instead of making changes directly - let me know if you want me to do direct edits.

tkurki:
@pod909 First a general comment: Signal K is at a stage where there is a non trivial amount of software out there. When you suggest changes please evaluate the change as to how big a change impact does it have on the current software.
- will existing servers/gateways work or require change?
- will existing clients work or require change?
- is the change backwards compatible or will it in effect create a new major version of Signal K?

I place great value in nurturing the nascent installation base out there. 

Respecifying stuff in a way that requires a lot of rework or is not backwards compatible should only take place when the current de facto Signal K way of doing things has substantial flaws. There are such, but I don't think _a little open ended with some assumptions made_ justifies making the breaking change you suggest.

tkurki:
>Everything should be made as simple as possible, but not simpler.

As the discovery process based on `/signalk` was created some time after adding the service names there is a bit of overlap: you can discover the ws endpoint by starting from the http service type advertisement. A simplish gateway with no full model, just delta streaming over ws, could advertise http service and then include only the ws endpoint in the `/signalk`response. Ws is built on top of http, so assuming http server capability in addition to ws is a reasonable assumption.

Should we remove the ws service type?

tkurki:
+1 on `swvers` and `swname`.

pod909:
What about services that are provided in the absence of http?

tkurki:
Non http services are "post SK 1.0".  Let's get what we have sorted out first.

SK over TCP would be one.

Also notice that in this context we are talking about mDNS/DNS-SD (a k a Bonjour) discovery in the local network. WAN discovery and non-ip networks are not in the scope.

sumps:
I think having an HTTP and WS service in Discovery is useful if people want to use different ports and we should also have HTTPS and WSS services as well.

gitbook-docs/connection.md
timmathews: Makes sense to me

tkurki:
http and ws service are in `/signalk` response. They can be https and wss there and on the port discovered from mDNS there can be a redirect to https. 

FYI Node server now supports both http/ws and https/wss on the same port.

timmathews:
Let's keep `signalk-ws` and `signalk-http`, but follow the DNS-SD recommendation such that those work for both the plaintext and the encrypted versions (with a redirect if necessary).

That, along with dropping `https` as a service type leaves us with:

```
* `_http._tcp` for the server's web interface
* `_signalk-http._tcp` for the Signal K REST API
* `_signalk-ws._tcp` for the WebSocket data stream
```

I think there is a benefit to having `signalk-ws` and `signalk-http` for the case where they're on different ports (as I believe is true of the Java server, ping @rob42).

The only other (Signal K related) service I can imagine advertising would be Signal K over TCP or UDP. In which case the service name should be just `_signalk._tcp` or `_signalk._udp` respectively. But those - as @tkurki said - are post v1.0.

timmathews:
Just want to clarify that the value in `self` should be in MRN format, so for UUID self values

```
urn:mrn:signalk:uuid:c0d79334-4e25-4245-8892-54e8ccc8021d
```

and MMSI self values

```
urn:mrn:imo:mmsi:123456789
```

PR #302, opened 2016-12-08
Change to Delta Format Example

sumps:
New Delta example taken directly from iKommunicate and also added all of the missing descriptions in the schema group files - except resources group, which I am leaving for @rob42


tkurki:
Merged manually, squashed related commits and fixed formatting.

A lot of your indentations were off like this:
image

Please use 2 spaces as indent.

PR #309, opened 2016-12-19
Improve Top level signalk overview

sumps:
Improve Overview of Signal K in main Readme.md and add typical Signal K system diagram.


rob42:
All good but I think the diagram needs editing to also show native signalk devices and maybe remove RaspberryPi ref.

sumps:
I agree but this diagram is done and I have provided the Dia file so it can be edited and axpanded upon later. It is a good starting point and shows legacy systems, a gateway, a server, a client and cloud service - think it is enough for V1

Issue #311, opened 2016-12-22
Non-leaf updates


jboynes:
What is the expected behaviour for an update for a non-leaf path e.g.
```json
{
  "context": "vessels.urn:mrn:imo:mmsi:234567890",
  "updates": [{
    "values": [{
      "path": "propulsion",
      "value": {
        "0" : {
          "revolutions": { "value": 16.341667},
          "boostPressure": { "value": 45500.000000},
          "tiltTrim": { "value": 48}
        }
      }
    }]  
  }]
}
```

Does this replace the entire `propulsion` object or is the value a sparse model that is merged? If so, what are the rules for merging?

Or, do we only allow updates for leaf nodes? If so, what are those nodes? Should we modify `keyswithmetadata` to remove non-leaf paths?

rob42:
I dont think updating objects was the intent of delta? @tkurki?
If we take the above case to extremes, we would have:
```
{
  "context": "vessels.urn:mrn:imo:mmsi:234567890",
  "updates": [{
    "values": [{
      "path": "navigation",
      "value": {
        ....the whole navigation tree....
      }
    }]  
  }]
}
```
This would then be much better suited to sending the full or sparse format.
But its a good point, we should clarify the intent and rules. As a starter:

delta: `values` will be leaf nodes (edge case, allow arrays) . Complex objects should be decomposed to leaf nodes so we dont have the merge/replace problem. Arrays replace the exiting array.

full: replaces the existing tree.
sparse: merge into exiting tree.

Then we need to identify the sparse/full format somehow I guess?

jboynes:
We should continue to allow structures as leaf nodes e.g. position. Otherwise we introduce synchronization problems between components especially where transports don't ensure order of delivery, or where optional parts may be lost in transit and not detected. The behaviour would be the same as for scalar values with the entire leaf node value being replaced. 

Not allowing `context` or `path` to refer to a arbitrary group makes this much more consistent and, would think, be backward compatible with existing clients if they only deal with leaf nodes. 

We should define what contexts are allowed. Suggestion:
* global
* vessel
* device

sumps:
A Delta is intended to communicate a value change in the most efficient way. Currently the most common cause of a value change is an NMEA sentence or PGN which may have just one value or multiple values.

The current Delta method works very well for this purpose and values in the Delta replace the values in the client/servers name space.

rob42:
Any delta with an object can be decomposed into a delta with a series of values. eg the sample above is equivalent to:
[...]

Issue #304, opened 2016-12-09
Key hierarchy causes confusion


jboynes:
Observing the discussions on electrical, propulsion and now environment, trying to determine an appropriate hierarchy for key names seems fraught and arbitrary. In the recent discussion on temperatures, there are names derived from the N2K spec source type, but also a generic catch-all `environment.temperature.<sourceID>` (which also needs to include an `<instanceID>`)

What value do we get from trying to figure out this hierarchy? Especially when combined with the idea of a manifest as proposed in [RFC0001](https://github.com/SignalK/specification/issues/264).

We could dispense with the hierarchy entirely, and instead:
- include the devices, with source and instance ids in the manifest
- label the device/source/instance with a human-readable name (e.g. "Sea Temp" or "Aft Cabin Temp")
- derive a unique but opaque id for the value e.g. from the device/source/instance components
- use that as the key name in the vessel's context

A display client would define its panel in terms of the opaque ids, can get the display label from the manifest and the value from the vessel context by id.

For example, on a Mareton TMP100 device, channel 2 may be measuring temperature in the fore cabin and channel 3 in the aft cabin. These two sensors may be given ids `tmp100-1234-2` and `tmp100-1234-3` respectively. In the manifest they would be defined as instances 1 and 2 of the `Inside Air` source type and be given the 'Fore' and 'Aft' labels. When a PGN was received, the server would map (tmp100, Inside Air 1) to `tmp100-1234-2` using the manifest and update the `tmp100-1234-2` in the `vessel.{vesselId}` context. A display client using the REST interface would see a reference to `tmp100-1234-2` in the display panel configuration, retrieve the label from the manifest entry for that key in order to display the "Fore Cabin" label and subscribe to the `vessel.{vesselId}.tmp100-1234-2` path to receive the updates for the temperature value.

By flattening the data we avoid all the problems of trying to figure out the right place in the hierarchy for values and make the model extensible for additional data. We need all the plumbing anyway to deal with undefined entries like `environment.temperature.<sourceID>` and to assign human-readable labels to instance ids.

tkurki:
I agree that the hierarchy causes trouble. It is essentially arbitrary grouping for humans. It plays a role in REST api, as you can retrieve a subtree's values by path.

However I think your reasoning is missing one important aspect, that I mentioned in #206: the schema establishes shared semantic meanings to the values. It is not just about that a value is a temperature, but it is a certain temperature that has a meaning.

[...]

Issue #282, opened 2016-10-31
/self should be required


tkurki:
/self should provide a well known path to retrieve the server's self identity. 

Under /vessels/self the id is in a variable path mmsi, uuid or url and not in a standard format, as mmsi is the actual number, not the id.

sumps:
Should we provide some guidance to equipment developers as to which ID should be given highest priority in the situation where more than one ID type is available i.e. MMSI, URL and then UUID ?

jboynes:
The typical way with HTTP like the REST API uses would be to have the server issue a 301 redirect to the resource for the vessel

You can see the `curl` command doing that in this trace:
```
$ curl -vL http://localhost:8080/signalk/v1/api/vessels/self
*   Trying ::1...
* Connected to localhost (::1) port 8080 (#0)
> GET /signalk/v1/api/vessels/self HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.43.0
> Accept: */*
> 
< HTTP/1.1 301 Moved Permanently
< Date: Thu, 15 Dec 2016 04:48:37 GMT
< Content-length: 0
< Location: /signalk/v1/api/vessels/234567890
< 
* Connection #0 to host localhost left intact
* Issue another request to this URL: 'http://localhost:8080/signalk/v1/api/vessels/234567890'
* Found bundle for host localhost: 0x7fa640c10040
* Re-using existing connection! (#0) with host localhost
* Connected to localhost (::1) port 8080 (#0)
> GET /signalk/v1/api/vessels/234567890 HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.43.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Date: Thu, 15 Dec 2016 04:48:37 GMT
< Transfer-encoding: chunked
< 
{...}
```

tkurki:
`/self` is meant as a well defined path to retrieve the self identity, not for accessing the self's data.

jboynes:
Except that a vessel has multiple identities in SK (mmsi, url, uuid) as well other ways to uniquely identify it such as IMO number, flag & documentation number, even port & berth or just position.

It seems to me the main reason for providing this resource in the API is to allow clients to use the API easily - it makes a simple entry point that allows a client to easily determine "the API resource for the vessel associated with this server" which is a very common use-case.

sumps:
[...]

PR #312, opened 2016-12-26
Specify vessel context and leaf path for delta

tkurki:
This change in documentation makes two things explicit:

- `context` is in practice always a vessel path
- `path` is a leaf path

My Signal K related code in lots of places assumes that context is a vessel path to discern  self data from other vessels' data. It would be possible to do just prefix matching, but grouping by vessel is a natural way to handle current Signal K data. One would need to extract the vessel identity anyway in these cases.

Omission of context meaning self context is also well in line with this thinking.

#311 posed the question about should paths be leaf paths. I believe this constraint should be explicit in the documentation. Discerning leaves from the current schema is not always very easy, witness #299, and I think we need to change the schema to make leaves more explicit.

### DIFF
--- a/gitbook-docs/data_model.md
+++ b/gitbook-docs/data_model.md
@@ -163,9 +163,10 @@ A delta message can be recognised from the other types by the topmost level havi
 `updates` is the only required property.
 If `context` is missing it is assumed that the data is related to the `self` context.
 
-Context is a path from the root of the full tree. In this case 'vessels.urn:mrn:imo:mmsi:234567890'. All subsequent data is relative to
-that location. The context could be much more specific, e.g. 'vessels.urn:mrn:imo:mmsi:234567890.navigation', whatever is the common root
-of the updated data.
+Context is a path from the root of the full tree to the _container object_, which is always a vessel with the current Signal K version.
+The delimiter in the context path is `.` (period).
+In this case the context is `vessels.urn:mrn:imo:mmsi:234567890`.
+All subsequent data is relative to that location.
 
 The `updates` holds an array (JSON array) of updates, each of which has a `source` and a JSON array of `values`.
 
@@ -197,7 +198,11 @@ A Signal K producer may not have access to a real time clock or UTC time.
 In these cases timestamp should be omitted.
 Elements in the Signal K processing chain, like a server receiving data from a producer, should fill in timestamp if it is missing in the incoming delta message.
 
-Each 'value' item is then simply a pair of 'relative path', and 'value'.
+Each `value` item is then simply a pair of `path` and `value`.
+The `path` must be a _leaf path_: it must be a path to a leaf the of the full model.
+A leaf is where the actual value of the Signal K property is and where `timestamp`, `$source` and `values` properties are in the full model.
+The value is often a scalar - a numeric value, as in the example above, but it can also be an object.
+For example a `navigation.position` value would be an object like `{"latitude": -41.2936935424, "longitude": 173.2470855712}`.
 
 ## Message Integrity
 
--- a/gitbook-docs/data_model.md
+++ b/gitbook-docs/data_model.md
@@ -201,6 +201,10 @@ The `updates` holds an array (JSON array) of updates, each of which has a `sourc
 An `update` has a single `source` value and it applies to each of the `values` items.
 In cases where you can get data from only a single source the source may be omitted and the receiver may fill it in when multiplexing data from several sources.
 
+A Signal K producer may not have access to a real time clock or UTC time.
+In these cases timestamp should be omitted.
+Elements in the Signal K processing chain, like a server receiving data from a producer, should fill in timestamp if it is missing in the incoming delta message.
+
 Each 'value' item is then simply a pair of 'relative path', and 'value'.
 
 ## Message Integrity
--- a/gitbook-docs/data_model.md
+++ b/gitbook-docs/data_model.md
@@ -7,14 +7,7 @@ format is the same as the full format, but doesn't contain a full tree, just par
 
 ## Full format
 
-The simplest format is the full format, which is the complete Signal K data model  as a JSON string. Abbreviated for
-clarity it looks like this:
-
-```json
-{"vessels":{"9334562":{"navigation":{"courseOverGroundTrue":{"value":11.9600000381},"courseOverGroundMagnetic":{"value":93.0000000000},"more":"a lot more data here...","wind":{"angleApparent":{"value":0.0000000000},"directionTrue": {"value":0.0000000000},"speedApparent":{"value":0.0000000000},"speedTrue": {"value":0.0000000000}}}}}}
-```
-
-Formatted for ease of reading:
+The simplest format is the full format, which is the complete Signal K data model  as a JSON string.
 
 ```json
 {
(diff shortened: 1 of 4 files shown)


fabdrol:
> `context` is in practice always a vessel path

I agree that this is how it's _used_, at the moment, in most code. Does constraining this have any real benefit other than clearing up some confusion that people don't need to have? Allowing more specific context increases flexibility of the deltas.

tkurki:
It makes client and server software easier to implement, which in and of itself is valuable.

fabdrol:
Okay, good point. It's not a massive deal to me so I'm not against this change +1

gitbook-docs/data_model.md
```diff
@@ -171,9 +171,10 @@ A delta message can be recognised from the other types by the topmost level havi
 `updates` is the only required property.
 If `context` is missing it is assumed that the data is related to the `self` context.
 
-Context is a path from the root of the full tree. In this case 'vessels.urn:mrn:imo:mmsi:234567890'. All subsequent data is relative to
-that location. The context could be much more specific, e.g. 'vessels.urn:mrn:imo:mmsi:234567890.navigation', whatever is the common root
-of the updated data.
+Context is a path from the root of the full tree to the _container object_, which is always a vessel with the current Signal K version.
```
jboynes: This is true for current implementations, but the current message format allows for updates to other contexts such as `resources` e.g. to add/remove waypoints. Line 172 also describes the `self` context and if we continue to allow that (but see #282) we should allow that to be explicitly set as well.

gitbook-docs/data_model.md
```diff
@@ -201,7 +202,11 @@ The `updates` holds an array (JSON array) of updates, each of which has a `sourc
 An `update` has a single `source` value and it applies to each of the `values` items.
 In cases where you can get data from only a single source the source may be omitted and the receiver may fill it in when multiplexing data from several sources.
 
-Each 'value' item is then simply a pair of 'relative path', and 'value'.
+Each `value` item is then simply a pair of `path` and `value`.
+The `path` must be a _leaf path_: it must be a path to a leaf the of the full model.
+A leaf is where the actual value of the Signal K property is and where `timestamp`, `$source` and `values` properties are in the full model.
+The value is often a scalar - a numeric value, as in the example above, but it can also be an object.
```
jboynes: In the full model a leaf node is _always_ an object - it needs to be to store the `timestamp` and other meta information about the value. The scalar quantity is stored in a `value` property of that object. However, for `object` values we [mix](https://github.com/SignalK/specification/blob/master/schemas/groups/navigation.json#L396) the commonValueFields into the leaf object which makes this inconsistent.

Perhaps the solution for that is to modify the full model to always use the `value` property for storing the value and stop mixing in.

gitbook-docs/data_model.md
```diff
@@ -171,9 +171,10 @@ A delta message can be recognised from the other types by the topmost level havi
 `updates` is the only required property.
 If `context` is missing it is assumed that the data is related to the `self` context.
 
-Context is a path from the root of the full tree. In this case 'vessels.urn:mrn:imo:mmsi:234567890'. All subsequent data is relative to
-that location. The context could be much more specific, e.g. 'vessels.urn:mrn:imo:mmsi:234567890.navigation', whatever is the common root
-of the updated data.
+Context is a path from the root of the full tree to the _container object_, which is always a vessel with the current Signal K version.
+The delimiter in the context path is `.` (period).
```
jboynes: So something that has been bugging me: why do we use `.` for the delimiter here but use the path delimiter `/` in the REST API? Can we be consistent?

Actually, this does not actually seem like a true path. When we say context "is always a vessel" then we are really specifying a context type (vessel) and identifier for a specific context (vesselId). Especially if we move away from the hierarchical model for values (#304) then it would be more flexible and easier to implement if we made `context` an object e.g.
```json
{
  "context": {
    "type": "vessel",
    "id": "urn:mrn:imo:mmsi:234567890"
  },
  "updates":[]
}
```
That means applications no longer need to parse the string to separate the two, and avoids any issues with needing to escape a `.` or `/` separator.

tkurki: But we are consistent: paths in property names and values like in full tree `$source` and `values` property keys use period delimiter. If the `values` property keys used / as delimiter and you wanted to retrieve a certain value via REST / would need escaping.

Not saying everything is perfect, but change to / would not be backwards compatible, as would the structured context you suggest.

gitbook-docs/data_model.md
tkurki: I'll create a separate issue about pushing object values under `value` in full, but this part of the doc is about value in delta, not in full.

gitbook-docs/data_model.md
jboynes: In the REST API the escaping mechanism is defined by the URI specification so that is already covered. We do not have an equivalent for our `.` delimited paths.

This change now says context is always a vessel whereas before it allowed any node in the model. That is not a compatible change.

It is introducing _container object_ and _leaf path_ as special concepts in the model. This is good in that it is reflecting reality better than the old hierarchy but is a fundamental change. Trying to force backward compatibility with the existing implementation is causing oddities like this one where applications have to parse `context` as a string to extract a context/id pair where there are simpler and more representative ways to convey that.

gitbook-docs/data_model.md
jboynes: The issue is with consistency between the two views of the model. For a given value, both the path used to identify it and its type should be the same in both views.

For example, to read a vessel's heading using the REST API, an application needs to GET from `{api}/{context}/{path}/value` where context is `vessels/{vesselId}` and path is `navigation/heading` whereas updates are sent as a delta where context is `vessels.{vesselId}` and path is `navigation.heading`.

A leaf value can also store `meta` information. If the `value` is the same as the full model (i.e. an `object`) then changes to that (e.g. changes to alarm zones) can be included in a delta. If it is just a scalar then it cannot and a separate mechanism will be needed.

jboynes:
By defining vessel as the only context, this change removes the ability to send a delta with updates affecting multiple vessels. For example, a race committee or fleet operator can no longer send an update with the current position/speed/heading of all vessels in the race/fleet.

fabdrol:
Very good point, didn't think about that. Especially for mobile clients a larger JSON delta is preferable over many smaller JSON deltas needing to be pushed imo.

sumps:
Am I missing something or can this "niche" use case not be satisfied with our current subscription option ?

The reason why we use / in the restAPIs URLs and . in the Deltas is that / cannot be used in JSON without storing it as \/ which would then need additional parsing. No developers have queried or not understood this inconsistency.

jboynes:
I created #314 as a proposal for how to address this.

tkurki:
>this change removes the ability to send a delta with updates affecting multiple vessels

That was never the intention afaik. See for example https://groups.google.com/forum/#!searchin/signalk/delta$20reads%7Csort:relevance/signalk/XFnlGYweN9Q/TvaxtPoLBwAJ
> Delta message reads logically "here are some updates about context X, here's stuff from source Y, here's the timestamp and here are the data values as key-value pairs"

Your fleet use case is covered by sending a delta per vessel, which is to me a lot more intuitive than fashioning a single delta.

PR #313, opened 2016-12-26
Itemize course properties

tkurki:
This PR "itemizes" course properties: each is pushed under `value` property and accompanied with timestamp, source and values structure.

This makes possible to update them individually and retain source and timestamp information. 

More importantly this is in line with using just leaf paths in delta.

The possible backwards compatibility issue with this is that if a source has produced object valued deltas it should now send scalar-valued deltas, with the properties in individual pathvalues.


tkurki:
Related to #311, see https://github.com/SignalK/specification/issues/311#issuecomment-269059616.

tkurki:
Solves the problems with https://github.com/SignalK/n2k-signalk/pull/45 and https://github.com/SignalK/n2k-signalk/pull/29.

Makes #295 obsolete.

Issue #155, opened 2016-01-23
Identity in Delta Messages


tkurki:
When we used raw mmsi and self-assigned alphanumeric ids as keys in the `vessels` array the `context` in delta was in a format like `vessels.12345679`, with period as the delimiter, denoting path in the full tree.

Now with more structured identities (uuid, mmsi, url) period  as the delimiter is not sufficient, as urls typically contain periods.

I suggest the following
- The _branch_(the main branch from the root) and the _identity_ of the vessel/other entity are explicitly in separate fields as below.
```
{
  "branch": "vessels",
  "id": "urn:mrn:imo:mmsi:230099999",
  "updates": [...]
}
```
- All `self` delta messages *must* omit the context, making it possible to distinguish `self`messages by the absence of identity. Saves a few bytes also.
- For non-self messages `vessels` is the default and *should* be omitted

This should clear up the delimiter issue, save a few bytes and make things explicit. All paths in the delta message are rooted at the two part path formed by branch and identity from the root.

Comments?

rob42:
The example above solves the problem of dots as delimiter but I dont like dropping the `self`. It makes `self` a special case rather than an alias as it is now.  

I also think we should use the existing `context` rather than `id`, as the change in language may confuse adopters, and 'id' is not specific to the vessels uuid as context has become.  Thus we get minimal change, but the same benefits.

We can just add `branch`. So the example above becomes:
```
{
  "branch": "vessels",
  "context": "urn:mrn:imo:mmsi:230099999",
  "updates": [...]
}
```
In some cases the branch is implicit in the context, eg `urn:mrn:imo:mmsi:230099999` must be a vessel (ignoring shore stations for now), but thats not always going to work out, and takes resources to check, so always including `branch` is a good idea IMHO

PR #314, opened 2016-12-31
Clarify context and allow for update-specific contexts

jboynes:
Allow `context` to be supplied as a object rather than a `.` delimited string (while still supporting the older format for backwards compatibility).

Allow `context` to be specified for each individual update to allow deltas to span multiple contexts (e.g. updating all vessels in a race or commercial fleet).

Proposes an alternative representation for an update as a simple object rather than path/value pairs. The two forms can be distinguished by the type of the `values` property making this a backward compatible, additive change.

### DIFF
--- a/gitbook-docs/data_model.md
+++ b/gitbook-docs/data_model.md
@@ -160,23 +160,33 @@ In more detail we have the header section:
 
 ```json
 {
-  "context": "vessels.urn:mrn:imo:mmsi:234567890",
+  "context": {
+    "scope": "vessel",
+    "id":"vessels.urn:mrn:imo:mmsi:234567890"
+  },
   "updates": [
     ...data goes here...
   ]
 }
 ```
 
-A delta message can be recognised from the other types by the topmost level having `updates` property.
-`updates` is the only required property.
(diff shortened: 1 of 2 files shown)


rob42:
way back when @tkurki proposed the delta format I assumed the context would be the longest common root of the values sent, since thats the most efficient format, eg why duplicate `navigation.`for every nav value, better to have the context `vessels.urn:xxxxxxxx.navigation`. But it causes complexity in parsing, and we decided simplicity was more valuable. So we always use `vessels.urn:xxxxxx`. We have discussed dropping the `vessels.` part as thats always the same.

Making context into an object, with attributes `vessel` is a good solution, but breaks the current format. By also including a `path` attribute we could get some efficiency gains by not duplicating common paths.

But the use cases can be fulfilled with the current format, without major issues and we are very close to V1 release.

I favour this change, but we need to include it in a later version as its not critical to V1.

tkurki:
See also #155

tkurki:
This kind of change needs also software support in the reference servers.

rob42:
Closing as no comments in 18months. Pls reopen if still valid

Issue #275, opened 2016-10-19
positionFixTimestamp (with miliseconds) is extremely important for racing apps


sailracerdotnet:
Problem: to calculate COG and SOC (other than provided by GPS) using different averaging methods in for flat or for choppy conditions it is a MUST to have proper time stamp of a coordinate fix. It is also a MUST calculating the time of crossing start/finish line or taking a mark or gates on backend. The updates/timestamp is not an option because it is local time of SignalK server and might contain errors in system tyme setup.

Possible solution - to add UTC time stamp with **miliseconds**  (provided by gps receiver) to:
option 1) `updates/values/*[path="navigation.position"]/value` `{latitude, longitude, timestamp}`
option 2) in `updates/values` add another node  `*[path="navigation.positionFixTimestamp"]/value="2016-01-01T00:00:00.000Z"`

tkurki:
>The updates/timestamp is not an option because it is local time of SignalK server 
The delta schema doesn't specify this to be the local time of Signal K server. Furthermore it is specified as UTC, so it is definitely not _local_. Is this mentioned someplace else in the documentation?

When you brought the issue up in Slack I mentioned that the current timestamp is probably not well enough defined for your purpose. That we can fix I think, by stating that

```The timestamp is the best available information about the time of origin of the related value.````

For example in your case of position the data may be coming from NMEA 0183 GGA message that carries the UTC time of the fix. This should be used as the timestamp when converting to Signal K. For example [nmea0183-signalk](https://github.com/SignalK/signalk-parser-nmea0183/blob/master/codecs/GGA.js) already does this, using system time to fill in date part and using GGA data for time.

I am not sure about N2K data - is there a source for accurate time of a position fix?

Anything downstream in the SK processing chain should not meddle with the timestamp if there is one in the input SK it is processing.

That still leaves the millisecond part of your requirement. I suggest we explicitly allow it, by specifying a regular expression for timestamp with millisecond precision at https://github.com/SignalK/specification/blob/90f7d9b08c7add281dab9b74055780e4a3ef3bb2/schemas/definitions.json#L8-L13 and add that to the spec documentation.

What do you think?

bkp7:
This seems like a good solution to me. The delta currently only specifies (optionally) timestamp at the high level, ie. there is not a timestamp for each 'path', 'value' pair. I can't see a reason why this should not be added.

[...]

Issue #41, opened 2014-12-11
Simplify navigation.current


timmathews:
`navigation.current` includes source and timestamp fields for each component (`drift`, `setTrue` and `setMagnetic`), but since these values are only meaningful together and always updated together, timestamp and source should be removed from each component and added to `current` directly

``` json
"current": {
  "timestamp": {"$ref": "../defintions.json#/definitions/timestamp"},
  "source": {"$ref": "../definitions.json#/definitions/source"},
  "drift": {
    "type": "number",
    "description": "The speed component of the water current vector in m/s",
    "example": 3.12
  },
  "setTrue": {
    "type": "number",
    "description": "The direction component of the water current vector in decimal degrees referenced to true (geographic) north",
    "example": 123.45
  },
  "setMagnetic": {
    "type": "number",
    "description": "The direction component of the water current vector in decimal degrees referenced to magnetic north",
    "example": 131.22
  }
}
```

PR #315, opened 2017-01-04
Object valued properties under value in full model

tkurki:
This PR is work in progress to treat object valued properties the same as scalar values: the object under `value` property, with `timestamp`, `$source` and `values` siblings.

TODO:
- [x] position first pass
- [x] values handling
- [x] Check other object valued properties
- [x] Generate proper keys documentation and keyswithmetadata.json

### DIFF
--- a/dist/delta.js
+++ b/dist/delta.js
@@ -1,0 +1,69 @@
+'use strict';
+
+var _typeof = typeof Symbol === "function" && typeof Symbol.iterator === "symbol" ? function (obj) { return typeof obj; } : function (obj) { return obj && typeof Symbol === "function" && obj.constructor === Symbol && obj !== Symbol.prototype ? "symbol" : typeof obj; };
+
+var _ = require("lodash");
+
+/*
+ * Credit for these function goes to @tkurki
+ */
+
+var addToTree = function addToTree(pathValue, source, tree) {
+  var result = {};
+  var temp = tree;
+  var parts = msg.path.split('.');
+  for (var i = 0; i < parts.length - 1; i++) {
+    temp[parts[i]] = {};
+    temp = temp[parts[i]];
+  }
+  temp[parts[parts.length - 1]] = msg;
+  return result;
+};
+
+function addAsNested(pathValue, source, timestamp, result) {
+  var temp = result;
+  var parts = pathValue.path.split('.');
+  for (var i = 0; i < parts.length - 1; i++) {
+    if (typeof temp[parts[i]] === 'undefined') {
+      temp[parts[i]] = {};
+    }
+    temp = temp[parts[i]];
+  };
+
+  //mapping produced an object like {latitude:...,longitude:...}
+  if (_typeof(pathValue.value) === 'object') {
+    temp[parts[parts.length - 1]] = pathValue.value;
+    temp[parts[parts.length - 1]].source = source;
+    temp[parts[parts.length - 1]].timestamp = timestamp + '';
+  } else {
+    temp[parts[parts.length - 1]] = {
+      value: pathValue.value,
+      source: _.clone(source),
+      timestamp: timestamp + ''
+    };
+    delete temp[parts[parts.length - 1]].source.timestamp;
+  }
+}
+
+function valuesToSubTree(delta) {
+  var valueTree = {};
+  var timestamp = delta.updates[0].source.timestamp;
+  delta.updates[0].values.forEach(function (pathValue) {
+    addAsNested(pathValue, delta.updates[0].source, timestamp, valueTree);
+  });
+  return valueTree;
+}
+
+function deltaToNested(delta) {
+  var result = {};
+  var contextPointer = result;
+  var pathPropertyNames = delta.context.split('.');
+  for (var i = 0; i < pathPropertyNames.length - 1; i++) {
+    contextPointer[pathPropertyNames[i]] = {};
+    contextPointer = contextPointer[pathPropertyNames[i]];
+  };
+  contextPointer[pathPropertyNames[pathPropertyNames.length - 1]] = valuesToSubTree(delta);
+  return result;
+}
+
+module.exports.deltaToNested = deltaToNested;
--- a/dist/fullsignalk.js
+++ b/dist/fullsignalk.js
@@ -1,0 +1,256 @@
+'use strict';
+
+/*
+ * Copyright 2016, Teppo Kurki
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+var _ = require('lodash');
+var signalkSchema = require('./');
+var getId;
+var debug = require('debug')('signalk:fullsignalk');
+
+function FullSignalK(id, type, defaults) {
+  //hack, apparently not available initially, so need to set lazily
+  getId = signalkSchema.getSourceId;
+
+  this.root = {
+    vessels: {},
+    self: id,
+    version: "0.1.0" // Should we read this from the package.json file?
+  };
+  if (id) {
+    this.root.vessels[id] = defaults && defaults.vessels && defaults.vessels.self ? defaults.vessels.self : {};
+    this.self = this.root.vessels[id];
+    signalkSchema.fillIdentity(this.root);
+  }
+  this.sources = {};
+  this.root.sources = this.sources;
+  this.lastModifieds = {};
+}
+
+require("util").inherits(FullSignalK, require("events").EventEmitter);
+
+FullSignalK.prototype.retrieve = function () {
+  return this.root;
+};
+
+FullSignalK.prototype.addDelta = function (delta) {
+  this.emit('delta', delta);
+  var context = findContext(this.root, delta.context);
+  this.addUpdates(context, delta.context, delta.updates);
+  this.updateLastModified(delta.context);
+};
+
+FullSignalK.prototype.updateLastModified = function (contextKey) {
+  this.lastModifieds[contextKey] = new Date().getTime();
+};
+
+FullSignalK.prototype.pruneContexts = function (seconds) {
+  var threshold = new Date().getTime() - seconds * 1000;
+  for (var contextKey in this.lastModifieds) {
+    if (this.lastModifieds[contextKey] < threshold) {
+      this.deleteContext(contextKey);
+      delete this.lastModifieds[contextKey];
+    }
+  }
+};
+
+FullSignalK.prototype.deleteContext = function (contextKey) {
+  debug("Deleting context " + contextKey);
+  var pathParts = contextKey.split('.');
+  if (pathParts.length === 2) {
+    delete this.root[pathParts[0]][pathParts[1]];
+  }
+};
+
+function findContext(root, contextPath) {
+  var context = _.get(root, contextPath);
+  if (!context) {
+    context = {};
+    _.set(root, contextPath, context);
+  }
+  var identity = contextPath.split('.')[1];
+  if (!identity) {
+    return undefined;
+  }
+  signalkSchema.fillIdentityField(context, identity);
+  return context;
+}
+
+FullSignalK.prototype.addUpdates = function (context, contextPath, updates) {
+  var len = updates.length;
+  for (var i = 0; i < len; ++i) {
+    this.addUpdate(context, contextPath, updates[i]);
+  }
+};
+
+FullSignalK.prototype.addUpdate = function (context, contextPath, update) {
+  if (typeof update.source != 'undefined') {
+    this.updateSource(context, update.source, update.timestamp);
+  } else if (typeof update['$source'] != 'undefined') {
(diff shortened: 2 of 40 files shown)


tkurki:
First problem caught by tests: how do you pass the name of an AIS target, since `name` is vessel object's  property, without `value`.

rob42:
First, we should define a list of reserved words for keys like `value, timestamp, $source, source,` etc. That will avoid awkward objects in future. 

In the case above I think we need to be explicit about the difference between branch and leaf, and the ability of a branch to have arbitrary properties, whereas a leaf can have ONLY `value, timestamp, etc`.

Otherwise it seems to me we will soon have a very complex recursive `value.object.path.value.object.value` chain.

tkurki:
@rob42 I didn't quite get your comment _In the case above_.  

So what to do about the `name` in incoming AIS targets? Do not send deltas about them at all, just in the full model? Node server uses deltas for all communication from the producers, so some solution needs to be there for the names to end up in the full model.

The reserved property names should be documented, but the issue is not very pressing, because schema changes should go via PRs anyway (unless we fumble).

rob42:
Oh, I see your problem! 

This is the code Ive used to decode deltas sent to the browser
```javascript
 ais.values.forEach(function(value) {
	//console.log(value);
	if(value.path === 'navigation.position.latitude'){
		//console.log(value.value);
		lat=value.value;
	}
	if(value.path === 'navigation.position.longitude'){
		//console.log(value.value);
		lon=value.value;
	}
	if (value.path === 'navigation.position') {
                   //console.log('Pos:'+value.value.latitude+':'+value.value.longitude);
                    lat = value.value.latitude;
                    lon = value.value.longitude;
            }
	if(value.path === 'navigation.courseOverGroundTrue'){
		cog=value.value;
		radians = value.value-(Math.PI*.5);//util.toRad(value.value-90);
	}
	if(value.path === 'navigation.speedOverGround'){
		sog = value.value;
	}
	if(value.path === 'mmsi'){
		mmsi = value.value;
	}
	if(value.path === 'flag'){
		flag = value.value;
	}
	if(value.path === 'port'){
		port = value.value;
	}
	if(value.path === 'name'){
		name = value.value;
	}
	if(value.path === 'communication.callsignVhf'){
		vhf = value.value;
	}
	if(value.path === 'navigation.state'){
		state = value.value;
	}
});
```
You can see I send `name` as a `path`, and `value` as the value.
Seems to work.
Rob

tkurki:
`name` is just an example - the same goes for for example `mmsi` and any other simple property, that does not require the value/values/$source/timestamp handling. 

Either we start sprinkling schema awareness to the part of app that creates the full model based on incoming deltas or figure out a way to specify such property updates in delta so that they can be handled generically.

I think making delta => full mapping schema aware is a bad idea in several ways. It does not allow one to use emerging properties, that have not yet made it to the schema, and makes the implementation a lot more complex.

Mulling this over I have come to like the idea of extending the `delta` format to cover this case.
For example like this:
```
{
  "updates": [{
    "context": "vessels.urn:mrn:imo:mmsi:236333000",
    "source": {...},
    "timestamp": "2014-08-15T19:06:44.137",
    "properties": [{
      "path": "name",
      "value": "HOOGE"
    },{
      "path": "mmsi",
      "value": "236333000"
    }]
  }]
}
```

This would clearly distinguish the simple properties from the more elaborately handled values. It would also be pretty backwards compatible, as we are adding a property to the delta's updates property.

Thoughts?

rob42:
Hmm, works - but are there security implications..I guess no worse than now as we dont record the source/timestamp for these properties anyway. And we do have the source and timestamp of the delta.

At a meta level we have created two types of leaf (value), those that have source/timestamp, and those that dont... maybe we need to fix that instead?

tkurki:
Yes on two leaf types - except that they've been there all along. 

What did you mean by _fix that instead_?

rob42:
maybe we should convert all the 'properties' to  values

That would simplify the processing and mean all leaf values are the same

rob42:
Just had a quick look, there are a lot of places where 'properties ' are used, See resources, json
Not practical or useful to convert them all.

joabakk:
I tend to agree with @rob42 on this one. Having tried to process environment.current in the same way as other paths, it does not lend itself as easily to automation, ref instrumentpanel etc. There may be workarounds I'm not familiar with, but it is easier to look for the value property

tkurki:
@joabakk are you talking about using environment.current streamed as delta or via rest as full?

joabakk:
Not really sure what I'm doing @tkurki ... For most measured values, I can get the value by adding .value to the path for say plugin `_.get(app.signalk.self, path)`, but not for the special cases like environment.current

bkp7:
Has a resolution for the above been agreed?

Looks to me like 2 issues in the above:

@joabakk raised the example of environment.current. It seems to me this just needs to be corrected in the schema. IMHO these values should have the full complex value, see in the example below the difference between current and wind we have at the moment which makes no sense to me. Why should wind.directionTrue have meta data and current.setTrue not? To my mind the meta data on these values is really important to allow the gauges and displays to work automatically.

 ```json
{
  "vessels": {
    "urn:mrn:imo:mmsi:230099999": {
      "mmsi": "230099999",
      "environment": {
        "current": {
          "drift": 3.12,
          "setTrue": 1.5,
          "timestamp": "2014-08-15T19:00:15.402Z",
          "$source": "foo.bar"
        },
        "wind": {
          "directionTrue": {
            "value": 1.8,
            "meta": {},
            "timestamp": "2014-08-15T19:00:15.402Z",
            "$source": "foo.bar"
          },
          "speedOverGround":{
            "value": 5.24,
            "meta": {},
            "timestamp": "2014-08-15T19:00:15.402Z",
            "$source": "foo.bar"
          }
        }
      }
    }
  },
  "version": "1.0.0"
}
```

I propose a new issue to go through the schema and check for/correct any cases like current. 

The second issue is the many other properties as identified by @rob42 and @tkurki which should not have, and don't need the full complex value. @tkurki has proposed 1 solution which looks like it should work. I can think of an alternative which is to explicitly include the .value for the complex cases:
```json
{
  "context": "vessels.urn:mrn:imo:mmsi:236333000",
    "source": {...},
    "timestamp": "2014-08-15T19:06:44.137",
    "updates": [
    {
      "values": [
        {
          "path": "mmsi",
          "value": "212345678"
        },
        {
          "path": "environment.wind.directionTrue.value",
          "value": 5.5
        }
      ]
    }
  ]
}
```
Not backwardly compatible, but has the advantage of additional flexibility because anything can be updated with the delta message. eg. changing the red warning zone on a tacho when engine goes into limp home mode.

Just some thoughts to try and move this along....

tkurki:
Current is actually like that per design, see #41. If we go through with this PR current properties should be pushed one level down, under `value`.

So whenever a property is fundamentally composite or it doesn't make sense in this day and age to handle the parts discretely it is object valued. 

Whether or not something is composite property or just a couple of discrete properties is another discussion - this PR deals with the mechanisms to represent them.

tkurki:
Thanks for the fresh thinking on appending `.value` to most paths. 

I am not overly fond of it, because it makes the most prevalent cases more verbose.

bkp7:
OK. Thanks for pointing out #41, I had not read that and it makes sense of the rationale behind other values which are not composite (Attitude, Keel, ...). IMHO that logic is flawed though, with gauge info now being stored in the metadata with each value. I have opened issue #329 to separate it from this one.

bkp7:
I tend to agree with your verbose point, however against that, having 2 separate sections 'properties' and 'values' adds some complication. In reality I think either approach will work, and presumably if your properties approach is taken any value could be set in there, including a composite value?

If so, you could argue that your properties section is essentially the same as what I proposed, and the values section offers a less verbose way to send the more frequently updated values that most instruments will be interested in.

Also, regarding the verbose point, in many cases there will be a lot more verbage caused by the issue I raised in #275. We are currently developing a native Signal K sensor suite, and due to the design it is not practical to read all the sensors at exactly the same time so our system will be producing verbose deltas in the format you suggested in #275 anyway.

Having said all the above, what seems to me to be important is to make a decision on this.

tkurki:
Yes on allowing composite values for properties.

About timestamps and verbosity: would you not want to send the data as soon as you read it from the sensor, so you would be sending out deltas for every reading anyway.

bkp7:
Some sensors are being read up to several hundred times per second and digitally smoothed to reduce noise. Others have a slower read rate so are read periodically 1-5 times per second say. This is done on a round robin basis so not all sensors are read at the same time.

Since there can be many subscribers to any given sensor, and each can have requested different update rates there is no link between the actual sensor reading and sending of the data. So the methodology is that when the time comes to send more data to a particular subscriber, the latest available data for each sensor is collected up and sent to the websockets queue.

We have also found that each websocket push has an overhead and if we send each measurement separately we have to deal with complex caching and queue management at the websockets layer.

So we have taken the view that if a subscriber asks for several items on the same refresh period we supply them all together in one delta message.

We have considered rounding the times against each sensor reading eg. to the nearest 0.2s and then treating that as the timestamp for all the sensor readings. The only problem with that is if the upstream consumer is trying to use the data other than for display. eg. to calculate rate of change from the raw data, this will then make any such derivation less accurate because the timings are off.

rob42:
This is drifting into the too hard list:-)
The language below is specific, `value` is the signalk sibling for `timestamp`, value is the stored data. Hopefully avoids the ambiguity of talking about 3.451 being the the value of `value`....
 
Just trying to think laterally and back to the core issue, we have 4 types of path:
1) a path that has at least one child path that is an object. example = `..navigation`
2) a path with a child `value`, or `values`. Value will be a primitive, or simple array. example: `..navigation.courseOverGroundTrue`
3) A path not ending in `value`, no children, and with a primitive value. Example `mmsi`, @tkurki 's _properties_ above
4) A path with children, including  `source` and/or `timestamp` Example `..navigation.position.* This is a composite object.

So in delta format:
1) and 2) above are the usual case in signalk. 
4) Send a delta with `source` and/or `timestamp` where `value` is an object.

3) Is harder to identify in a delta. Maybe send a separate delta, in `updates`  send `values` without `source` and `timestamp`. Or we need a sibling to `updates`, but not `properties` as that will be jsoschema hell.

rob42:
Case 2 above, easy to recognise composite values in delta:
```
{
    "context": "vessels.urn:mrn:imo:mmsi:234567890",
    "updates": [{
        "source": {
           ....
        },
        "timestamp": "2010-01-07T07:18:44Z",
        "values": [{
            "path": "navigation.position",
            "value": {
								"longitude": 173.1693,
								"latitude": -41.156426,
								"altitude": 0
								}
        }]
    }]
}
```

rob42:
Case 3 above works well in delta like this:
```
{
    "context": "vessels.urn:mrn:imo:mmsi:234567890",
    "updates": [{
        "values": [{
            "attribute": "mmsi",
            "value": "234567890"
        }]
    }]
}
```
In the full tree it can be recognised as per my rule 3 above. 

Does that solve the issue in all scenarios? Except for the addition of `attribute` to the delta its backwards compatible too.

tkurki:
Good analysis and solution, @rob42. I like the direction this is going. I will revisit the implementation.

tkurki:
Does anybody have a good idea on how to find any object valued properties we are using? Can't recall anything but `position` and `current`.

joabakk:
There is closestApproach from #366 if that is ok

tkurki:
Hmmm, I came up with a way to shortcut the delta-to-properties-with-no-metadata problem with the following rule: _if a delta pathvalue has empty string as the path then merge the value's structure into the full model (without source & timestamp & value/values handling)_. This allows a delta like 

```
{
  "context": "vessels.urn:mrn:imo:mmsi:276810000",
  "updates": [
    {
      "source": {...},
      "timestamp": "2014-08-15T19:02:31.507Z",
      "values": [
        {
          "path": "",
          "value": {
            "name": "WRANGO"
          }
        }
      ]
    }
  ]
}
```
to set `name` property under a vessel's root. 

Thoughts? To me this looks like a very tidy solution to _"properties"_.

tkurki:
Finally, I'm ready to merge this, bringing finally balance to the full model!

tkurki:
- [ ] attitude

tkurki:
This affects notifications as well: they are object valued and must go under `value` as well, if we are to be consistent.

- [ ] notifications

tkurki:
Notifications done, (again) not aware of any more work to be done. N2k-signalk changes done. Nmea0183 parser new version did not have any validity tests, added some to obvious object valued properties and fixed any issues there.

tkurki:
@sbender9 continue here, not really n2k-signalk specific
>Now we also don't get meta for things like design.beam. Not sure if that is handled by another change? Or is even possible now?

With the current master:
http://node-master.signalk.org/signalk/v1/api/vessels/self/design/beam/meta
```
{
units: "m",
description: "Beam length"
}
```
http://node-master.signalk.org/signalk/v1/api/vessels/self/design/beam
`Cannot GET /signalk/v1/api/vessels/self/design/beam`

So we do have meta for units & description available with direct HTTP call, even if there is no value there.

`design.beam` is actually a normal value with the metadata and `value/values`.

Digging into the schema a bit deeper still: design has some object valued fields, that should be pushed under `value` and some that are normal `numberValue`s.

- [ ] design.length
- [ ] design.keel

sbender9:
I don't think any of those design fields should be normal numberValues.  They can actually come from multiple sources.

tkurki:
Can or can not?

If multiple sources then they need to be numberValues.

sbender9:
Sorry, brain fart. They CAN have multiple sources. And SHOULD be numberValues. `design.beam` for example.

tkurki:
- [ ] undo dist addition before merging

tkurki:
The design section is not quite so clear as to how ti should be fixed. Merging object valued properties, as there is considerable value in harmonising this for for example position in the full model.

Issue #246, opened 2016-08-31
Switch to better validation tool


tkurki:
tv4 appears dead as a project and its error reporting is sometimes really confusing.

Consider switching to another JSON Schema validator, such as http://epoberezkin.github.io/ajv/

See also https://github.com/ebdrup/json-schema-benchmark

bkp7:
@tkurki, just wanted to check we agree on the objectives before I spend too much time.

The end goal is to change from tv4 to ajv (which still looks to be active and good) and to automatically test schemas, samples, tests and the examples to be put in the documentation.

I would also like to make it easier to add and remove tests, which for me would just mean adding or removing a json file and the tests then automatically running. The present arrangement where each test has to be added in js is not straightforward at all. I propose the tests are placed in a folder structure:
- tests
  - full
    - valid
    - invalid
  - delta
    - valid
    - invalid
  - electrical
    - valid
    - invalid
  - sails
    - valid
    - invalid
  - etc ....

I think for unit testing it makes sense to separate the lower level schemas as there ought to be quite a few tests for each. For samples and documentation examples I think they should all be top level schema compliant (ie full, delta, subscribe, unsubscribe, discovery). Do you concur?

It would also be great to build something for the documentation examples so that sections can be minimised automatically (the scaffolding you mention in #320).

At present the tv4 tests have the banUnknownProperties flag set, which along with other issues in tv4 makes reliable testing and validation difficult. With this set, the current 'loose' json schema we have appear to be 'hardened', when other validators do not see it that way. I can't see that ajv has an equivalent flag. This will mean that changing to ajv will probably require some changes to the schemas to achieve the same results.

Keeping eyes on the end goal I propose:
1. Move/modify existing tests into the new folder layout but still running with tv4 to ensure we don't lose any existing test results
2. migrate to ajv
3. add in handling of documentation examples

Finally, a nice to have, would be a SignalK web page validator into which json can be copied and it validates it. I believe ajv can do this for us.

Your thoughts?

tkurki:
Please see http://node-master.signalk.org/@signalk/playground/dist/ fir an interactive validation tool. Where would you have found thus link? I mean how ti tell people about it?

To be useful the schema validation must be run so that the data only properties present in the schema. If ajv can not do that I don’t quite see how it would be useful.

This issue is about the validation tool. Reorganising tests is another - but I do agree with you on running all json files against the schema automatically. If you start this way please PR early, let’s review it early.

[...]

PR #320, opened 2017-01-22
Fix schema files to be valid against http://json-schema.org/draft-04/schema#

bkp7:
Fixed Schemas so that they validate. It seems the schemas themselves are not included in the tests.

Also I discovered that the uuid could be set to anything at all because the additionalProperties had not been set to false, so I've fixed that too.


tkurki:
Great stuff! I am all for merging this, just one question: did you manually validate all these? It would be great to add automatic tests so that everything does not degrade over time as things evolve. The lack of runnable tests is imho the root cause for these issues.

If you did this manually it would be great to share command line instructions in the readme.

bkp7:
They were validated using Altova. 

I will be working through a number of other issues we have found with the schemas over the next couple of days (on different branches), but I suggest you leave this here and I will aim to add the tests in.. 

Any objection to adding in tests for the samples so they stay relevant?

Also, is there any reason the tests are not all valid SignalK json? Some are snipets. I would have found it a lot easier if all the tests validated against one of the top level schemas which I think are: signalk.json, delta.json, subscribe.json, unsubscribe.json and discovery.json. If that's correct I think it should be in the documentation.

tkurki:
See also
https://github.com/SignalK/specification/tree/ajv-validator
https://github.com/SignalK/specification/issues/246

It would be absolutely wonderful to add validation of the samples to the tests.

The samples and tests have grown over time and are frankly a mess. You may consider also removing some of the samples or replacing them with better ones.

Another point to consider is that the samples used in the documentation under `gitbook-docs` would be separate files, included in the documentation, so that we can have validation tests for them as well. Even if what's needed in the documentation is a snippet we can add a test that validates it with some scaffolding.

Are you sure you want to keep this open? I would rather merge this and continue off master.

bkp7:
OK, merge this and I'll look at adding tests for the schemas later

PR #321, opened 2017-01-23
Fix notifications schema

bkp7:
The notification/alarms branches were not validating correctly because any object was allowable. Thus incorrectly formatted alarms would still validate against the schema. Several additional tests added. When running all the tests against unmodified schema these ones failed:

- alarms-invalid3.json must not have sub trees beneath well-defined notifications
- alarms-invalid6.json must not have incorrectly spelled keys
- alarms-invalid7.json must not use well-defined notifications at deep levels
- alarms-invalid8.json must not have additional keys in notifications

To fix the problem the  notifications.json schema has been reworked to define 2 objects: 

- a notification object which is the alert notification itself, unchanged but with some properties now required and crucially additionalProperties set to false.
- a notificationBranch object which must either contain a further notificationBranch (recursive), or a notification object. Crucially the pattern properties for a notification Branch object exclude the well-defined names such as mob and in addition additionalProperties are set to false. This means the schema makes a clear distinction between a notificationBranch and a notification object.

Clarification/confirmation needed regarding:

1. source v $source. In the notification the schema specified source, but I have assumed that in line with the rest of SignalK this should be $source.
2. I have specifically forbidden the well-defined names such as mob, sinking, etc from being anywhere other than at the top level of notifications. I can't see any reason why they should be allowed further down and in fact if they were, one could imagine an implementation missing them if they were actually looking for top level notifications only. See alarms-invalid7.json for an example of this
3. which properties are required in the notification? Generally speaking one would expect that if a property is not required its default value would be well defined. I have assumed:

Required:

- "method" - without this how would the alert system know how to deal with the notification? I can't come up with a good rationale for which default could be set....
- "state" - the states [alert, alarm, etc] have specific meanings and without it being specified which would be the best to use as a default? 
- "message" - no use having an alert without a message

Not Required:

- "timestamp" - I can imagine a very simple sensor with limited functionality and no RTC which might want to raise an alert/notification. Therefore the timestamp if not supplied can be assumed to be 'now'. I'm sure  I saw some discussion/comment about timestamp somewhere but I can't find it now, perhaps someone could point me in the right direction.
- "$source" - Currently not required anywhere else, is there a view on whether the source should be required? Again thinking of a very simple sensor should we insist it reports a source? I think maybe some thought needed here as due to the way notifications are created not having a source might make things difficult. Also should this be the source of the value or the source of the alert? I propose leaving $source as not required for now and I will open up an issue for discussion on the topic.


bkp7:
@rob42 since doing this work I have questioned whether this should be recursive giving unlimited levels. Perhaps you could consider that point, happy to rework if just 2 levels required.

I am on Family holidays this coming week, but can look at again when back in March.

rob42:
The general concept of an arbitrary alarm or notification for key `x.y.z` is that its placed in `notifications.x.y.z`
That makes it possible for a gauge that is showing `x.y.z` to subscribe to notifications for only `x.y.z`, but the main screen might subscribe to `notifications.*`, and show a full list.

The concept behind `mob` and similar is to record events for which we dont have any sensors, and are important in nature.  I guess if we had a crew sensor we could have a gauge that showed "crew remaining", and "fires onboard" etc :-)

bkp7:
Yes, quite right, the full recursive design is correct.

As regards the MOB for another vessel, it will appear under that vessel's branch along with all the other notifications for that vessel. So a vessel could subscribe to `vessels.*.notifications.mob` which is no different than now.

rob42:
Re MOB - yes thats correct. 
Merging

PR #324, opened 2017-01-25
Define timestamp as JSON Schema date-time

bkp7:
Added the "format": "date-time" directive to timestamp definition within the definitions.json schema.

This identified many tests and samples which needed to be modified to comply with having the correct date format.

See #275


tkurki:
Thanks - good stuff.

tkurki:
One more point: the value needs to now be a real timestamp, but nothing checks that it is UTC.

PR #325, opened 2017-01-26
Improvements to the chart model

emilecantin:
- Added a "bounds" key to store the chart's bounds (useful when it's not part of the chart format)
- Added a "wms" chart type
- Added a "chartLayers" key for WMS charts.


tkurki:
@rob42 comments?

rob42:
Looks ok to me

Issue #322, opened 2017-01-24
Add timezone to allow local time to be displayed


bkp7:
All times are stated in UTC within SignalK.

A vessel may have several clocks on board which could be Signal K enabled. In order to allow a consistent time zone to be in place across the vessel I propose two optional properties being added to vessel:

1. "timeZone", a string value being the TZ (eg. "Europe/London") from the tz database http://www.iana.org/time-zones. The default value of this being "Etc/UTC"

2. "timeOffset", a number being the current offset with a default value of 0

This should work in all cases. A system with no ability to change or set a time zone would work with the defaults. A system with the ability to set offsets but which did not have external connectivity to get the TZ database and DST details etc, can still work using the static etc/... TZ values which can be hard coded. And finally a system with external connection can use the TZ database in full.

bkp7:
Pull request #328 deals with this. slight changes from the proposal above is that the new key is called timezoneRegion as timezone was already a key within environment.time. Also a default is not set for the new key, instead if the timezoneRegion is set then the timezone must also be set. What this means is that the server which sets the timezoneRegion must also expose the offset so that all systems wishing to show the time can use that offset without having any knowledge of the tz database and daylight saving rules etc.

PR #328, opened 2017-01-27
Added timezoneRegion to environment.time

bkp7:
This allows the timezone to be defined as a region and thus allow for daylight saving as is commonplace in other systems. Also added tests for the new key.  Deals with issue #322.


rob42:
I dont follow why we need the time zone region?
We store the dateTime in UTC, isnt the timezone a local UI specific decision?

bkp7:
@rob42, yes arguably it is......

However, we have been working on setting up many displays on board (basically replacing all fixed displays and MFDs etc with SignalK displays at 2 helm positions and other locations too) and setting the time zone for each is a real pain. This is designed so that when the vessel moves to a new time zone it can be set once on a server and then picked up by each display. There is no reason the UI would be obliged to use the server supplied time zone, nor is supplying the timezone mandatory for the server.

bkp7:
@rob42, sorry my last comment was misleading, I should have read my own notes again before commenting from memory. This was highlighted when we had multiple displays and wondered whether they would update for Daylight saving time.

There is already an existing a timezone property which supports an offset set in the server which then flows out to each display. A timezone eg. London could be set within the server (but outside the SK design) and the server could insert the correct offset in hours:mins depending on the time of year, which would allow for Daylight saving.

However the display/end user doesn't know whether the server has a fixed offset which will go wrong with DST, or whether a zone has been set which will vary the offset according to the time of year. If a display shows '12:05 (UTC+01:00)' I would not expect that to deal with DST, ie it's a fixed offset, but if a display shows '12:05 (Europe/London)' I would expect it to automatically take care of DST.

So, this change adds the optional timezoneRegion which the UI can display. Note that timezone must be present if timezoneRegion is present.

I have updated the description to provide more clarity.

Your point about naming the timezone as onboardTimezone, I think is a good one, but I have not changed that yet as both the existing 'timezone' and my proposed 'timezoneRegion' would logically become 'onboardTimezone' and 'onboardTimezoneRegion'. I'm reluctant to rename an existing property???

If we do decide to rename the existing property I think that 'onboardTimezoneOffset' and onboardTimezoneRegion' might be better choices.

rob42:
Actually I have another idea. Change `timezone` to `timezoneOffset` for clarity. If we leave the out 'onboard' bit out its more generic, and we can later re-use the time object in other trees, like `aircraft` and `fixed`.

bkp7:
Agreed and done.

Also slightly changed wording in description to say 'on vessel' rather than 'onboard', at least to be consistent with current 'vessel' use elsewhere.

bkp7:
Oops, finger trouble, sorry.

rob42:
Can you close this pls, no button for me

PR #330, opened 2017-02-02
Add Alternator definition, expanded charger definition

thomasonw:
Per conclusion from RFC0003, this addition adds an independent Alternator (and now Solar) section to the Electrical spec.  Also corrected some small spelling errors / typo.

To the existing Charger - added  'four stage' capability  (for the, rather uncommon, charge profiles which utilize the over-charge phase ) as well as a target / goal voltage (what voltage is the charger attempting to regulate to vs. the actual voltage being delivered)

### DIFF
--- a/schemas/groups/electrical.json
+++ b/schemas/groups/electrical.json
@@ -277,9 +277,9 @@
             }
           ]
         },
-        "chargerMode": {
+        "chargerRole": {
           "type": "object",
-          "description": "Mode of charger",
+          "description": "How is charging source configured?  Standalone, or in sync with another charger?",
           "allOf": [
             {
               "$ref": "../definitions.json#/definitions/commonValueFields"
@@ -299,7 +299,7 @@
             }
           ]
         },
-        "mode": {
+        "chargingMode": {
           "type": "object",
           "description": "Charging mode i.e. float, overcharge, etc.",
           "allOf": [
@@ -311,11 +311,11 @@
                 "value": {
                   "type": "string",
                   "enum": [
-                    "charging bulk",
-                    "charging acceptance",
-                    "charging overcharge",
-                    "charging float",
-                    "charging equalize",
+                    "bulk",
+                    "acceptance",
+                    "overcharge",
+                    "float",
+                    "equalize",
                     "unknown",
                     "other"
                   ]
--- a/test/data/electrical.json
+++ b/test/data/electrical.json
@@ -106,15 +106,15 @@
               "$source": "foo.bar",
               "value": "two stage"
             },
-            "chargerMode": {
+            "chargerRole": {
               "timestamp": "2014-08-15T19:00:15.402Z",
               "$source": "foo.bar",
               "value":  "standalone"
             },
-            "mode": {
+            "chargingMode": {
               "timestamp": "2014-08-15T19:00:15.402Z",
               "$source": "foo.bar",
-              "value": "charging float"
+              "value": "float"
             },
             "current": {
               "timestamp": "2014-08-15T19:00:15.402Z",
@@ -179,15 +179,15 @@
               "$source": "foo.bar",
               "value": "three stage"
             },
-            "chargerMode": {
+            "chargerRole": {
               "timestamp": "2014-08-15T19:00:15.402Z",
               "$source": "foo.bar",
               "value":  "slave"
             },
-            "mode": {
+            "chargingMode": {
               "timestamp": "2014-08-15T19:00:15.402Z",
               "$source": "foo.bar",
-              "value": "charging bulk"
+              "value": "bulk"
             },
             "voltage": {
               "timestamp": "2014-08-15T19:00:15.402Z",
@@ -219,15 +219,15 @@
               "$source": "foo.bar",
               "value": "three stage"
             },
-            "chargerMode": {
+            "chargerRole": {
               "timestamp": "2014-08-15T19:00:15.402Z",
               "$source": "foo.bar",
               "value":  "standalone"
             },
-            "mode": {
+            "chargingMode": {
               "timestamp": "2014-08-15T19:00:15.402Z",
               "$source": "foo.bar",
-              "value": "charging acceptance"
+              "value": "acceptance"
             },
             "voltage": {
               "timestamp": "2014-08-15T19:00:15.402Z",
@@ -252,15 +252,15 @@
               "$source": "foo.bar",
               "value": "three stage"
             },
-            "chargerMode": {
+            "chargerRole": {
               "timestamp": "2014-08-15T19:00:15.402Z",
               "$source": "foo.bar",
               "value":  "master"
             },
-            "mode": {
+            "chargingMode": {
               "timestamp": "2014-08-15T19:00:15.402Z",
               "$source": "foo.bar",
-              "value": "charging bulk"
+              "value": "bulk"
             },
             "voltage": {
               "timestamp": "2014-08-15T19:00:15.402Z",


tkurki:
I have found myself & others making repeatedly mistakes when working with just schema files - they are just not easily digested by the human mind.

Could you please include a sample document & an associated test that verifies it against the schema?

This will act as a safeguard in the future for making sure that one does not make schema changes with unexpected side effects.

See for example
- https://github.com/SignalK/specification/blob/master/test/electrical.js
- https://github.com/SignalK/specification/blob/master/test/data/electrical.json

It would probably be a good idea not to expand that single doc to cover everything, maybe create another one.

tkurki:
Why are you using `meta`? It has already a standard usage in SK, and this is different.

http://signalk.org/specification/master/data_model_metadata.html

rob42:
aka,  "meta" is a reserved word in signal K.

thomasonw:
Yes, will expand a sample test file to include new items.    

`meta` -- that is in the current electrical.json schema, I copied it forward unchanged so as to not risk breaking compatibility.   But it did seem odd.  I also found odd `timestamp` and `source` inside the `mode` object.  Are these not already brought in elsewhere?

rob42:
`timestamp`, `source` and `$sourceref` are added to a leaf node to provide the information on age and where/how the information was obtained. So they will appear for most leaf nodes.
`meta` holds information for alarm zones, and to control display in a UI. 
The ones you found may be remnants. Im not familiar with the electrical spec, so I will leave that to thers. @Tim maybe?

tkurki:
I think meta in electrical is probably my handiwork, but won't be able to look into this in more detail for a week or so.

Sorry about giving you hard time over it Al...

thomasonw:
Tkurki - No problem, it is no secret I my JSON / JS are minimal at best - so I try to follow existing examples (smile).

I did make a revision to this outstanding pull based on the feedback - removed the daily/weekly/current summaries from solar.  I add into the overall `dcQualities:`  `chargeTransfered` and `energyTransfered` to allow a consistent method for exposing this data from low level devices.  Does this work for people?

 Tkurki, if you (or someone else) is able to review electrical.js for consistency (`meta`, that `timeStamp` & `source` inside the inverter, and the use of `allOf` are all things I wonder about) would be great.  Once things are stable I can go back and fill out the test file you mentioned above.

schemas/groups/electrical.json
```diff
@@ -427,63 +504,126 @@
...
-                  "source": {
-                    "$ref": "../definitions.json#/definitions/source"
-                  }
-                }
+               "revolutions": {
+                "description": "Alternator revolutions per second (x60 for RPM)",
+                "$ref": "../definitions.json#/definitions/numberValue",
+                "units": "Hz"
+              },
+               "pullyRatio": {
```
tkurki: pulley probably

tkurki:
About meta usage: electrical.json uses it for metadata about an entity, namely battery: https://github.com/SignalK/specification/blob/0fe4e2bcfcceb6f4f01e091c1485df5d5fcc9763/test/data/electrical.json#L8-L16 and usage here is in line with that.

I think that this is in line with using `meta` to group metadata for a property.

tkurki:
As I stated previously the next step would be to either expand the current electrical.json test data https://github.com/SignalK/specification/blob/master/test/data/electrical.json or create a new one.

Contrary to what I stated previously I think it would be good idea to expand the existing one.

thomasonw:
@tkurki Thanks for the pointers - let me take some time and study this, as well as other resources and see if I can clean things up.  (Spelling as well, so sorry for that - lifelong issue).

thomasonw:
All - after getting some off-line hints, and studying / comparing with other schema I have updated the electrical spec as well as the associated test data file.  `npm test` passes on my machine, however the check-in test fails.  I can not see where there is a  missing timestamp for the charger mode in either the spec, not the test data.   Could it be that the check-in does not take into account the updated test data file??  

Outside that, does this round of edits seem to bring the schema up to date?

And I noted one small thing:  Existing items use plural vs. singular:  `chargers` vs. `charger`, etc.   I did not change this, but think it makes more sense to have them be singular.  In any case there is a mix now (`solar` for example) and should move towards one way or the other.

OK - those JSON experts out there, is this getting closer?  And any thoughts on the checkin failure?

thomasonw:
Anyone have an idea why the check-in test is failing?  The 1st reported item is:
```
  1) Electrical in the full tree Sample full tree is valid:

     AssertionError: [object Object]/vessels/urn:mrn:signalk:uuid:c0d79334-4e25-4245-8892-54e8ccc8021d/electrical/chargers/main/mode:Missing required property: timestamp
```

and yet there is a timestamp in the test data file
```
            "mode": {
              "timestamp": "2014-08-15T19:00:15.402Z",
              "$source": "foo.bar",
              "value": "charging float"
            },
```

More so, `npm test` passes on my machine before I pushed these edits.

tkurki:
Maybe you forgot to commit something? This PR does not have any changes to the test data.

thomasonw:
Yes, that is it.  As I have been pushing up changes to my github copy, they have automatically need being forward to Signal-k.  But it seems that is only for the original file, not the additional test data file!

OK, will get this figured out some how.

thomasonw:
And there it is.

thomasonw:
Any thoughts in this?  Are the edits for 'Meta' correct, or should the usage in the schema as it is today be retained?  Would like to get this Electrical group stable, am awaiting further additions (ala, BMS module) until then...

test/data/electrical.json
```diff
@@ -37,12 +35,12 @@
               }
             },
             "current": {
-              "value": 0.1,
+              "value": 124.2,
```
tkurki: Why these changes? Pretty heavy current...

schemas/groups/electrical.json
```diff
@@ -238,263 +252,428 @@
...
-                "allOf": [{
-                  "$ref": "#/definitions/identity"
-                }, {
+          },{
+            "$ref": "#/definitions/chargerQuantities"
+          }]
+        }
+      }
+    },
+    "alternator": {
```
tkurki: I think this should be plular, like `batteries` and `chargers`.

test/data/electrical.json
```diff
@@ -154,14 +157,133 @@
...
+              "value": 14.3
+            },            
+            "current": {
+              "timestamp": "2014-08-15T19:00:15.402Z",
+              "$source": "foo.bar",
+              "value": 12.2
+            }
+          }
+        },
+        "solar": {
```
tkurki: There are two `solar` branches.

schemas/groups/electrical.json
```diff
@@ -184,7 +198,7 @@
       "properties": {
         "associatedBus": {
           "type": "string",
-          "description": "Name of BUS source is assocated with (if applicable, may = NULL)"
+          "description": "Name of BUS source is associated with (if applicable, may = NULL)"
```
tkurki: If a property is not applicable it can be omitted, as it is not required by the schema. I would remove the NULL note.

schemas/groups/electrical.json
```diff
@@ -238,263 +252,428 @@
...
+              }
+            }
+          ]
+        },
+       "setpointVoltage": {
+          "description": "Target regulation voltage",
+          "$ref": "../definitions.json#/definitions/numberValue",
+          "units": "V"
+       }, 
+       "setpointAmperage": {
```
tkurki: Really not my expertise are, but shouldn't this be current?

schemas/groups/electrical.json
```diff
@@ -238,263 +252,428 @@
...
+                "$ref": "../definitions.json#/definitions/numberValue",
+                "units": "A"
+            },
+            "panelTemperature": {
+                "description": "Temperature of panels",
+                "$ref": "../definitions.json#/definitions/numberValue",
+                "units": "K"
+            },
+
+            "load": {
```
tkurki: Should this be a boolean, something like `connected` or `underLoad`? If the value is n/a then just omit this key, otherwise it is true or false.

schemas/groups/electrical.json
```diff
@@ -174,6 +178,16 @@
             }
           }]
         }
+      },
+      "chargeTransferred": {
+        "type": "number",
+        "description": "Amount of charge produced(+) or consumed(-) sense last report",
```
tkurki: What does "last report" refer to here? Also typo?

test/data/electrical.json
thomasonw: Just showcasing different capabilities between the two demo batteries - one large, one smaller.

schemas/groups/electrical.json
thomasonw: That was an overall question I had:  Use singular, or plural (but be consistent).  I note a mix throughout Signal-k:  Singular:  Chart, Tank, transmission, Depth, phoneNumber  -- and  plural: Waypoints, Notes, Routes.  (Do not believe there is a plural form of the noun `Solar` w/o making it `Solar Controllers`.  

 Q:  Is there an overall style for Signal-k?

test/data/electrical.json
thomasonw: Port and Starboard array, showing two different controllers.  Is the form incorrect for that?

schemas/groups/electrical.json
thomasonw: Edited - will include in next push refresh

test/data/electrical.json
tkurki: One solar with port and starboard as children. Like batteries for example.

schemas/groups/electrical.json
thomasonw: Removed n/a.
on/off --> Enabled / Disabled?

thomasonw: (Controller may not know if anything is actually attached to the port, only that it has turned it on)

schemas/groups/electrical.json
thomasonw: chargeTransfer and energyTransfer are integral measurements reset to 0 each time their value is reported up.   Would " ... reset to zero after each reporting" be clearer?

test/data/electrical.json
thomasonw: Tks, edited for next push

tkurki:
Re: plural vs singular: there are `tanks` and `batteries` in the current schema. `Propulsion` is a bit something else.

I would with plural where it fits and let `solar` just be `solar`.

schemas/groups/electrical.json
```diff
@@ -174,6 +178,16 @@
             }
           }]
         }
+      },
+      "chargeTransferred": {
+        "type": "number",
+        "description": "Amount of charge produced(+) or consumed(-) since last report",
```
tkurki: I still don't get the _since last report_ part.  Couldn't this be something like a trip odometer, reset manually and then keep on accumlating? What does _last report_ actually mean? Last delta? Then it makes a difference if you report this once per second or minute.

thomasonw: These are akin to trip meters, but I don't like manual (user initiated) resets; just will not get done on many installs. Some charging devices could auto-reset: alternators could start at 0 each time they have power applied, solar often resets accumulators at night. But not all are so simple. Wind – at times go on for days? AC charger held in Float mode indefinitely? And batteries..

My thinking for universal applicable is to reset each time an accumulated value is pushed up to the SignalK server. It does not matter if the reporting period is short or long. Time being a component of `J` and `C` no information will be lost, only smaller/larger accumulated values will be pushed up at each report. With the app aware of when a new value is available it could maintain its own a running sum(s) over the duration(s) it wishes to cover.

tkurki:
@timmathews @rob42 any thoughts on this? We have ignored this far too long. I am for merging.

thomasonw:
@tkurki your ears must have been burning the past week.  The one area I would like to suggest confirmation on is the chargeTransfered value.  The concept started out bringing forward information from Solar controllers, which often have something like ‘AHs produced this day’ – there was a bit of concern over those direct values, so my attempt was to generalize it with the chargeTransfered.  My question to people writing display aps, will chargeTransferd as defined be useable, or is there something more needed.

 

I do not want to continue this VERY LONG pull request, but would be good if we can get some confirmation this approach will work for the display apps.

sbender9:
I definitely think the `chargeTranfered` does not make sense without a time period. "since last report" does not make sense here, because here's no concept of that in this context. If we add the timePeriod, then I can display something useful. "chargeTransfered today" or "chargeTransfered the last 60 mintues", etc.

sbender9:
My Victron solar controller provides 'yield today' and 'yield yesterday' I'd like to see these and this is where they should be. Multiple values here that include a time range? Not sure.

thomasonw:
@sbender9 -- two thoughts:  

1. Do not packets come through with a timestamp?  The delta from the prior one would be the duration.  Would that work for the display apps?  (they would have to keep a sum of the values as they come in, applying the time delta as well.)

2. The solar controllers do often have daily / weekly / life values, they can also often determine when night is.   If we want to make it general, to cover say wind turbines - they do not have a ready day/night reference.

And we should perhaps consider that not all devices will have a reliable timer in them - using the Signalk timestamp would at least provide a level of consistency for all accumulated values.

Perhaps the phrase '..from last value' is too unclear, maybe use delta, or...?

tkurki:
Timestamp difference does not work:
- you connect and get the first one - what is the duration?
- you get deltas every 10/60 seconds - the number is pretty meaningless
- you use http api and get only the last timestamp. you fetch data again after x seconds and the timestamp is new - how many updates did the server get meanwhile?

thomasonw:
ok, timestamp is not sufficient.  But even if we added a duration value I see a real issue in your last bullet - where if the display app is pulling, as opposed to receiving pushed data, the app will have no way to reliably accumulate values over time.  

 

So -  does this lead back to using the originally proposed approach?  Device specific values - where the timeframe is predefined in the parameter and based on some point in time reference which is meaningful to that device. perhaps:  

 - Solar gets:  charge-transferred-today, charge-transferred-yesterday.

- Alts get:  charge-transferred-from-power-on

- Batteries get:  xx-this-cycle  (+ for charging, - values for discharging)

- Battery charger gets: xx-this-charge-cycle, xx-prior-charge-cycle

  - (No idea what to do about devices w/no natural reference point, e.g. wind)

tkurki:
I would like to get some of this merged and exclude the more problematic things for more work.

thomasonw:
@tkurki : Agreed, I pushed up an edit removing `Charge & Energy transferred`.

thomasonw:
@timmathews @rob42 @tkurki   <== so, is this ready to be merged?

schemas/groups/electrical.json
```diff
@@ -238,263 +242,427 @@
...
+                    "four stage",       
+                    "constant current",
+                    "constant voltage",
+                    "custom profile"
+                  ]
+                }
+              }
+            }
+          ]
+        },
```
tkurki: I know this has been a long battle, but please bear with me. I want to get this thing merged!

Not being well versed in anything but very simple charging setups the `chargerMode` and `mode` are not intuitively obvious. 

The mode enum values, all but _unknown_ and _other_, are all prefixed _charging_, so I was about to suggest that the property be named `chargeMode` and drop the prefixes. Furthermore enum values with spaces are look unwieldy to me. 

But then I discovered _chargerMode_ above. This property really needs better documentation. This is about multiple charger setups? Should the property name carry that word?

thomasonw: @tkurki  'chargerMode' defines the relationship between multiple (and paired) charging sources.  An example is when two inverter/chargers are configured to allow for the US power standard of 120v & 240v. Or to parallel two or more inverter / charger to increase the wattage capability.  In that case there is often a small cable between the two inverter/chargers and they run in master/slave mode.   Magnum refers this to 'Stacking', and other vendors have the same capability as well.  In this configuration, one acts as the master and the other(s) become slaves.

We could perhaps call it 'chargerConfiguraiton' instead if that is clearer.

The mode defines what phase a given charging source is operating in during a battery charge cycle (typically:  Bulk, Acceptance, and then Float).  I think the extra 'Charging ' and space is a carry over from before there was the common attribute of `chargerQuantities`, and I think it would be appropriate to remove that prefix, just leaving the Bulk, Acceptance, etc..

tkurki:
Should the property names be with or without the charging prefix? Algorithm, configuration, mode. I would go for _charging_ for all of them.

thomasonw:
I would also say without.

thomasonw:
I think adding _charging_ prefix would be fine.

thomasonw:
Please see changes I just pushed.  After thinking on it some, I believe `chargingMode` and `chargerRole` are perhaps more descriptive.  How does this sound to others?

tkurki:
Looks good to me! Asked Al to squash and submit a new PR.

thomasonw:
Closing the pull request (not integrated) as it is being replaced by a cumulative (squashed) pull request #375

PR #332, opened 2017-02-12
feature: add metadata handling to js lib

tkurki:
This PR adds functions `getMetadata(path)` and `getUnits(path)` to the published js module and uses these to include metadata in the main `/vessels`branch of the full model.

### DIFF
--- a/gitbook-docs/data_model_metadata.md
+++ b/gitbook-docs/data_model_metadata.md
@@ -13,6 +13,8 @@ Currently the Signal K server has a set of specific alarm keys. These grow over 
 Each data key should have an optional ```.meta``` object. This holds data in a standard way which enables the max/min/alarm and display to be automatically derived.
 ```json
 {
+  "units": "Hz",
+  "description": "Engine revolutions (x60 for RPM)",
   "displayName": "Tachometer, Engine 1",
   "shortName": "RPM",
   "warnMethod": "visual",
--- a/gitbook-docs/data_model_metadata.md
+++ b/gitbook-docs/data_model_metadata.md
@@ -38,7 +38,7 @@ The alarms functionality then becomes generic, and grows with the spec. This is 
 
 #### Meta.units value
 
-All keys in the specification must have `units`. If a client requests the `meta.units` for a valid key eg  `GET /signalk/v1/api/vessels/123456789/navigation/speedThroughWater/meta/units`, the REST interface MUST return proper value.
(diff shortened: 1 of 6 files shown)


rob42:
Not sure we need this. Isnt it equivalent to asking for `key.a.b.c.meta.units` ? Its then up to the server implementation to send back the correct value. Not sure why we need a second way of accessing it?

tkurki:
The idea here is to provide a standard client lib instead of everybody rolling their own. Implementation is client oriented. Maybe this should be in signalk-js-client.

Does java server provide units based on schema for all paths - also ones not present in its data set? Node server doesn't provide any units, need to fix that...

What if you need access to units without any SK server available? Like as part of a client's configuration. Maybe a bit far fetched...

rob42:
If its client oriented and not dependant on a signalk server then adding to signalk-js-client is the place for it.
If its adding to the REST api on a signalk server then its a duplicate of `key.a.b.c.meta.units` through the existing api. 
Having a convenience shortcut is fine, but will it lead to dozens of other shortcuts and possible confusion of the REST interface?

tkurki:
@rob42 is/should it be part of the spec that the server can provide path.meta.units for any SK path, not just ones currently present? Just thinking about implemenation on server and client.

rob42:
I think it should be implementation dependent. The fallback for the client is to pop a modal asking the user to select the units - probably with human translations, eh "Temperature" rather than "Kelvin"
In the end there will always be cases where a new or custom key appears on a boat and even the best server wont know what units.

rob42:
Are we agreed this is an implementation thing for clients and servers? eg the signalk spec should state  that when you request `key.a.b.c.meta.units` the server SHOULD(MAY?) provide it, with the lookup being implementation dependent. If it doesnt its up to the client to handle the problem.

tkurki:
I don't think it should be up to the implementation. Otherwise I agree on the meta.units.

rob42:
How would we specify the request in the spec? We already have the REST call to recover it, and if you make the retrieval of a unit a MUST, then its very difficult to meet the spec as there will always be additional units that the server is unaware of. 

I guess we can say " the REST interface  MUST return `key.a.b.c.meta.units` if the value is included in the published version of the spec. That means that you only need to be able to return units for a limited number of keys, the ones that are in the version of spec you support?
 And 404 for unknown ones?

Lets write that in and close this item. Where do I write that now?

tkurki:
Yes on _ the REST interface MUST return key.a.b.c.meta.units if the value is included in the published version of the spec...._

In data_model_metadata.md ?

rob42:
Yes to using data_model_metadata.md as the reference.

tkurki:
The discussion got a bit sidetracked - this just updates the js module. When node server starts to use this version the full model rest responses will include populated `meta` elements and the Node server can respond to arbitrary requests like _key.a.b.c.meta.units_ if the path is included in the published schema.

Issue #277, opened 2016-10-19
Schema files versioning


tkurki:
We have now reasonable versioning scheme for Signal K specification documentation (a major part of the Specification, in addition to the schema files).

However we are currently not versioning the schema in any effective way: the [published schema files](https://github.com/SignalK/specification/blob/1fdb0e1761f1f4b13c5ab2d3042c3047e5cb5873/schemas/signalk.json#L4) state
`"id": "https://signalk.github.io/specification/schemas/signalk.json#"``
which provides no way to specify a version.

We should be able to publish explicit versions and keep also development versions available.

See also http://snowplowanalytics.com/blog/2014/05/13/introducing-schemaver-for-semantic-versioning-of-schemas/

tkurki:
This came up with https://github.com/SignalK/specification/pull/276 when the published schema files were (a) different from local files (b) not the ones in `master`.

bkp7:
Getting the versioning right is vital, and it needs to be documented exactly what it means so each implementation can correctly deal with compatibility even against versions coming into existence after the product was released.

The example at snowplow you point to is well thought out. However, as things currently stand the SignalK schemas are very open. ie there is not much restricting of properties. As such if it is left like that it will be most unusual to see 'ADDITION' increments on this project as things change going forwards.

It is really vital that the changes immediately following release of v1 can be managed well as this will be the make or break of the project.

Finally, a decision need to be made as to the format of the version number. Should it be v1.0.1 or just 1.0.1 for example. Currently the Version numbers in the examples are 'varied'.

tkurki:
Todo:
- [ ] publish the schema under signalk.org/version/schema (currently http://signalk.org/specification/schemas/signalk.json)
- [ ] publish the docs signalk.org/version/docs/ (currently http://signalk.org/specification/master/ and versions in line with npm published artifacts like http://signalk.org/specification/0.0.1-1/)
- [ ] publish the npm package with versioning in line with the doc

rob42:
FWIW the npm `0.0.1-1-alpha` style is also valid maven structure. The suffix is allowed to have a numeric start,  which is a build iteration increment, and normally maven will just choose the latest (largest number) when you ask for a `0.0.1`.  It will sort these numerically, but if you use `0.0.1-alpha-1` the sorting is not numeric anymore ( it becomes alpha-numeric on the suffix) and it becomes unpredictable.

tkurki:
@SignalK/core 
[...]

PR #336, opened 2017-02-23
Pattern for version

bkp7:
Added a regex pattern requirement for the SignalK version.

In example and test files there was no consistency to the version. So as a tidying up exercise I have added a pattern which allows version formats like:
1.0.0
0.0.1-1
1.0.0-Beta

but disallows:
v1
v1.0
1.0

See [https://regex101.com/r/URVmwS/2](url) to test the regex and for unit tests.

I have also updated sample and test files so they have version in this specified format.


rob42:
This is a good addition. Follows the std java/maven pattern, which is well proven for CI and devops deployments. Add a ref to https://cwiki.apache.org/confluence/display/MAVENOLD/Versioning and we get precise definition and use for free :-)

bkp7:
@rob42 do we really want to link to their version numbering? It looks to me like we need a subset of it, but I'm not familiar with Maven at all.

How about adding to the description:
<model>.<revision>.<addition>-<stringQualifier> Cannonical version number with string qualifier. If the qualifer is present it is always taken as an earlier version than if not present. The stringQualifier is sorted in Unicode order.

This results in:
0.0.1-alpha
0.0.1-alpha1
0.0.1
0.1.0-alpha
0.1.0-beta

I don't think the exact usage of version numbers has been defined in SK yet - see #277, but I don't see that stopping this change going in now.

rob42:
maven is the defacto way to manage build lifecycles and dependencies in java. Its versioning scheme is well tested in a massive eco-system of java libs, so  its solid. Also very similar but more powerful than  semver mentioned in #277, and a lot like linux package versioning.
Referencing the maven doc just saves time on doing a detailed spec. Sorts out the tricky bits like `-SNAPSHOT` for nightly builds, and best practices.

Issue #171, opened 2016-02-12
Add vessel.type, and update spec to better support data obtained from AIS.


faceless2:
There is currently no space in the specification for a few bits of (potentially) important information which will be obtained from other vessels over AIS. Here's a list with suggested modifications to the spec to accomodate them. $vessel means any map under "vessels".

1. AIS type 5 (vessel static information) includes a "destination" field. This doesn't quite map to our "$vessel.navigation.destination" field, which has no space for a human-readable name, just a waypoint UUID. Suggest the addition of "$vessel.navigation.destination.name".

2. AIS type 5 again specifies four dimensions - distance from bow, stern, port and starboard to the position reported by the GPS. This is useful information - imagine you are trying to pass behind a 300m long supertanker in fog. We already have "$vessel.details.length" and "$vessel.details.beam" - I suggest we keep these but add optional fields "$vessel.details.bowOffset" and "$vessel.details.portOffset".

3. AIS class is useful - it gives us some information on the data that's going to be reported and the frequency of the reports (positions ever 30s for Class B, or every 10s for Class A) so we know if the data is still valid. Suggest adding a "$vessel.communication.ais.class" field which is "A" or "B" (or see last paragraph below for an alternative). 

4. There is no way to record broad "type" of vessel in the current spec. I'd argue this is useful information in general for our own vessels, but more so for others we see over AIS.

Specifically, the broad category of the AIS transmitter can be derived from the MMSI: for example any MMSI beginning with 970 is a search-and-rescue transmitter. We can also obtain more detailed vessel type information from AIS type 5 messages, which specifies one of about 100 codes including passenger vessel, hovercraft, "Tanker, Hazardous category A" (whatever that is) etc.

I would suggest using two fields for this data: "$vessel.type", which is an enum that includes at least the following values:
```
sart
mob
epirb
fixed
aircraft
ship (default if not specified)
```
While this is directly derivable from the MMSI, the sart (search-and-rescue transponder), mob (man-overboard) and EPIRB are things that you want clearly marked on your moving map, so I'd argue there's value in making this information explicitly available to a consumer. "fixed", "aircraft" and "ship" are also derivable from the MMSI prefix (ship being everything that's not one of the others, so we can make that a default if the value is missing). These values are enough to choose an icon to represent the AIS station on a map.

[...]

Issue #237, opened 2016-07-30
Validation: $source integrity


tkurki:
There should be an easy way to assert that the `$source` references are valid, eg. that `$source: "a.b"` has matching source under `/sources`.

tkurki:
https://github.com/SignalK/specification/issues/236

PR #337, opened 2017-02-24
Ais additions

rob42:
See #171
Added nearly all the missing AIS values from https://docs.google.com/spreadsheets/d/1wZTlvLNAN4GrXk3TfEtJne-76pmzCgBVRd_HMQfcwSQ/edit#gid=1961961172 as discussed in the issue.

Remaining to resolve:
* AidToNavigation See #338 

See #339 
* BaseStation
* Airplane
* MOB
* SART
* EPIRB
[sigh]


rob42:
No comments? Must be all good :-) 
Will merge in 24hrs if no objections

schemas/groups/navigation.json
```diff
@@ -293,6 +293,10 @@
       "description": "The intended destination of this trip",
       "type": "object",
       "properties": {
+      	"commonName": {
+          "description": "Common name of the Destination, eg 'Fiji', also used in ais messages",
+          "type": "string"
```
tkurki: No timestamp, source etc?

schemas/groups/navigation.json
```diff
@@ -435,14 +460,31 @@
...
+			"type": "object",
+			"description": "Special maneuver such as regional passing arrangement. (from ais)",
+			"allOf": [{
+        "$ref": "../definitions.json#/definitions/commonValueFields"
+      }],
+			"properties": {
+				"value": {
+					"type": "string",
+					"enum": [
+						"Not available (default)",
```
tkurki: Does `default` make sense here, or should it be just `Not available`?

tkurki:
Would it make sense to add support for these for these for NMEA0183/N2K at the same time, to validate that this implementation is feasible? Did we have somebody volunteer for doing either of those?

schemas/groups/navigation.json
sumps: I think the (default) is left over from a cut and paste from the AIS spec which states that this is the default setting when a Class A transponder is turned ON.

From Signal K point of view we will just show "Not Available" if this is what we have received from the AIS.

sumps:
I am not sure about the way you have used sources for AIS.

It is very important that for any vessel other than SELF in the schema that we have a clear indication of where the vessel data came from. In most cases it will be AIS, but in the future it could also be DSC, a mobile 3G/4G application, from the Cloud, etc.

I am not sure that you have addressed this in your use of sources. We already know from the source in a Delta message that a vessels position/cog/sog/etc. has come from an AIS but once the vessel is stored in the schema we lose this source. 

The sources group records the sources of data for SELF, it cannot be used to record the source for each vessel or have I missed something ?

rob42:
`sources` is specific to `vessels.self`. The `source` path is free-form, but SHOULD follow the norm of `sources.protocol.bus.device.data`, see #237  That should be enough to differentiate the source of the data?

I dont think its practical to record the history of the source of the data (blockchain anybody?) so when its received via signalk over tcp, I would record the source as either the claimed source in the message, or the actual source i received it from. What matters to me is how much trust I put in the data. If it came from a random internet relay, then I would probably want to know that. If its my VHF, then that is more trustworthy.

sumps:
No I understand about how the AIS receiver/transponder fitted to the SELF vessel can appear in the Sources, what I am referring to are all of the Vessels. We have no way of idenitifying for a particular vessel, where its data came from i.e. AIS, DSC, Marine Traffic, BoatBeacon App, Cloud service, etc. 

I think this is an important issue moving forward being able to track where other vessels, aircraft, marinas, etc. have come from.

tkurki:
Most of the fields in an AIS originated vessel object have $source fields
pointing to the AIS source, 0183 or n2k, so I do not quite agree on '
We have no way of idenitifying for a particular vessel, where its data came
from'.


ti 7. maaliskuuta 2017 klo 15.46 sumps <notifications@github.com> kirjoitti:

rob42:
If you receive a signalk update from vessel.x, you cant know if they  did receive the info from their onboard ais, or are just passing on some other vessels message. Hence we receive a `source` but its not tamper-proof so it has lower trust than our onboard ais receiver.
Of course the ais message can always be faked too, it sometimes is here in the pacific by dodgey foreign fishing boats in no-go zones, so Mk1 eyeball is best :-)
If we were to try to track the definitive history of data in a tamper-proof way we would need to use a blockchain or something, which is not possible without internet and a whole lot of complexity.

So my position on `source` is the same as other data, if its from `self` I trust it, if its from some-one I trust thats pretty good, from others I use it, but confirm separately.

schemas/groups/navigation.json
rob42: Destination is currently a composite object with `waypoint` instead ov `value`. Fixed it making `commonName`, `eta` and `waypoint` normal value objects

schemas/groups/navigation.json
rob42: Fixed

rob42:
@Paul are there any specific properties without $source that need change?
@tkurki re add support for these for these for NMEA0183/N2K, I think we merge and close this, then complete that as a separate pull. We can change if required, but if we wait then it holds up everything.

schemas/groups/navigation.json
```diff
@@ -315,17 +312,36 @@
...
+              ]
+            }
+          }
         },
         "methodQuality": {
           "type": "object",
           "description": "Quality of the satellite fix",
+          "allOf": [{
+						"$ref": "../definitions.json#/definitions/commonValueFields"
+					}],
```
tkurki: allOf should include properties, no? Same thing in integrity.

tkurki:
How about a single ais test file, that shows the data typically received for an AIS target in the full model?

tkurki:
There is quite a bit of mixed spaces/tab indentation in the examples - 2 spaces, please.

sumps:
A full example would be good as I really struggle to make sense of all this stuff. I know what an AIS sends out and if I can see where the data goes into each key I can give it my vote.

schemas/groups/navigation.json
rob42: The $ref pulls in multiple properties. This is how its done elsewhere.

rob42:
Best ref is https://docs.google.com/spreadsheets/d/1wZTlvLNAN4GrXk3TfEtJne-76pmzCgBVRd_HMQfcwSQ/edit#gid=1961961172

Quite difficult to do an example "AIS message mapped  to signalk" as there are 23 messages, and they are horribly complex and unreadable 6-bit values. eg `!AIVDM,1,1,,A,15MvJw5P0NG?Us6EaDVTTOvR06Jd,0*22`, is a type1 msg, which in the java server  is mapped to:
```
 vessels.366951720.mmsi.value=366951720,
 vessels.366951720.navigation.courseOverGroundTrue.value=5.740387909809349,
 vessels.366951720.navigation.headingTrue.value=0.005235987755982988,
 vessels.366951720.navigation.position.altitude=0.0,
 vessels.366951720.navigation.position.latitude=37.79895833333333,
 vessels.366951720.navigation.position.longitude=-122.33355,
 vessels.366951720.navigation.speedOverGround.value=0.0,
 vessels.366951720.navigation.state.value=Under way using engine,
```
Only option is to go through the message defs and spreadsheet really :-(

sbender9:
Do we really want aisShipType to be a number? We usually use an enum for this type of thing.

sbender9:
AIS data includes a "destination" name, do you think we should include that somewhere?

rob42:
We have a destination in navigation already. I dont want to duplicate the (rather horrible) ais structure, just as make sure we have a home for all the info they provide or need.

sbender9:
Yes, but there's no 'name' there. And no name under waypoint. Not a big deal, just might be nice to see.

rob42:
Its in this PR. `destination.commonName`

afischerdev:
I'm not very happy with the rule to set values direct to enums. In this case aisShipType and atonType.
In my two cents a server delivers a value and a client organizes the represantation.
Is there a way to a definition that handles both aspects? 
Otherwise the server has to convert the value to text, a client has to convert the text back to a value and phase it into its standard routines of represantation (include localization). I guess all AIS clients already have a routine for that.

sbender9:
Great point. I like not having to keep a bunch mappings up to date on the client side. But the language thing is an issue. 

This is not an issue specific to this pull request, it's all over the spec, so we should probably deal with this in 2.0 ;)

tkurki:
Rebased on master to pull 67d93d0 in.

tkurki:
@afischerdev I assume you created your _collection_ with https://github.com/sbender9/n2k-signalk/tree/ais-updates? I think that's a very good idea to get an idea how things look with real data. 

I would go as far as to create a test input file in this repo with full model from real world data, excercising the different ais additions in this PR.

tkurki:
Hmm the code by @sbender9 has sensor id as `ais` - so where id @afischerdev  your data come from?

tkurki:
I have to say that the aisShipTypes as a combination of number and string feel like the worst of both worlds. I see the multiple similar strings problem. 

How about using the numeric value here and adding the mapping to strings to the schema as free form metadata? 

```
"aisShipType": {
  "type": number,
  "multipleOf" : 1,
  "minimum": 20,
  "maximum": 99,
  "mappings": {
    "en": {
        20: "Wing in ground (WIG), all ships of this type",
        .....
    }
  }
}
```

A client can then choose to use this mapping or come up with its own.

afischerdev:
@tkurki server: no, I've my own copy of signalk-parser-nmea0183 and ggencoder and mixed this with a node server for testing.

@tkurki aisShipType: great idea I'll try it tomorrow

rob42:
@afischerdev @tkurki the enum vs number problem, especially the conflicts between NMEA, N2K and AIS vessel state is a mess. It came from using vessel state lists from various sources and appending as we added n2k, ais etc.
But it highlights two issues:
1) consistency - not much we can do about conflicting existing standards, but we should be consistent internally. Ais/NMEA etc will just have to be mapped to work.
2) The number vs enum, considering internationalization issues etc.

Using enums neatly limits the list, numbers are a bit too open to gaps and invalid ranges as we have less schema control over them. But English enums are not a good solution for internationalization, etc

The best solution to both problems I can see is a range of enums, used like constants, eg
```
"WIG_HAZARDOUS_A",
"WIG_HAZARDOUS_B",
"FISHING",
"HSC_..."
```
Those in turn can be mapped by files to any language using the common mapping_EN_NZ.json style of internationalization. In fact an aisNumberToSignalk.json could also map the raw AIS number to the enum, as could n2kToSignalk.json

We should do this for all enums, but probably V2 for exisiting enums. We may be able to make it backwards compatible  by a reverse mapping too?

If this is acceptable I'll change the pull to suit.

sbender9:
Unless I'm missing something, the problem with that approach is that a server can only support one language?

rob42:
No, any langauge. To support a langauge we create a file in a wellknown dir, eg in `intl/` we make `common mapping_EN.json`. Then since we want a german one, we create `common mapping_DE.json` with mapping to the german version.
I speak kiwi english, so I create a `common mapping_EN_NZ.json`
When the browser (or server) runs it picks up the appropriate file for the locale, `common mapping_[locale]_[variant].json` If its not found it recurses to `common mapping_[Locale].json`, eventually to `common mapping.json`.

afischerdev:
aisShipType and similar values:
From my view as a client developer I like the short way. 
Values comes in as integer and should forwarded as integer by a server. My client e.g. already reads direct the nmea sentence and all values have its place inside the app. So no need to get a description from server.

You can't fit all needs of a client. Localization is not the only problem a client have. Layout is an other one - e.g. abbreviations or short sentences in lists, different display sizes and so on.

Anyway, there may be some applications that require text, e.g. for log files or something. So it is good to have a text presentation of the integer. But not as a must. Btw the node server already has an translator inside modul ggencoder.

A short form with range control as sampled from @tkurki should be enough.

sbender9:
@rob42 should aisShipType be a numberValue? Shouldn't everything be a 'value'?

schemas/groups/navigation.json
```diff
@@ -435,14 +454,32 @@
         }
       }]
     },
+    "manouver": {
```
joabakk: check spelling

rob42:
@sbender9 I made aisShipType a 'property' rather than a  'value' as it doesnt  change. Once set thats probably it for the life f the vessel.

sbender9:
Ok. I was thinking more about it having a 'source', etc. there could also be more than one source for value. For example, from AIS and from MarrineTraffic.com.

rob42:
I will merge this is 14 hrs if there are no more  objections. Its starting to go stale

sbender9:
I would have loved to resolve my issue, but I vote for moving forward. Please merge this before my node server changes get stale too.

rob42:
@sbender9 - a difficult one - the shipType could come from two places, one being wrong or out-of-date I guess, but that can happen for mmsi number etc, and is a whole can-of-worms for later maybe :-(
@joabakk - sorry - forgot to fix spelling, done now.

Issue #316, opened 2017-01-04
Add Slack badge to readme


tkurki:
Like at https://github.com/open-guides/og-aws/blob/master/README.md

bkp7:
See #340

PR #340, opened 2017-02-24
Add slack badge

bkp7:
Added slack badge per #316


tkurki:
Hmm why is the github diff so extensive?

timmathews:
> Hmm why is the github diff so extensive?

Line ending changes? I assume the edits were made in Windows and core.autocrlf was not set correctly.

rob42:
Merging this as we should have the slack link, and the diff problem is not a big deal.

PR #342, opened 2017-03-04
Rewrite multiple sources/values documentation

tkurki:
The multiple sources/values documentation is really out of data and does not reflect current thinking and implementation. This PR adds a test case for multiple values in the full tree and....

TODO
- [ ] rewrite data_model_multiple_values.md

### DIFF
--- a/gitbook-docs/data_model_multiple_values.md
+++ b/gitbook-docs/data_model_multiple_values.md
@@ -1,6 +1,7 @@
 # Multiple Values for a Key
 
 There are two use cases for multiple data:
+
 * Multiple instances of a common device - eg two engines or multiple batteries
 * Multiple devices providing duplicate data - multiple values for the same Signal K key from different sensors, eg COG from both compass and gps or multiple depth sounders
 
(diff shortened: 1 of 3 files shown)


rob42:
@tkurki - are you happy with this change? We should merge it if so.

gitbook-docs/data_model_multiple_values.md
```diff
@@ -26,99 +26,153 @@ It is quite possible for a key value to come from more than one device. eg posit
...
-To see all the entries, use the REST api or subscribe to the parent object. A given device may choose to subscribe to a specific entry in the object, allowing multiple displays of the key, or users of the various values. The 'list' verb used in a query message can provide available keys.
+```
+{
+  "context": "vessels.urn:mrn:signalk:uuid:c0d79334-4e25-4245-8892-54e8ccc8021d",
+  "updates": [
+    {
+      "source": {
+        "label": "nmea2",
+        "type": "NMEA0183",
+        "II": {
```
tkurki: This is not correct. `source` should have just talker field, `II` and `sentences` are just in the full tree.

tkurki:
The exact format of the delta message(s) that result from a subscription depends on the chosen policy: for example with `instant` policy a client will receive updates as they become available. 

With `fixed` the server may send aggregate deltas (one delta with data for all sources) or individual deltas (for two sources two deltas every `period` - or even one delta per `period` per source, but staggered).

Should this be pointed out in the doc?

PR #344, opened 2017-03-11
Added SAR aircraft, updates tests to handle aircraft, add vessel.mmsi…

rob42:
… check

See #337


rob42:
Added aton as well.

rob42:
aton commit includes fix to design.json to use commonValues rather that source/timestamp

rob42:
The model I have used for these additions is to create a new root branch (eg aton) and duplicate the current tree we have for vessels, trimming the obviously useless bits like `sails` :-)
This means the same structure exists across all root object `vessels`, `aton`, `aircraft` and `sar` so there is minimal complexity in working with them if you already know `vessels`.

Remaining category is AIS coastal stations and AIS groups...

sbender9:
I don't see aton type anywhere?

sbender9:
Also aton name?

sbender9:
I do get a name for AtoN's from AIS.  On the chesapeake there are sometimes location/direction related: SC (South Center?), NE, SW, etc. Sometime's it's the buoy? number.

Issue #341, opened 2017-03-01
Add a Date and Time for when Trip Log was Reset


sumps:
In NMEA2000 the Date and Time that a Trip Log is reset is transmitted and we have supported this in iKommunicate. It would be good to include this in Signal K so that the data is not lost.

Add two new keys;

navigation.logResetDate
navigation.logResetTime

rob42:
@paul, I created a trip object, and combined date and time.

tkurki:
FIxed by #345

PR #345, opened 2017-03-11
Added trip.log, and trip.lastReset

rob42:
Created a trip object as I can see the trip data expanding as we start to want more info on voyage distance, vessel distance, leg distance, trip distance etc.
#341


tkurki:
LGTM, just that I am not quite sure if the bad examples add real value here - they are actually testing schema validation mechanism, not trip related stuff.

@sumps ?

sumps:
Yep this is good, we have used slightly different keys in iKommunicate but we will tidy that up later when we do a V1 audit. I think having a "trip" sub-group is a good idea.

rob42:
The trip bad test is just to confirm the schema is actually recognising and validating the `trip` object. I have had situations in the past where objects allowed `additionalProperties` and typos were not found.

Issue #303, opened 2016-12-08
Environment.water Sub-Group is a bit inconsistent


sumps:
Just adding some missing descriptions and noticed that the Environment.water sub-group is inconsistent as it has **environment.water.temperature** which is the temperature of the water mass that the vessel is sailing in, **environment.water.salinity** which is the the salinity of the water mass that the vessel is sailing in (so far so good) and then it tags on **environment.water.liveWell** and **environment.water.baitWell** at the end which is water but completely different water in a completely different location.

I think these two should be locations in their own right, like outside, inside and perhaps water should become waterMass or seawater, even though sometimes it may be lakewater or riverwater, but I think seawater is OK as a generic term for the water the boat os floating in.

If we are going to have environment sub-groups based on location we have to be consistent.

sumps:
Alternatively we could place the baitWell and liveWell as part of the environment.inside subgroup like a freezer as they may not be inside the hull but the wells are enclosed and not outside.

timmathews:
+1 for renaming water to seawater. Its a well-known term and I doubt any freshwater mariners would object in this context.

And I agree, there is definitely some inconsistency in **environment**. It was one of the first groups defined and we were still sort of feeling our way about. And there have been a lot of changes to this section see: #170, #186, #214, #216, #217.

**outside** was intended to contain environmental data about your immediate surroundings (excluding the water you're floating in). In fact, it used to be called **air**. Essentially, the sort of data that you might get from a weather station. When viewed in that light, the data that is contained in **outside** makes perfect sense. Of course in retrospect, water temperature and salinity should probably be contained in **outside**. Hopefully your seawater is _always_ on the outside.

**inside** on the other hand doesn't really follow _any_ rules. It's mostly temperatures (some very specific) and one generic **humidity**. Viewed that way, **liveWell** and **baitWell** belong in **inside** a lot more than they belong in **water**.

That said, **liveWell** and **baitWell** have more in common with tanks than environmental data. They should probably live under **tanks** and not in **environment** at all. That way we could also monitor their levels, which seems important.

Here's a handy cross reference between NMEA defined temperature sources and Signal K paths

NMEA Temperature Source | Current Signal K Path
----|----
00: Sea water | environment.water.temperature
01: Outside air | environment.outside.temperature
02: Inside air | environment.inside.temperature
03: Engine room | environment.inside.engineRoom
04: Main cabin | environment.inside.mainCabin
05: Live well | environment.water.liveWell
[...]

PR #346, opened 2017-03-11
Tanks

rob42:
See #303 
Removed `liveWell` and `baitWell` from `environment.water`, and added attributes to the `tanks` to compensate.
This will affect N2K mapping for pgn 130312. 
This cleans up the majority of the tanks part of #303


rob42:
Improved the `environment.inside` zones objects to have more data. These can be extended in the future after v1. If this is all good @bkp7, I will merge and close

PR #350, opened 2017-04-14
Add page specifying versioning in Signal K

rob42:
See #277.
The page specifies our versioning strategy. A combination of SchemaVer with a suffix to denote  interim or special releases

### DIFF
--- a/gitbook-docs/SUMMARY.md
+++ b/gitbook-docs/SUMMARY.md
@@ -20,3 +20,4 @@
 * [How Can I Help?](how_to_help.md)
 * [Appendix A: Keys Reference](keys/index.md)
 * [Appendix B: Changelog](changelog.md)
+* [Appendix C: Versioning](versioning.md)
--- a/gitbook-docs/versioning.md
+++ b/gitbook-docs/versioning.md
@@ -4,8 +4,6 @@ Our versioning is based on http://snowplowanalytics.com/blog/2014/05/13/introduc
 
 Most of this document is reproduced from there.
 
-# Signal K versioning is based on SchemaVer
-
 When versioning a data schema, we are concerned with the backwards-compatibility between the new schema and existing data represented in earlier versions of the schema. This is the fundamental building block of SchemaVer, and explains the divergence from [Semantic Versioning](http://semver.org/).
 
 Heres a simple formula for our SchemaVer:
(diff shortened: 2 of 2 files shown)


tkurki:
Good stuff! 

SNAPSHOT a pretty mavenish, but as such pretty understandable. How should it relate to development taking place in git `master` branch - is it ok that development takes place in master or should it be some.version-SNAPHOST?


Should this PR also actually carry out this? We need to at least 
- [ ] publish the schema under version specific urls (currently http://signalk.org/specification/schemas/signalk.json)
- [ ] publish the docs under version specific urls (currently http://signalk.org/specification/master/ and versions in line with npm published artifacts like http://signalk.org/specification/0.0.1-1/ NOTE no docs or similar in the path)
- [ ] publish the npm package with versioning in line with the doc

I think we should go for https://signalk.org/version/docs/ and https://signalk.org/version/schema/.

rob42:
When you have releases and dev versions you need some way to tell them apart. Linux uses odd numbers for dev, even for releases. The java world uses the maven *-SNAPSHOT suffix for dev, and the simple version number for releases. Java is a popular language, so a good chance the scheme is familiar to a dev.

SNAPSHOT is not normally  a branch name. The current_version-SNAPSHOT is in the master branch as normal, but the version attributes eg pom.xml(java), package.json(node), etc are set to -SNAPSHOT. 

In java-land maven has a 'release' process that removes the SNAPSHOT tag, builds and releases a version, increments the version number, and adds the -SNAPSHOT again. There is a similar grunt task in npm. In java the released artifacts are pushed to nexus or maven central, in node they go to npm repository.

For our needs we should publish under http://signalk.org/[version]/specification/, with an alias http://signalk.org/current/specification/ pointing to the appropriate release directory. We can have a http://signalk.org/[version-SNAPSHOT]specification/ dir for the current dev branch - good for dev reference and doc proofing .

We should have the same scheme for docs. I think it should be  https://signalk.org/version/* as it will be easy to create a listing of versions etc, and /schema and /doc will remain together.

I think this PR publishes the page, so its there, then another PR for a release tool and process. We have too many long-running PR's, If we are agreed on this part then lets not bog it down.

rob42:
One detail that works well is a small hidden panel in the header. When the page url is not the same as the current version (eg you are on a historic version) the panel appears with "This is a superceded version. Go [here] for the current version"
See https://docs.influxdata.com/influxdb/v0.9/

Issue #242, opened 2016-08-24
Magnetic Deviation


tkurki:
For example HDG NMEA 0183 has heading, I believe we are missing it.

A bit tricky, as it is associated with the sensor and current heading, so where should it go?

timmathews:
Deviation should be stored wherever we define static properties of a sensor. Its naturally a table of heading ranges and deviation angles. It should probably be added to the `headingMagnetic` and `headingTrue` values before they're sent out on the wire.

What is the NMEA convention here? In the HDG sentence, if deviation is present should it be added to the heading value or is purely informational and assumed to be already accounted for by the instrument?

A deviation table schema might look like:

```
{
  "deviation": [
    {
      "start": 0,
      "stop": 15.99,
      "deviation": 3
    },
    {
      "start": 16,
      "stop": 30.99,
      "deviation": -1
    },
    {
      "start": 31,
      "stop": 45.99,
      "deviation": 7
    },
    { ... }
  ]
}
```

sumps:
@timmathews further to our discussions last night, I think we should have a dynamic deviation value in Signal K that we take from the NMEA 0183 HDG sentence and NMEA2000 PGN 127250 if present that is the current deviation value that is being used by the heading system. I am not sure how many systems output this data but even if they do not, if we have deviation defined in Signal K then sensors can use it or if we also support a Deviation table, a Signal K server or client could interpolate the current deviation and populate the current deviation key.

The deviation table should be a table with deviation values every 15degs, so 24 values which then allows for a very accurate table with all values entered or a less accurate table with values for every 30degs or even 45degs.

sumps:
A server could also add value by working out magnetic variation using this code from NOAA...

https://www.ngdc.noaa.gov/geomag/WMM/soft.shtml

rob42:
This has more to do with correcting data than the real value of the heading or any other key.
If I understand correctly the deviation we are talking about here is the compass error on the N2K compass for the given heading? There is no point in storing both the inaccurate result and the correction, we should correct the data, then store the accurate value in heading.

This same adjustment is made by various sensors internally (calibration) and needs to be applied to some data by the server, eg wind vane offset. 

There was discussion on this recently (slack?) and I proposed a 'correctIncomingSensorData' functionality, but I think its an implementation choice. Not sure we can have a std format for the data we might need to store. Definitely a bad idea to start sending data plus corrections (that the receiver must apply anyway).

[...]

Issue #93, opened 2015-10-11
Handling invalid data values


sumps:
How should we handle invalid data values in the SK spec? Two examples are;
1) Depth value when the depth has been lost i.e. turbulence hàs caused the depth unit to lose the bottom reading, a depth sounder normally flashes the last reading.
2) GPS unit temporarily loses position and as a result all readings; position, COG, SOG etc. are not valid.

tkurki:
So the information we need to convey is that a sensor says "I'm operational & connected & ok, but I am unable to function for some reason" so it is not just the absence of data / updates.

Several options:
- send a delta without the `value` field or have a missing value field in the full tree. Both make normally required fields optional, making validation problematic
- replace `value` property with a special marker property like `dataMissing: true`, making either mandatory. Clearly marks missing values instead of just omitting them.

sumps:
I think having a special marker property like dataValid: false, would be a good solution and suitable for most sensor data. Depth and GPS are the obvious ones and we should start with these ones but we might need to add this for other sensors if we find that they have a valid flag or some other way of indicating data invalid.

rob42:
Hmm, this brings up other issues - how does a sensor communicate its condition via signalk. Looking at the simple cases:
1) ok
2) ok, but invalid data
3) failed, but alive
4) dead

Well 4) is easy - its just gone, so nothing, no comms etc. 
1) Since this is most common, we can take the position its all ok unless notified otherwise.

2)  For each data key we currently have `value, timestamp, source`. We could add an optional `state` key, that is sent to provide for case 2) . 

3) The `source` tree should have some way to register device failed for case 3) above. Maybe a `state` there too?

faceless2:
For what it's worth I'm not sure I see any practical difference between a sensor that is dead/missing, and a sensor that is alive but not reporting any usable values.

If this were a requirement then you could do this with a "last seen" timestamp on the source - this is the last communication with the source, which is distinct from the timestamp associated with any data fields it updates. But given I'm not sure I understand the problem this is trying to solve, this might not help.

joabakk:
Just came to think about this. I'm not sure how a depth sensor would react when the depth is greater than it could measure. But there is value in a measurement saying "it's really deep here", rather than "the sounder is not measuring since 2 mins ago"

sumps:
Traditionally depth sounders flash the last good value and they can lose depth for a number of reasons, not just going "out of their depth" in deep water. GPS position is another value that can become invalid.

[...]

PR #351, opened 2017-04-14
Add note on data accuracy

rob42:
See #242

Not critical, but best to say it to remove doubt.


rob42:
See #93 about invalid values

tkurki:
Expanded this a bit. If my addition is ok go ahead and squash & merge.

PR #353, opened 2017-05-04
Fixed typos and text formatting

aplathan:
Removed extra spaces from metadata description fields, and corrected m3 units to use superscript 3.


tkurki:
Thanks!

PR #355, opened 2017-05-22
 [WIP] Convert json id from https://signalk.github.io/specification/schemas…

rob42:
Convert json id from https://signalk.github.io/specification/schemas/ to https://signalk.org/specification/schemas/v1/

This includes all instances (including docs) in master, but probably doesnt address misc pulls, and maybe gh-pages.  All tests still pass :-)

### DIFF
--- a/index.js
+++ b/index.js
@@ -39,23 +39,23 @@ var FullSignalK = require('./src/fullsignalk');
 function getTv4() {
   var tv4 = require('tv4');
   var vesselSchema = require('./schemas/vessel.json');
-  tv4.addSchema('https://signalk.org/specification/schemas/v1/vessel.json', vesselSchema);
+  tv4.addSchema('https://signalk.org/specification/schemas/0.9.0-SNAPSHOT/vessel.json', vesselSchema);
   var aircraftSchema = require('./schemas/aircraft.json');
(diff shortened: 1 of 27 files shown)


tkurki:
Is there something left to do, as the PR name is [WIP]?

rob42:
Was waiting for #357

tkurki:
I don't quite get 0330e14b449d61056261dfa53f5deac54eb4cec6 : you add a schema:version task that replaces npm_package_version with the correct version, but at the same time modify all the schemas so that they have hardcoded `0.9.0-SNAPSHOT`in them.

index.js
```diff
@@ -39,23 +39,23 @@ var FullSignalK = require('./src/fullsignalk');
 function getTv4() {
   var tv4 = require('tv4');
   var vesselSchema = require('./schemas/vessel.json');
-  tv4.addSchema('https://signalk.github.io/specification/schemas/vessel.json', vesselSchema);
+  tv4.addSchema('https://signalk.org/specification/schemas/0.9.0-SNAPSHOT/vessel.json', vesselSchema);
```
tkurki: Let's extract a constant here and not duplicate (my copypaste, I know...)

tkurki:
I take back my comments, as I finally understand that `schema:version` will replace schema versions with the one from the package.json, whatever they are. The constant extraction is not really relevant either.

PR #357, opened 2017-06-03
Split gitbook doc keys list

### LINKED ISSUE #352: Keys list generation does not work
Generating http://signalk.org/specification/master/keys/ no longer works because of the very large number of keys added when the sar/aton/aircraft introduced so much duplication
- technically it fails with stack overflow when doing the .md => .html transformation
- usabilitywise it also fails: the list is goddamn long and has a lot of wasteful duplication

tkurki:
A very crude way to move forward with the docs: split the single
keys list to vessels and others docs.  Fixes #352.

Ping @rob42 - after this #355 can move forward.


tkurki:
image

tkurki:
No comments? I'll merge soon then, this is stopping schema id change that needs to be done asap.

rob42:
Looks fine - merge it, and #355 pls

PR #358, opened 2017-06-04
Documentation and schema fixes

rob42:
Items 5 and 19 on our TODO list, making sure the json in examples is valid and up to date


tkurki:
Hmm. I was sort of hoping and fearing it would come to this - I think we still have not reached a common understanding on how _sources_ and _$source_ should be used. The schema is really lenient on this.

Some of the examples you overwrite in this PR are actually copypaste from Node server output.

Let me explain:
image
The way this is supposed to read is 
- we have a top level source (bus, serial input - something discrete) called `nmea2`, whose label is `nmea2` and type is `NMEA0183``
- it has one talker (a multiplexed 0183 input may have several) and we have received one type of sentence from it, VTG, with last timestamp as indicated

So according to Node server in essence a NMEA0183 source's identity and the way it should be referenced from value tree is a combination of label and talker. In addition to that the source data retains a list of sentences received and their timestamps.

Contrast that to your example:
image

- I assume that you have `0183` as top level to group sources by type? If we want to do that I think it should be documented and somehow enforced. In Node server all inputs/sources are at the same level.
- You have `/dev/ttyUSB0` as an identifier. I can understand the reasoning, but it makes the $source references a bit cumbersome as a string having both `.` and `/`characters. Or is it just me that the reference like `"0183./dev/ttyUSB0.GP.RMC"` looks verbose and cumbersome? I think that information like `/dev/ttyUSB0` should really be metadata about the source, not part of its id
- why is `"type": "NMEA0183"` so far down the hierachy`? It is the type for everything under `0183` in your version and  `nmea2`in mine
- `talker` should be one level up - all the sentences from talker `"GP"` will have the same talker id
- in my version `label` is a label for the input, in yours it appears to be a label for a device, but in your version it is under sentence
- I don't think it makes sense to include the actual source data under `sources`. It is supposed to be metadata - if you want the actual data there are far better ways to get it. I don't mind a server making the original data available (Node server gives you 0183 over tcp by default), but I don't think it should be in the canonical examples in the docs.

I would like to strive for concise $source values.

Thoughts?

rob42:
I started from the definitions.json schema, 

Reference to the source under /sources. A dot spearated path to the data. eg [type].[bus].[device]",
"example":"NMEA0183.COM1.GP"

So the linux variant "0183./dev/ttyUSB0.GP", then in my data sample I have several $GP* sentences, I edited it down to "RMC" for simplicity. I agree about /dev/ttyUSB0 being a bit awkward - "ttyUSB0" would probably be as good really.

I added `label, type, talker` from your node examples, `label` is a required field.  You are right about `talker` it should be up a level.

Those changes result in:
```
	"sources": {
		"NMEA0183": {
		"type": "NMEA0183",
			"ttyUSB0": {
				"GP": {
					"talker": "GP",
					"RMC": {
						"label": "GPS-1",
						"sentence": "$GPRMC,061404.000,A,4117.6201,S,17314.8224,E,0.38,354.82,030417,,*11",
						"timestamp": "2017-04-03T06:14:04.451Z"
					}
				},
				"II": {
				"talker": "II",
					"HDM": {
						"label": "IMU",
						"sentence": "$IIHDM,318,M*36",
						"timestamp": "2017-05-16T05:15:54.006Z"
					}
				}
			}
		}
	}
```

rob42:
Re `label`, I never really understood what that signified. Reading the definition (RTFM!) you are correct. So I edited above to match.

tkurki:
How about
- removing the top "NMEA0183" level - not really needed
- gathering sentences under a distinct object. I think that mixing apples and oranges or talker id and sentences is bad practice and a talker often produces multiple sentences

So your example becomes:
```json
"sources": {
  "ttyUSB0": {
    "type": "NMEA0183",
    "label": "/dev/ttyUSB0",
    "GP": {
      "talker": "GP",
      "sentences": {
        "RMC": {
          "sentence": "$GPRMC,061404.000,A,4117.6201,S,17314.8224,E,0.38,354.82,030417,,*11",
          "timestamp": "2017-04-03T06:14:04.451Z"
        }
      }
    },
    "II": {
      "talker": "II",
      "sentences": {
        "HDM": {
          "sentence": "$IIHDM,318,M*36",
          "timestamp": "2017-05-16T05:15:54.006Z"
        }
      }s.    }
  }
}
```

You did not comment on my preference for not having the source data in the sources, which is metadata about sources?

rob42:
I prefer to keep the top level [type], as it partitions and reduces the potential size of the sub-branches. That may not seem to be a problem with one vessel, but if you assume this must also cater for a system with a thousand vessels, then collecting data without the top-level separation becomes a problem due to name/path duplication. Probably its badly named as [type], its really just a separator like `engine1`.

Looking at the `sentences` tag, and whether it should hold the actual sentence shows that we have not really done a very good job of defining the known types (0183, n2k, ais, etc) as our current structure just has a mix of all the tags in one level. 

At present its a confusing mix, and as we can see from this discussion it means different things to different people. I will try to sort something clearer today and update this

rob42:
First observation from looking at the spec again, is that the path under `sources` is valid for both our examples.
```
"sourceRef": {
      "type": "string",
      "pattern": "^[A-Za-z0-9-_.]*$",
      "description": "Reference to the source under /sources. A dot spearated path to the data. eg [type].[bus].[device]",
      "example":"NMEA0183.COM1.GP"
    },
```
So your path (at the very top) of `nmea2` is ok, as is `[uuid].[type].[bus].[whatever]`. since the `$source` holds this value and  points to the location in `sources`. The real issue is what the `source` object should look like.

Second is that when the `source` object is sent in a delta or full message its the object at the `$source` path, and does not contain path itself. So there is no difference in the outgoing messages in spite of having different `$source` values.

For the examples that this PR changes I will drop the top level as you suggest, since its still a valid path and  easier to understand as an example.

rob42:
Unfortunately the `source` object itself is a mess. The description says "An object containing **at least** the properties defined in properties", and since we have both 0183, n2k and ais properties, we obviously cant ever do all of them! Changing the description is easy, but it shows the structure is wrong. 

We have the required field `label` which is "A label to identify the source bus, eg serial-COM1, eth-local,etc ." It should be called something more descriptive `bus` or `bus-label`. And `type` should probably be an enum, `["0183","N2K","SIGNAL-K", "AIS", "OTHER"]. Then we should have specific objects for the named types, and a generic object for other. Add extra types as versions progress.

BUT that will cause a bunch of incompatibilities with current implementations. I'm not keen to dump that on our community, but I'm not sure how much of a problem it would cause AFAIK we currently dont use or rely on `source` much at all. So it would make existing implementations non-compliant, but probably not break them?

Comments pls

tkurki:
Hmm. I think there are some things mixed up here.

Delta spec says that there needs to be either a `source` or a `$source`. These refer to `definitions/source` and `definitions/sourceRef´. 

SourceRef is easy: essentially a dot delimited path referring to the `/sources` subtree under the server's root.

`source` object as used in delta is what you are quoting. This is the `source` in delta -  *not* the structure under `/sources` - these are two different things. The description is a bit off, but the structure allows for a combination of
- label,  type, src, pgn and optional instance (this is what Node server is outputting for N2K)
- label, talker, sentence (the three letter sentence type, not the sentence string)

We could split these to subschemas of their own and define `source` as `anyOf` them, including a _generic_ schema. BTW non-NMEA sources can be handled just by passing arbitrary `$source` values and not use `source` - this is what OpenPlotter does I believe.

You added `aisType` as a property in this delta `source`object. All the aisType examples are about having `aisType` under  `/sources`. `aisType` is also defined under `/sources`. Does aisType need to be in delta `source`?

So are we talking about `source` structure in delta or `/sources` subtree?

BTW an easy way to play with the JS code that convers `source` values to `/sources` is http://node-master.signalk.org/signalk-playground/ and use deltas without values

Try pasting this as delta:
```
{
  "context": "vessels.urn:mrn:imo:mmsi:230099999",
  "updates": [
    {
      "source": {
        "type": "NMEA0183",
        "talker": "II",
        "sentence": "VMW",
        "label": "0183-1"
      },
      "timestamp": "2010-01-07T07:18:44Z",
      "values": [
      ]
    },
    {
      "source": {
        "type": "NMEA2000",
        "src": "101",
        "pgn": 127488,
        "label": "N2K-1"
      },
      "timestamp": "2010-01-07T07:18:44Z",
      "values": [
      ]
    }
  ]
}
```

rob42:
I have the delta `source` object and the `sources` subtree confused.  Starting to see how this works, sorry but its an epic long post...
Your example above, when I put it through playground produces
```
{
  "vessels": {
    "urn:mrn:imo:mmsi:230099999": {
      "mmsi": "230099999"
    }
  },
  "self": "urn:mrn:imo:mmsi:230099999",
  "version": "0.1.0",
  "sources": {
    "0183-1": {
      "label": "0183-1",
      "type": "NMEA0183",
      "II": {
        "talker": "II",
        "sentences": {
          "VMW": "2010-01-07T07:18:44Z"
        }
      }
    },
    "N2K-1": {
      "101": {
        "n2k": {
          "src": "101",
          "pgns": {
            "127488": "2010-01-07T07:18:44Z"
          }
        }
      },
      "label": "N2K-1",
      "type": "NMEA2000"
    }
  }
}
```
So I am referring to the `sources` subtree here.
Looking at the structure the objects are in branches by the [label], eg `N2K-1`. The n2k object above in there has a structure like
```
    {
     [src] : {
        "n2k": {
            [src],
            [pgns] eg type specific sub-objects
        }
      },
     [label],
      [type]
    }
```
`[src]` is really a device id, and its equivalent in 0183 is `[talker]`. Having the same tag name would be helpful, eg `device`
For 0183  objects we have a different structure:
```
   {
     [talker] : {
         [talker] ,
         [sentences] type specific sub-objects
        }
      },
    [label],
    [type]
    }
```
For 0183 there is no object to group the 0183 specific attributes. Since we know that its 0183 because `type` says so its also not needed.  Currently the following is legal but nonsense:
```
 {
     [src] : {
        "n2k": {
            [src],
            [pgns] eg type specific sub-objects
        },
        [talker] ,
        [sentences] type specific sub-objects
      },
    [label],
    [type]
    }
```
It should be more like this, and limited by the schema to only have the appropriate keys based on `type` :  
```
 {
     [src] : {
           "device" :  [src],
            [pgns] eg type specific sub-objects
      },
    [label],
    "type" : "NMEA2000"
    }
....
{
     [talker] : {
            "device" :  [talker] ,
             [sentences] type specific sub-objects
      },
    [label],
   "type" : "NMEA0183"
    }
```
Then we can see that `device`=`parent_key` and `[label]` will equal path to `type.parent`
So we could get the same results from the much simpler:
```
 {
     [src] : {
            [pgns] eg type specific sub-objects
      },
    "type" : "NMEA2000"
    }
.....
{
     [talker] : {
             [sentences] type specific sub-objects
      },
   "type" : "NMEA0183"
    }
```
That gives a consistent structure for all sources, and one that can be enforced by the schema (I think).

rob42:
The for the delta version we just recover the `sources` object and flatten it.
```
 {
      "source": {
        "type": "NMEA2000",
        "device": [type.sibling,or pgn.parent],
        "pgn": 127488,
        "label": [path to type.parent]
      },
```
```
 "source": {
        "type": "NMEA0183",
        "device": [type.sibling,or sentences.parent],
        "sentence": "VMW",
        "label": [path to type.parent]
      },
```

tkurki:
In N2K land `src` is actually not a device id, that is separate, so the abstraction is misleading and obfuscates things. I don't think we need to abstract things here - N2K and NMEA0183 sources have different properties, so why not spell it out?

tkurki:
See also  https://github.com/SignalK/specification/blob/master/test/data/sources.json

tkurki:
We should also be wary of changing things just because of very minor issues, as there is more value in getting to 1.0 than getting every single detail absolutely correct.

tkurki:
`label`, `talker` and `src` are used as both subtree id and a separate independent property with value to allow retrieval of the value by itself with a fully qualified path ending in `.../talker`,`.../label` or `.../src` and/or as a value of a property in the json, not a property name. Somewhat same for the `type`.

rob42:
I agree that there is little benefit in making `sources` changes currently, this has become a monster, but at least I better understand it now.  So we have a couple of options:
1) Go with the current model, deal with the inconsistencies forever
2) Minimally change the format so we have an `0183` wrapper object to match the current `n2k` and `ais` objects. Move the `talker` and `sentences in there. It wont be ideal, but consistent with the others.
3) rework the whole thing as above 

I think 2) makes the most sense, and has little effect on existing implementations, eg its easy to add the wrapper to existing implementations.

I dont like 1) as I think data sources will become quite important in the future, both for security and for data integrity so I dont want an inconsistent format frozen into the spec.

I'm happy to make the schema changes.

rob42:
The pull request above makes two changes:
a) As in option 2 above the `talker` and `sentences` attributes now live inside a new `0183` object. This makes the NMEA0183 structure the same a s current `ais and `n2k` structure. 
b) Added an `i2c` object, and an `other` object for any other types not yet defined. I needed this as I could not get jsonschema to validate properly without defined object names.

rob42:
In summary the sources object in the full format now looks like:
```
 [arbitrary.path.to.object]
   {
     [src|talker|anything] : {
        [n2k|0183|ais|i2c|other]: {
            [src|talker|aisType],
            [pgns|sentences] eg type specific sub-objects
        }
      },
     [label],
      [type]
    }
```

tkurki:
Why is `ais` a sibling of `n2k` and `0183`? I would guess that 99% of ais data comes from nmea sources. Spec says that `aisType` is ais message type, but how is that an attribute of a source? ais-source.json does not use `ais` source subtree.

rob42:
The decision to make ais a sibling of 0183 and n2k was fairly arbitrary.  Until this PR there was no consistent 0183 object type, and ais seemed sufficiently unique in format and target use that I gave it a separate 'ais' object. 

See http://catb.org/gpsd/AIVDM.html#_aivdm_aivdo_sentence_layer The !AI* sentences have some similarity to NMEA sentences but not much. Also if we make it a child of 0183, then what about n2k, and vice-versa. I think it makes sense to keep it separate.

`aisType` is an attribute of source the same way `talker` and `pgns` are. It identifies the sentence type to allow decoding.

I dont see the error in ais-sources.json, can you clarify pls

tkurki:
What is the data that you would want to put in the `aisType` property? The spec says that it is the _ais message type_, see http://arundale.com/docs/ais/ais_message_types.html.

How can that be a single property associated with the source, since it is a part of every single ais message? If it _identifies the sentence type to allow decoding_ how is it related to the sentence? All ais originated value under `vessels/` will be pointing the `$source` property to this single value.

rob42:
Good point. `aisType` should be in a structure like `pgns`. I will change that. More problematic is identifying ais messages per vessel. The current structure doesnt do that, so even with multiple `aisTypes` they still collide from different vessels.
The path will need `mmsi` I think, eg 
```
 "sources": {
    "vhf": {
    	"ais-b":{
    		"12345678": {
					"ais":{
						"aisTypes": {
							"15":	"2010-01-07T07:18:44Z"
							}
						},
					"label": "AIS-12345678",
					"type": "AIS"
				}
				
			}
		}
  }
```

tkurki:
I don't think aisType belongs at all under sources, which is for source metadata.

rob42:
Isnt it equivalent to pgn or talkerid?

mxtommy:
New to this, but this caught my eye:

> assume this must also cater for a system with a thousand vessels, then collecting data without the top-level separation becomes a problem due to name/path duplication. 

If thousands of vessels are to be supported, I think sources should be grouped by vessel no? 

I'd think using the same identifier as used by /vessels. (so "/sources/[uuid/mmsi/self]/...")
Or, you could just move it to the vessels object (so "/vessels/[uuid/mmsi/self]/sources/...")

gitbook-docs/data_model.md
```diff
@@ -13,48 +13,88 @@ The simplest format is the full format, which is the complete Signal K data mode
...
+        "position": {
+          "altitude": 0.0,
+          "latitude": 37.81479,
+          "longitude": -122.44880152,
+          "source": {
+            "label": "/dev/ttyUSB0",
+            "type": "NMEA0183",
+            "talker": "GP",
+            "sentence": "$GPRMC,061404.000,A,4117.6201,S,17314.8224,E,0.38,354.82,030417,,*11",
+            "timestamp": "2017-04-03T06:14:04.451Z"
```
timmathews: Why have `timestamp` in the `source` and in the data (ln 90)? Also, are the values different because this is an example or do they actually mean different things.

tkurki: Source should not have timestamp.

gitbook-docs/data_model.md
```diff
@@ -19,19 +19,21 @@ The simplest format is the full format, which is the complete Signal K data mode
       "navigation": {
         "speedOverGround": {
           "value": 4.32693662,
-          "$source": "0183./dev/ttyUSB0.GP.RMC",
+          "$source": "0183.ttyUSB0.GP.RMC",
```
tkurki: This is not how the Node server works right now. I assume in this example 0183 is the label, but the ttyUSB0 does not belong in the server's internal, symbolic reference.

For NMEA0183 data there are just two hierarchy levels, the first for label and the second one for talker. A single nmea0183 input, identified by label, may have data from several talkers, identified by talker id. A talker has a list of sentences it produces with timestamps for the last received ones:

image

(Dunno what's wrong with SK talker id, that is probably an error in the new 0183 parser, need to check that.)

tkurki:
@timmathews apparently you are not aware of 4f36a3ed9af82be134022fafa48529f56dd210f7 in master that removes permissions.md, as discussed on Slack.

I don't find it particularly wise to cram all kinds of reorganisation to this single PR - this was supposed to be what the title says: _Update and fix signalk json in examples_. Long lived branches that contain all kinds of things are baaad.

Like for example all the example formatting fixes are no brainers, but some of the more involved examples, coupled to doc reorg, are not so straightforward to review.

timmathews:
I was not aware of that commit. I’ll merge master.

I think this is a perfectly reasonable way to manage the documentation fixes. Perhaps merging master more often would make sense though.

I’m not going to open a branch per documentation section.

timmathews:
Are there any outstanding reasons not to merge this into master?

gitbook-docs/rest_api.md
```diff
@@ -1,22 +1,62 @@
...
+      "version": "1.alpha1",
+      "signalk-http": "http://localhost:3000/signalk/v1/api/",
+      "signalk-ws": "ws://localhost:3000/signalk/v1/stream"
     },
-    "timestamp": "2015-08-31T05:45:36.000Z"
+    "v3": {
+      "version": "3.0",
+      "signalk-http": "http://localhost/signalk/v3/api/",
+      "signalk-ws": "ws://localhost/signalk/v3/stream",
+      "signalk-tcp": "tcp://locahost:8367"
```
tkurki: Typo

gitbook-docs/rest_api.md
```diff
@@ -1,22 +1,62 @@
...
+This response is defined by the `discovery.json` schema. In this example, the server supports two versions of the
+specification: `1.alpha1` and `3.0`. For each version, the server indicates which transport protocols it supports and
+the URL that can be used to access that protocol‘s endpoint. Clients should use one of these published endpoints based
+on the protocol version they wish to use.
+
+The server must only return valid URLs and should use IANA standard protocol names such as `http`. However, a server
+may support unofficial protocols and may return additional protocol names; for example, the response above indicates
+the server supports a `signalk-tcp` stream over TCP at on port 8367.
+
+A server may return relative URIs that the client must resolve against the base of the original request.
```
tkurki: Why add this? Not against, more curious. js client does not support this, does it?

gitbook-docs/subscription_protocol.md
```diff
@@ -1,194 +1,198 @@
...
 
-You can subscribe to multiple data keys multiple times, from multiple apps or devices. Each app or device simply subscribes to the data it requires, and the server and/or client implementation may combine subscriptions to avoid duplication as it prefers on a per connection basis. At the same time it is good practice to open the minimum connections necessary, for instance one websocket connection shared bewteen an instrument panel with many gauges, rather then one websocket connection per gauge.
+## Multiple value handling in subscriptions
 
-#### Multiple value handling in subscriptions
+A subscription to a key is for all the updates to that key. If there are multiple sources generating data for that key
+the client will get all their updates.
 
-A subscription to a key is for all the updates to that key. If there are multiple sources generating data for that key the client will get all their updates.
+If a client wants only the values of a single source it should subscribe to a path that includes the full path under
```
tkurki: Duplication from data_model_multiple_values.md, maybe should be here and reference in multiple values?

gitbook-docs/subscription_protocol.md
```diff
@@ -1,194 +1,198 @@
...
+`values` including the source reference key of the source. The source reference should be enclosed in square brackets:
+`navigation.speedThroughWater.values[n2kFromFile.43]`. The client can retrieve the relevant data via REST API.
 
-If a client wants only the values of a single source it should subscribe to a path that includes the full path under `values` including the source reference key of the source. The source reference should be enclosed in square brackets:  `navigation.speedThroughWater.values[n2kFromFile.43]`. The client can retrieve the relevant data via REST API.
+## Single use, or intermittent data
 
-#### Single use, or intermittent data
+When data is required once only, or upon request the `subscribe/unsubscribe` method should not be used. If the client
+is http capable the REST API is a good choice, or use `get/list/put` messages over WebSockets or TCP.
```
tkurki: Node server does not support GET/LIST/PUT over ws. Maybe I've forgotten something, but I have yet to see good justification for inclusion of it in the core protocol.

I suggest we remove these from the spec.

test/fullsignalk.js
```diff
@@ -95,6 +95,7 @@ describe('FullSignalK', function() {
     };
     var fullSignalK = new FullSignalK("urn:mrn:imo:mmsi:276799999", "mmsi");
     fullSignalK.addDelta(aisDelta);
+    console.log(JSON.stringify(fullSignalK, null, 2))
```
tkurki: Remove?

schemas/groups/sources.json
```diff
@@ -4,28 +4,90 @@
...
+    "sourceObject":{
+      "type": "object",
+
+      "patternProperties": {
+        "^((?!^0183$|^n2k$|^ais$|^i2c$|^other$|^label$|^type$|^manufacturerName$|^productName$|^softwareVersion$|^hardwareVersion$|^serialNumber$|^installationNote[1-2]$|^manufacturerInfo$)[A-Za-z0-9-])+$": {
+          "type": "object",
+          "oneOf": [
+            {
+              "required": [
+                "ais"
```
tkurki: I still don't understand how AIS is a type of source on its own. If i receive AIS data via 0183 or N2K, what is the source? If AIS why not 0183/N2K? 

Below an AIS source has message type property - what goes there? The type of the last message received? Why?

Or am i missing something here completely?

gitbook-docs/README.md
```diff
@@ -47,23 +53,29 @@ Signal K defines methods for combining arbitrary data from the Data Model into v
...
-Signal K does not define the transport or wire protocol. Signal K messages are JSON text and can be sent over almost any
-transport layer. However, we do provide guidance on how to make an initial connection, handle negotiation,
-subscription, and disconnection for a given transport (e.g. TCP/IP or serial).
+Signal K does not define the transport or wire protocol. Signal K messages are JSON text and can be sent over almost
+any physical transport layer. However, the Signal K standard does provide guidance on how to establish an initial
+connection, handle negotiation, subscription, and disconnection for a given transport (e.g. TCP/IP or serial).
 
-Where possible we use well established standards like HTTPS, REST, WebSockets, MQTT, STOMP, etc. But within each method
-there are always many options in message addressing, formating, or transfer (GET, POST), etc.
+Where possible Signal K uses well established standards like HTTPS, REST, WebSockets, MQTT, STOMP, etc. However, these
```
tkurki: This implies that we have defined MQTT and STOMP mappings, which is not the case. I'd remove them here.

gitbook-docs/data_model.md
```diff
@@ -1,104 +1,178 @@
...
+    },
+    "sources": {
+      "ttyUSB0": {
+        "label": "ttyUSB0",
+        "type": "NMEA0183",
+        "GP": {
+          "talker": "GP",
+          "sentences": {
+            "RMC": "2017-04-03T06:14:04.451Z"
+          },
```
tkurki: Indentation is off

gitbook-docs/data_model.md
```diff
@@ -1,104 +1,178 @@
...
+## Full Format
 
-The simplest format is the full format, which is the complete Signal K data model  as a JSON string.
+The full format is conceptually the simplest representation of data in Signal K. It contains all of the data from a
+Signal K node, which in the case of a Signal K server could me many hundreds of data points.
+
+```json
+{
+  "version": "v1.0.0",
+    "self": "urn:mrn:signalk:uuid:705f5f1a-efaf-44aa-9cb8-a0fd6305567c",
```
tkurki: Self should be relative to root

gitbook-docs/data_model.md
```diff
@@ -1,104 +1,178 @@
...
+          "value": {
+            "altitude": 0.0,
+            "latitude": 37.81479,
+            "longitude": -122.44880152
+          },
+          "source": {
+            "label": "ttyUSB0",
+            "type": "NMEA0183",
+            "talker": "GP",
+            "sentence": "$GPRMC,061404.000,A,4117.6201,S,17314.8224,E,0.38,354.82,030417,,*11"
```
tkurki: `Sentence` is in the schema and a string, but schema documentatio says it is the _sentence type_, not the sentence data. Node server handles only the three letter sentence type, not data.

If we define a way to pass along the raw data then we should do so explicitly and cater for N2K data as well. 

I think it would be a useful feature for debugging and testing and handling propietary N2K pgns, but that warrants a separate issue that includes software changes in RI.

gitbook-docs/data_model.md
```diff
@@ -1,104 +1,178 @@
...
+string or another object. Signal K keys that are object valued are object valued because the values don‘t have much
+semantic meaning individually. For example position – latitude doesn‘t have much meaning without an associated
+longitude. Therefore, these (and altitude) are grouped together in a single `navigation.position` key.
 
-The ordering of keys is also not important, they can occur in any order. In this area Signal K follows normal JSON
-standards.
+The values are always SI units, and always the same units for the same key. Therefore, `speedOverGround` is always
+meters per second, never knots, km/hr, or miles/hr. This means you never have to send units with data, the units are
+specific for a key, and defined in the data schema. A simplified version of the JSON schema with the units is available
+in [Keys Reference in Appendix A](keys/index.md).
```
tkurki: A link to `meta` http api to retrieve units would be useful here?

gitbook-docs/data_model.md
```diff
@@ -1,104 +1,178 @@
...
 
-The full format is useful for backups, and for populating a secondary device, or just updating all data, a kind of 'this
-is the current state' message.
+In addition to the `value` and `$source` (or `source`) properties, the data object may also have `meta` and `_attr`
+properties. These are discussed in detail in the [metadata](data_model_metadata.md) and [permissions](permissions.md)
```
tkurki: Remove permissions.

tkurki:
Hmm I just wrote comments for everything, would you like me to do the trivial stuff like typo corrections?

gitbook-docs/rest_api.md
timmathews: I didn't add it. This came from http://signalk.org/specification/0.0.1-12/doc/urls_etc.html which was merged in here. The client doesn't support it yet, but it could. I am however OK with removing it.

schemas/groups/sources.json
timmathews: Not my change, so I may not have the justification correct, but AIS could be received by a Signal K host directly via SDR or something. Also, giving them a source type of NMEA* sort of obscures their true origination. 0183, n2k, i2c, etc all imply a "local origination point". Whereas AIS is generally of a "non-local origination".

That said, this doesn't really belong in this PR. If we want to discuss adding AIS as a distinct source, that is it's own question.

gitbook-docs/data_model.md
timmathews: We have #396 open to address this I thought?

gitbook-docs/rest_api.md
tkurki: Minor, let's leave it.

schemas/groups/sources.json
tkurki: Ais stuff actually predates this PR, the changes seem to be related to formatting here. Just goes to show that autoformatting in a precommit hook or (enforced) by CI would make things simpler.

So this is not a stopper for this PR, a separate issue.

tkurki:
I am good for merging.

Issue #359, opened 2017-06-04
Polar diagram representation


joabakk:
In working with polars, a sort of table is required. For a specific wind speed, a table is made for the specific vessel. https://github.com/joabakk/signalk-polar does this by a database method, but has no place in the specification to store the table itself, only some of the main values.
If https://github.com/SignalK/freeboard-sk/issues/13 is to succeed or any other implementation of the same, such a table should be available in the schema in order to standardize.
Ideally, this should be open for any relative wind angles, to cater for any source of this table without interpolating or need to discard good data. I'm not sure this contradicts https://github.com/SignalK/specification/blob/cb728f95f8912b11b066ee8d05ffa329004bbf61/gitbook-docs/design_notes.md, and how this best would be captured in an organized manner, otherwise I would have done a PR

rob42:
Ah found my old code :-)
I stored the data by recording a point every time I crossed a 15deg heading boundary and stored it in a structure like:
`"performance.polar.[segment].wind": vesselSpeed` where [segment] is 0, 15, 30, 45... degrees. Wind was rounded up, but value was recorded for a segment. When a better result was obtained for the same segment/wind it was replaced.
 
That was enough for Kevin Kemps algo to derive the rest of the polar. The more you sail, the more accurate the polar.

rob42:
Hmm....given the Signalk-IMU gives attitude data, you could derive seastate, and have different polars for different wind/sea conditions.... @raffmont - a thesis for one of you students?

joabakk:
That is one approach, but given that the available data will vary in resolution, and that starting to use degrees here where the rest of the spec (sauf position) is in radians, it may violate the predictability of the paths. Here is an extract of an ORC certificate, where the main angles are most usually not in that 15 deg interval. I am storing any better data with the same wind speed and true angle, which makes the resolution close to infinity. There are many ways to skin a cat, but how do we best represent such a table in the schema?
<img width="541" alt="skjermbilde 2017-06-05 kl 23 26 26" src="https://cloud.githubusercontent.com/assets/2553029/26804004/b8ac96c6-4a46-11e7-80fb-26e010693201.png">

rob42:
deg vs radians - yes, but the degrees are just a marker and the equiv radians are not an int. It was easy at the time :-)
[...]

PR #360, opened 2017-06-15
Polar representation

joabakk:
#359 
We should try to couple sail inventory into here and also separate between a request for polar data from the response, if necessary.


joabakk:
Sails could be as simple as `sails.area.active`. That should be unique and cater for most vessel configurations without awkward sail configuration nomenclature

tkurki:
Polars are specific to a set of sails - I don't think just the total area cuts it in this respect.

Did you notice my work at https://github.com/SignalK/sailsconfiguration/blob/master/index.js#L68-L106

joabakk:
Not until now. Nice. It was on my agenda as well. 
My reasoning was tht most of the time, I have full main and full genoa, and that corresponds to a specific area. I'd be unlucky if I got a reefed status that corresponds exactly to a full main and full storm foresail. But of course you are correct tht it is not the same.

tkurki:
Think about multiple different sails (material, age, shape) with the same area dictated by class rules.

joabakk:
True. Then it's back to how do we name a configuration

rob42:
This is why I suggested  `performance.polar.[name].[seastate].angle.wind`. The `name` can be an arbitrary reference. It could also have a property `performance.polar.[name].sails` or something that referenced a sail setup, bit like `$source`

tkurki:
I don't think it makes sense to reference something else here, but we should have an in line structured representation of the sails setup. Adding an extra layer of indirection just obfuscates things.

`Maxi_Fenix_fullMain_fullGenoa` is one more example of trying to encode structured data into SK path instead of creating a proper structure for it.

joabakk:
I agree, but I may need some help on that. Can we have an array of sails with id and state for all active sails, or will the order inside the array make for duplication (main1+genoa1 vs genoa1+main1)?

@rob42 I have mostly followed your suggestion, except I switched angle to the end. That way if I poll `performance.polar.[name].[seastate].wind` I could get all angles

@tkurki re your question in issue https://github.com/SignalK/specification/issues/359#issuecomment-306809336 "I think the full set of values should be retrievable and a server may support querying for any combination. How should the server report back that "otherwise your http request is ok, but I have no polar data (extrapolation) for this combination" as opposed to "I have no such polar you're referring to"?", that may be solved by the location of the `null`? `performance.polar.name=null`would mean that there is no polar for that name/id etc?

joabakk:
Thinking about it a bit more, I think we should have a tiered approach to the polar representation. Many will not go in and enter sails, reef etc in the system, they will just sail. What you are talking about then is a full polar, where any combination of wind speed and angle requires the best combination of sails to achieve this target. Others again will enter everything and want a specific polar diagram. 
My https://github.com/joabakk/signalk-polar plugin currently does not store any of this additional information. Also, my front end webapp https://github.com/joabakk/signalk-polar-graphing has some hard coded design polars from the measurement sertificate of my boat. Ideally this should be entered by the user in the plugin and retrievable from a server API much like maps.

tkurki:
https://github.com/SignalK/polars-plugin

tkurki:
For inspiration see http://jieter.github.io/orc-data/site/

To summarise my work with the polars plugin:
- Polars are often expressed in csv format, one row per true wind angle, columns per true wind speed, often expressed in degrees and knots => the api format should be "friendly" to this format. The plugin allows entry in csv format, but the http API response is JSON, in SI units, and the original values in knots and degrees are available as human oriented labels
- The api data model assumes multiple polar sets and has the concept of one active polar set
- Metadata for a polar set includes _name_ and _description_ as well as id in the form of_uuid_, that can be used to uniquely identify the polar set in the api
- **Current implementation places the data outside the full Signal K data model**. I firmly believe that "one gigantic data model to rule them all" approach and the associated tooling is keeping us from expanding the SK API. I believe we should start expanding the SK api with modular, separate apis that are normally accessed individually. In practice this means that even if we place the url structure under `/signalk/v1/api/` the polar data will **not be included** in the response for a request for `/signalk/v1/api/`

API endpoints:
- `/signalk/v1/api/vessels/self/polars/`returns an object keyed by polar uuid's with object values containing polar structures
- `/signalk/v1/api/vessels/self/polars/active` returns the currently active polar set
- `/signalk/v1/api/vessels/self/polars/<uuid>|name` returns the identified polar set. If name refers to multiple polar sets with the same name (only uuids are unique) the server should return 400 Bad Request (not implemented)

The polar set data looks like this:
```
{
    "uuid": "85a9b66d-d01e-4a5a-912d-4a6b13a4db6d",
    "name": "aava",
    "trueWindSpeedLabelUnit": "kn",
    "speedThroughWaterLabelUnit": "kn",
    "trueWindAngles": [
        0.9075712110370513,
        1.0471975511965976,
        1.3089969389957472,
        1.5707963267948966,
        1.9198621771937625,
        2.0943951023931953,
        2.356194490192345,
        2.6179938779914944
    ],
    "trueWindAngleLabels": [
        "52",
        "60",
        "75",
        "90",
        "110",
        "120",
        "135",
        "150"
    ],
    "polarData": [
        {
            "trueWindSpeed": 3.086664,
            "trueWindSpeedLabel": "6",
            "polarSpeeds": [
                2.72140876,
                2.88603084,
                3.0352196000000005,
                3.0352196000000005,
                2.92204192,
                2.83973088,
                2.5773644399999998,
                2.1812425600000003
            ],
            "polarSpeedLabels": [
                "5.29",
                "5.61",
                "5.9",
                "5.9",
                "5.68",
                "5.52",
                "5.01",
                "4.24"
            ]
        },
...
```

TO DO / TO BE ADDRESSED:
- run/beat VMG sets to be added
- tack / gybe angle set to be added
- Title and description are human-oriented, but I think we should cross reference sail inventory once we get around to specifying it definitively
- not sure about the property naming yet, would appreciate second set of eyes. Polar set / polar data / polar speeds
- how to set the active polar set? POST the uuid to `..../polars/active` maybe?
- normal GET + POST, PUT, DELETE for CRUD operations
- "optimal polar set": afaik there such a thing as an optimal polar set, given the sail inventory, that only lists the optimal performance that should be achieved with optimal sails

I think this covers well enough the simple case (one polar set) and the more complex cases.

joabakk:
I like your thinking. But I would like the api response to be cleaner and SI only. Sure, we can arrange import and export to csv, but in real world SK use, I would imagine that you want to get angles and speeds for a specific wind speed, and do graphs or calculations based on that. 

I would say keep the uuid, name, add description, then windspeed {angle(rad), speed(ms), VMG(ms)}.

By adding vmg, the optimum tack and gybe values are there without having to re-calculate. Often these values are present in the generated polars, also in the one you linked to, except in the csv export. 
```
{
	"uuid": "85a9b66d-d01e-4a5a-912d-4a6b13a4db6d",
	"name": "aava",
	"description": "the best I could do last year",

	"polarData": [{
		"trueWindSpeed": 3.086664,
		"trueWindAngles": [
			0.9075712110370513,
			1.0471975511965976,
			1.3089969389957472,
			1.5707963267948966,
			1.9198621771937625,
			2.0943951023931953,
			2.356194490192345,
			2.6179938779914944
		],
		"polarSpeeds": [
			2.72140876,
			2.88603084,
			3.0352196000000005,
			3.0352196000000005,
			2.92204192,
			2.83973088,
			2.5773644399999998,
			2.1812425600000003
		]
	}]
}
```

webmasterkai:
Is there advantage to two different arrays instead of a single array of objects with two fields? Perhaps a `values` property that has the shape of `{ trueWindAngle: 0.9075712110370513, polarSpeed: 2.72140876 }`. Then other fields can be added if needed like labels.

joabakk:
I agree, and I meant to add vmg as well to each value. The reason I dod not want to split out wind directions, is that for my dynamic (brute) polar diagram, I don't have the same angles for each wind speed.

rob42:
Just looking at this again. 

The meta-issue with polars is that a given data point for vessel performance depends on many other data points, wind, sea-state, sails, angles, etc, etc. Its not practical to store all that extra data in the performance data structure  but its needed to make sense of the performance.

If the underlying data model in the server is a time-series db (TDB) then this becomes simple to solve. 

For a given time point in history we have all the other data we need. So we can answer the question "Am I sailing at optimal speed?" by searching for similar situations in the past and correlating with the current values. Stats methods like regression can provide help here. 

Real-time regressions may be beyond a single little RPi, but the same technique can derive the polars as a lookup table, eg  make the RPi do its processing at the dock while you are away working :-)

tkurki:
I get what you’re saying, but to me the use case here is not devining the polar performance values. 

To me this issue is about storing traditional polar data in the sk data model in a standard way for use in calculations and for display. There is demand for this I believe, for example Ian Boston’s sail instruments uses polar data for calc and display.

The data comes from Velocity Prediction Programs, not from history, and there is a de facto standard for displaying that data. 

Target speed and angle are also separate, discreet values that warrant a home in the data model in their own right, independent of how you devine them.

tkurki:
Side note: while interesting and useful I see two problems with the history approach.

No input data,  no output. A new boat, or no data for certain areas in the problem space.

How do you know you have been sailing at optimal speed? Maybe you have a systematic handicap in your trim. With history based performance targets the more you sail to those targets the more confidence the model has in them.

But as I said this is very interesting it is not the issue here. You can just output perf targets in this approach. But if you want to draw a polar graph based on data analysis you need a standard way to represent that in the data model.

rob42:
If theres a std format for storing polars we should support that, eg like  http://jieter.github.io/orc-data/site/
That will work for data  from Velocity Prediction Programs, and is a good comparison for you own performance.  I dont think we need to store any more than that data.

The history based approach is more about analytics based on actual performance. It should show up the effects of a new sail, clean bum, or trimming technique. It would allow the kind of skill or performance development that professional teams get ;-)

tkurki:
With _standard_ I meant the overall approach: a set of target speeds for different true wind angles for different true wind speeds.

I am not aware of any real standard or specification, the orc data I linked was for inspiration in terms of use cases, that type of graphs and tables I'd like to be able to produce & store.

tkurki:
How you derive the data is not a spec issue. How the data is presented in the data model is. Could we concentrate on that?

fabdrol:
Fubar, I wrote a comment but apparently didn't press "comment" before closing the window. Anyway, here we go again!

**My .2 ct:**
- I agree with the thought that the actual polar should not be part of the state tree itself. Polar is, IMO, a resource like charts, to be requested only when and if required. The state tree can simply reference the resource using an URI, like we do in other places. 
- It doesn't make sense to me to put human readable lables in the polar definition. We don't do that anywhere else, it's up to the client. If I have a mobile app that shows polars, I may want to have the option to show the speed in kts, m/s, mph, etc and change that on the fly. 
- I'd prefer a simpler format rather than a more complex format (like @joabakk's). That way, the data remains relatively simple and nobody's tempted to put more data points in when this is not yet supported. This could always be expanded in future versions
- I'd propose adding the following properties to the data structure: `version` (since the file may be stored on disk somewhere - making it portable - it would be good to know what SK version it was intended to be used with) and `timestamp`, to make historical polar data possible at some point
- I need to read up on run/beat VMG and tack/gybe angles before I can add them to the data model, as I don't fully understand there place in the data right now. 

Something along these lines, mixing @joabakk and @tkurki 

```json
{
  "uuid": "ED7498BD-0759-4152-9951-6F795410C2F5",
  "name": "my amazing polar",
  "version": 1, // major only
  "trueWindAngles": [
    0.9075712110370513,
    1.0471975511965976,
    1.3089969389957472,
    1.5707963267948966,
    1.9198621771937625,
    2.0943951023931953,
    2.356194490192345,
    2.6179938779914944
  ],
  "polarData": [
    {
      "trueWindSpeed": 3.086664,
      "polarSpeeds": [
        2.72140876,
        2.88603084,
        3.0352196000000005,
        3.0352196000000005,
        2.92204192,
        2.83973088,
        2.5773644399999998,
        2.1812425600000003
      ]
    },
    // ...
  ]
}
```

joabakk:
I'm all for simplifying. I think we have different ways of thinking of polars. My approach has been the brute force one, storing values as I go along and looking for the best speed for each angle in similar winds. So the simplified version would not work for this. I may have arbitrary angles for every increment of wind in the database. I use the design values, but only for a visual check in the live polar diagram. If I had to choose between interpolating and using historical data, I would pick the latter. My response from today's sailing to `/plugins/signalk-polar/polarTable/?windspeed=4&interval=0.1` is:
```
[
{
"angle": 2.4310182916991168,
"speed": 1.8931560351550847
},
{
"angle": 2.4321784118257836,
"speed": 1.8880115894073806
},
{
"angle": 2.4870479266767274,
"speed": 1.7182448797331473
},
{
"angle": 2.4943590198431944,
"speed": 2.1658116597833983
},
{
"angle": 2.4992905370767002,
"speed": 2.1555227682879905
},
{
"angle": 2.569216349095629,
"speed": 1.635933747769883
},
{
"angle": 2.5705140515793925,
"speed": 1.4507337008525376
},
{
"angle": 2.591025019680996,
"speed": 1.7182448797331473
},
{
"angle": 2.602480641269265,
"speed": 1.8520004691734524
},
{
"angle": 2.6108355271541286,
"speed": 1.8520004691734524
},
{
"angle": 2.6173922446591167,
"speed": 1.8057004574441158
},
{
"angle": 2.648219438741466,
"speed": 1.8880115894073806
},
{
"angle": 2.670943732088495,
"speed": 1.8468560234257483
},
{
"angle": 2.703800337825788,
"speed": 1.5176114955726903
},
{
"angle": 2.709113311220408,
"speed": 1.7696893372101876
},
{
"angle": 2.722098368337149,
"speed": 1.7542559999670757
},
{
"angle": 2.729997739051836,
"speed": 1.8931560351550847
},
{
"angle": 2.730079941826865,
"speed": 1.8777226979119723
},
{
"angle": 2.73025825278494,
"speed": 1.5639115073020264
},
{
"angle": 2.7316271097801654,
"speed": 1.8931560351550847
},
{
"angle": 2.75323092197041,
"speed": 1.5433337243112102
},
{
"angle": 2.849930292882592,
"speed": 1.8468560234257483
},
{
"angle": 2.857298276791531,
"speed": 1.7645448914624837
},
{
"angle": 2.8996229215331235,
"speed": 2.1761005512788065
}
]
```

So the simplified one (I may be contradicting myself) would mean I would have to create one named polar table for each of the combinations of wind and wind interval, otherwise the angles would not match the speeds recorded.

My plugin API does the db lookup each time, so it may be a hog, but I like it and would like to keep it. That's why I don't want the angles to come first and then the wind speeds. May I propose:

```
{
	"uuid": "ED7498BD-0759-4152-9951-6F795410C2F5",
	"name": "my amazing polar",
	"signalKVersion": 1,
	"polarData": [{
		"trueWindSpeed": 3.086664,
		"trueWindAngles": [
			0.9075712110370513,
			1.0471975511965976,
			1.3089969389957472,
			1.5707963267948966,
			1.9198621771937625,
			2.0943951023931953,
			2.356194490192345,
			2.6179938779914944
		],
		"polarSpeeds": [
			2.72140876,
			2.88603084,
			3.0352196000000005,
			3.0352196000000005,
			2.92204192,
			2.83973088,
			2.5773644399999998,
			2.1812425600000003
		]
	}]
}
```

tkurki:
I think we are dealing with specifying a new api, not a file format. Like joabak’s query param usage.

A good way forward might be to work this out in code: harmonize joabak’s work, the polar plugin that I wrote and Ian Boston’s sailinstruments so that they work together.

With angles we can probably eat and save the cake: if they are available on the top level the speeds relate to them, if on the lower level then the speeds are for those. This would allow also for partial data.

fabdrol:
Okay, so we'd end up with something like this?

```
{
	"uuid": "ED7498BD-0759-4152-9951-6F795410C2F5",
	"name": "my amazing polar",
	"signalKVersion": 1,
        "trueWindAngles": [
            0.9075712110370513,
            1.0471975511965976,
            1.3089969389957472,
            1.5707963267948966,
            1.9198621771937625,
            2.0943951023931953,
            2.356194490192345,
            2.6179938779914944
          ],
	"polarData": [{
		"trueWindSpeed": 3.086664,
		"trueWindAngles": [
			0.9075712110370513,
			1.0471975511965976,
			1.3089969389957472,
			1.5707963267948966,
			1.9198621771937625,
			2.0943951023931953,
			2.356194490192345,
			2.6179938779914944
		],
		"polarSpeeds": [
			2.72140876,
			2.88603084,
			3.0352196000000005,
			3.0352196000000005,
			2.92204192,
			2.83973088,
			2.5773644399999998,
			2.1812425600000003
		]
	}]
}
```

joabakk:
I think the ORC standard approach with 52, 75, 90 etc degrees struggles when you want to include beat and run angles, ehish is what you really want. So the high level angle array may not be all that useful. I am all for harmonizing, where do we start? Should the plugins push the available polar tables to the server in some way?

tkurki:
We have JSON, we can have a proper representation for beat and run angles and not try to stuff everything into a few arrays.

Can we make it so that my plugin's visualisation shows data from your plugin? Query the server for a "default" set of polars via http, then plot what it gets, allowing for missing data. 

If you don't enable my plugin it will not create any http endpoints, but the webapp will be usable.

joabakk:
OK @tkurki about beat and run. Thinking about this, I'm not sure how to get my brute polars into your webapp. It could give the total table, but updating that everytime the table is called would probably be too much overhead for most. It could be a snapshot though.

fabdrol:
@joabakk is this still [WIP]? Almost a year old!

joabakk:
Hard brain work @fabdrol. I should be able to make a move within a few weeks

fabdrol:
@joabakk I'm confident your brain can handle it! let me know if I can help.

joabakk:
I'm currently leaning towards using the structure from my post on 6 Jun 2018, but instead of signalKVersion (or in addition), use the well known source. 

```{
	"uuid": "ED7498BD-0759-4152-9951-6F795410C2F5",
	"name": "my amazing polar",
        "active": true,
	"signalKVersion": 1,
        "source": "signalk-polar",
	"polarData": [{
		"trueWindSpeed": 3.086664,
		"trueWindAngles": [
			0.9075712110370513,
			1.0471975511965976,
			1.3089969389957472,
			1.5707963267948966,
			1.9198621771937625,
			2.0943951023931953,
			2.356194490192345,
			2.6179938779914944
		],
		"polarSpeeds": [
			2.72140876,
			2.88603084,
			3.0352196000000005,
			3.0352196000000005,
			2.92204192,
			2.83973088,
			2.5773644399999998,
			2.1812425600000003
		]
	}]
}```

I will now abandon my approach of arbitrary angles, and conform to specific intervals (which may be adjusted). If this is an approach we can agree upon, I can take @tkurki 's challenge and include some interoperability to prove implementation.

**Comment by fabdrol** [comment]
@joabakk what's the `active` for? Is that an indication that this is used by the server to calculate other data? What if I have more than one "active" polar (i.e. for different sail configurations)

In terms of the specific interval, should that be part of the polar specification (so that different polars use different intervals, etc)?

Cheers!

**Comment by tkurki** [comment]
Recalling thinking here _active_ is _current_, so that everything that builds on polars (perf calculations, visualisation) can share this fact and you do not need to configure them separately.

**Comment by joabakk** [comment]
Sure, no need to include active here then. Otherwise OK? My thinking about activating is that you should be able to show/hide polars, and select the master from which derived parameters (beat/run/vmg) are calculated

**Comment by rob42** [comment]
Rather than include 'active' in the polar, it would be better to have an external key that points to the correct polar, eg 'performance.polar.current:  uuid'

**Comment by tkurki** [comment]
An explicit, external key is a good idea, but from an api consumer’s perspective a mechanism like what we have with `self` would be convenient: a well known single path to access the active polar instead of forcing the client to do two requests. This would be for reads.

Then we need to define how writes happen, maybe a PUT to the explicit key path?

**Comment by fabdrol** [comment]
Aren't PUT writes already standardised @tkurki? In any case, I feel that that is out of scope of this PR and would just hold it up.

**Comment by joabakk** [comment]
Ok let's keep active out of this for now. I will work on this basis. Another thing is that a polar array may show several `trueWindSpeed`s, and you may not want to show all in a graphic representation. 
Anyway, I will start with this and see where it leads us.

**Comment by tkurki** [comment]
If we create both a convenience path like `...performance/polars/currentData` (that provides the currently selected polar data) as I suggested above and have `...performance/polars/currentPolarId` (that holds the id of the currently selected polar) as Rob suggested we need to specify how you update the currently selected path:
- by PUTting a whole polar data set to `...performance/polars/currentData`
OR
- by PUTting just the id to `...performance/polars/currentPolarId`

I don't think this is such a huge deal. If it holds up the polars in general we just push the whole _current/selected_ handling further and add just the basic polar structure.

**Comment by fabdrol** [comment]
> I don't think this is such a huge deal. If it holds up the polars in general we just push the whole current/selected handling further and add just the basic polar structure.

Let's do that, finish this and move on right away with a _separate_ PR for that spec. It's physically a different thing anyway, right? I.e. this is JSON schema stuff whereas that is documentation.

**Comment by joabakk** [comment]
PR updated. Anyone for merge?

**Comment by fabdrol** [comment]
Content looks good to me. @tkurki ?

**Comment by joabakk** [comment]
Thanks @tkurki. 

So it would be something along the lines of:
```
{
  "uuid": "urn:mrn:signalk:uuid:c0d79334-4e25-4245-8892-54e8ccc8021d",
  "polars": [
    {
      "a56bc52a-877e-41f1-a977-f81f8d50c68f": {
        "name": "my amazing polar",
        "source": {
          "label": "signalk-polar"
        },
        "polarData": [
          {
            "trueWindSpeed": 3.086664,
            "trueWindAngles": [
              0.9075712110370513,
              2.6179938779914944
            ],
            "polarSpeeds": [
              2.72140876,
              2.1812425600000003
            ]
          },
          {
            "trueWindSpeed": 4.086664,
            "trueWindAngles": [
              0.9075712110370513,
              2.6179938779914944
            ],
            "polarSpeeds": [
              3.72140876,
              3.1812425600000003
            ]
          }
        ]
      }
    },
    {
      "a54ffe69-b8f6-47c8-9437-ccea8ecc2d77": {
        "name": "last season polar",
        "source": {
          "label": "signalk-polar"
        },
        "polarData": [
          {
            "trueWindSpeed": 4.086664,
            "trueWindAngles": [
              0.9075712110370513,
              2.6179938779914944
            ],
            "polarSpeeds": [
              2.71,
              2.17
            ]
          }
        ]
      }
    }
  ]
}
```

**Comment by tkurki** [comment]
Yes.

I don't think the true wind angles need repeating, should be the same for all wind speeds. See the earlier snippets.

A description field would be useful.

I think we should support the data available at http://jieter.github.io/orc-data/site/, so we should have  also gybe & beat/tack angles and VMG.

**Comment by joabakk** [comment]
@tkurki I have added vmg, beat and run. The example is taken from orc-data, and I have left the angle duplication in order to include the beat and run angles for each wind speed. I have also made beat and run arrays so you can have an asymmetrical polar. 
I think the definition and the example match, but the test passed long before I had made many of the changes. Not sure why that is

**Comment by joabakk** [comment]
Actually hold on with merge. There is something strange with the test. As long as the top level is array, the test will accept anything. I don't find any examples to compare

**Comment by joabakk** [comment]
I'm happy now. The tests ar passing but I have also tested manually for invalid json samples, so the test is no longer passing just because it was an array, as was the case in the last commit. @tkurki could you have another look and lift the change request if you are happy?

**Comment by joabakk** [comment]
Thanks @tkurki 
I have added a real world example from my sailing history (don't judge), and there you can see the rationale of arrays for beat and gybe values, as the performance form one tack to the other may vary slightly.

**Comment by joabakk** [comment]
V2.0 of signalk-polar and V2.0 of signalk-polar-graphing now conforms to the proposed spec. You can enter a csv and have it visualized on the webapp. If you allow storing of measured values, you will get another polar to choose from which will update itself. The router paths are still plugin paths as I don't want to clutter official paths yet.

**Comment by tkurki** [comment]
Are we compatible with https://github.com/seandepagnier/weather_routing_pi/tree/master/data/polars ?

It would be cool to have all these polars available and selectable out of the box with the plugin.

**Comment by joabakk** [comment]
Sure can do that. The only difference is the delimiter.

**Comment by joabakk** [comment]
Phew, three different file formats with dialects of csv. I want to test the `combined` branch of signalk-polar before I release it, but it's working on the PC at least. The combined branch is a plugin+webapp

**Comment by tkurki** [comment]
Well you can also issue a PR to homogenize the data.

**Comment by joabakk** [comment]
I have not had much time to test my plugin real time, but as POC it works and the changes in this spec PR are useable for my purpuse and hopwfully others'. I suggest we merge this

File: test/data/vessel-valid/polars.json [hunk 1]
```diff
@@ -0,0 +1,349 @@
+{
+  "uuid": "urn:mrn:signalk:uuid:c0d79334-4e25-4245-8892-54e8ccc8021d",
+  "polars": [
+    {
+      "91c8d96d-4bbd-4c2f-ab36-295b04e83589": {
```
tkurki, line 5: This extra layer does not make sense or I am missing something: either we have objects in the array where all the data, including the uuid is in the object's properties, or have a container object with uuids as property keys.

test/data/vessel-valid/polars.json
```diff
@@ -0,0 +1,349 @@
+{
+  "uuid": "urn:mrn:signalk:uuid:c0d79334-4e25-4245-8892-54e8ccc8021d",
+  "polars": [
```
tkurki, line 3: Should this be under `performance` and not under vessel root?

tkurki (changes requested):
I reread the discussion. With a fresh pair of eyes / brain I think we should go for simpler but more opaque data structure, where the beat and gybe angles and data "rows" are number arrays and you just have to consult the documentation to figure out what they mean.

For an application developer this makes much more sense, so you don't need to access three different arrays. [angle, speed] tuples are essentially polar coordinates, not unlike [lat,lon], and vmg can be thought of as an optional third element as elevation in geo coordinates.

Something like this 
```
{
  "uuid": "urn:mrn:signalk:uuid:c0d79334-4e25-4245-8892-54e8ccc8021d",
  "performance": {
    "polars":
    {
      "91c8d96d-4bbd-4c2f-ab36-295b04e83589": {
        "name": "dynamicPolar 2021-02-26",
        "description": "Generated by polar plugin 2021-02-06",
        "windData": [
          {
            "trueWindSpeed": 8,
            "optimalBeats": [
              //one element if the data is symmetrical, two if tacks are different
              //angle, speed
              [0.29644225714765154, 2.5876562110951293],
              [-0.23247131479853, 2.71626735478773]
            ],
            "optimalGybes": [
              //one element if the data is symmetrical, two if tacks are different
              //angle, speed
              [1.7766377696359377, 2.227545008755847],
              [-2.1683169232599098, 1.8983004809027886]
            ],
            "angleData": [
              //angle, speed, vmg - vmg optional
              [-2.2689280275926293, 1.7696893372101876, -1.0244238433266701],
              [-2.0943951023931966, 2.05263385333391, -0.9281871475493136]
              // .......
            ]
          },
          {
            "trueWindSpeed": 10
            // .......
          }
        ]
      }
    }
  }
}
```

joabakk:
Changed to suit your proposal @tkurki. I'm ok with this, but I have not squashed, just in case. I will have to re-align the POC plugin as well once agreement is made.

joabakk:
Not sure why the tests fail, runs fine locally and on https://travis-ci.org/github/joabakk/specification

joabakk (commented):
All changed since.

joabakk (commented):
All changed since

test/data/vessel-valid/polar-data.json
```diff
@@ -0,0 +1,37 @@
+{
+  "uuid": "urn:mrn:signalk:uuid:c0d79334-4e25-4245-8892-54e8ccc8021d",
+  "performance": {
+    "polars": [
+      {
+        "91c8d96d-4bbd-4c2f-ab36-295b04e83589": {
+          "name": "dynamicPolar",
+          "$description": "Dynamic polar diagram from actual sailing",
```
tkurki, line 8: `$` prefix looks like a typo - or why should this field name be start with $ ?

test/data/vessel-valid/polars.json
tkurki: Imho the array level should be removed so that `polars` is an object. See how it is in my comment https://github.com/SignalK/specification/pull/360#pullrequestreview-584842629

This would allow retrieving individual polars with paths like `.../polars/91c8d96d-4bbd-4c2f-ab36-295b04e83589`.

Or am I missing something?

test/data/vessel-valid/polar-data.json
joabakk: fixed

PR #362, opened 2017-06-16
Add reef and furl state to sails [WIP]

joabakk:
This is an attempt to solve https://github.com/SignalK/specification/issues/188 by allowing both reefable and furlable sails to be status-ed. 

This has come up again with @tkurki 's https://github.com/SignalK/sailsconfiguration where `state` is introduced, but not in schema. Either a sail is reefable (full, 1, 2, ..) or furlable. Other means of reducing sails that I know of such as for gaff sails and square sails are temporary by nature. (correct me if I am wrong).


joabakk:
replaced by #367

PR #365, opened 2017-06-21
Add intakeManifoldTemperature to propulsion schema

anajavi:
Here's a new parameter to propulsion schema.

Is there anything else I need to change for this?


tkurki:
Thanks!

PR #366, opened 2017-06-22
Add CPA and TCPA

joabakk:
Placeholders to calculate for collision avoidance

### DIFF
--- a/test/data/ais-cpa-tcpa.json
+++ b/test/data/ais-cpa-tcpa.json
@@ -8,12 +8,10 @@
           "timestamp": "2014-08-15T18:00:00.755Z",
           "pgn": 130577,
           "$source": "sourceLabel.160"
-        }
-      },
-      "navigation": {
-        "pointOfApproach":  {
-          "closest": 250,
-          "timeToClosest": 45,
+        },
+        "closestApproach":  {
+          "distance": 250,
+          "timeTo": 45,
           "$source": "foo.bar",
           "timestamp": "2014-08-15T19:00:15.404Z"
         }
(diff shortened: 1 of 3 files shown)


tkurki:
I think this should be a single object valued property with properties distance and time plus normal source and timestamp.

Time is UTC, not a number. 

Documentation must include the fact that the values are relative to self.

Should include a test.

joabakk:
Distance and time in one object, isn't that how it is proposed now, except timestamp and source is missing? 
The time to closest approach is not a timestamp, but the time until CPA. That is why number of seconds is proposed here. 
I'll look for a test. Not too many tests to look at for inspiration here...

tkurki:
There are plenty of tests where we have some full data and a test that it validates against the schema.

`timeOfClosest` does not sound like to time *to* the closest approach.

Your structure creates two different values with timestamp & source information, like navigation.speedOverGround and navigation.courseOverGroundTrue for example.

joabakk:
You are right, again

test/data/ais-cpa-tcpa.json
```diff
@@ -0,0 +1,24 @@
...
+      "uuid": "urn:mrn:signalk:uuid:c0d79334-4e25-4245-8892-54e8ccc8021d",
+      "navigation": {
+        "courseOverGroundTrue": {
+          "value": 206.9,
+          "timestamp": "2014-08-15T18:00:00.755Z",
+          "pgn": 130577,
+          "$source": "sourceLabel.160"
+        }
+      },
+      "navigation": {
```
tkurki: Double navigation property - only the latter will end up in the parsed object.

joabakk: Agree. I tried to keep to the CPA and TCPA nomenclature, but this is cleaner and more understandable. Test file fixed

joabakk:
Better now?

tkurki:
I was a bit too hasty, as if you have commonValueFields you don't need source and timestamp. I'll fix those up in #315.

PR #367, opened 2017-06-25
Add reef and furl state to sails

joabakk:
replaces #362

### DIFF
--- a/schemas/groups/sails.json
+++ b/schemas/groups/sails.json
@@ -4,7 +4,6 @@
   "id": "https://signalk.org/specification/1.0.0/schemas/groups/sails.json#",
   "description": "An object describing the vessels sails if the vessel is a sailboat.",
   "title": "sails",
-
   "properties": {
     "inventory": {
       "type": "object",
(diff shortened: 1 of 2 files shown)


joabakk:
#188

schemas/groups/sails.json
```diff
@@ -58,6 +58,32 @@
               "units": "m/s",
               "default": 666
             },
+            "reducedState": {
+              "oneOf": [
+                {
+                  "properties": {
+                    "Reefable": {
```
fabdrol: @joabakk `Reefable` should be `reefable` as we use `camelCase` everywhere.

If I read this correctly, it comes out something like this:
`reducedState.reefable.reefs = 0`
`reducedState.furlable.furledRatio = 0`

Where the "reefable" and "furlable" seems unnecessary and incorrect English. Also, referring back to your comment on #188, it may be nice to have a user-friendly boolean property "reduced". So the model would look something like:

```javascript
reducedState: {
  reduced: true,
  reefs: 1,
  furledRatio: 0
}
```

joabakk:
I have changed as you asked for (I agree), but I don't see the conflict in test/data/sails.json stated in GH gui

fabdrol:
Changes look good, I'm not sure why it complains about merging this either. Trying to figure out using CLI..

bkp7:
I think the problem is line 4 conflicting with Teppo's commit of 1/1/2018. If you Fetch changes from the Master repo that should fix it.

fabdrol:
Thanks @bkp7. @joabakk could you try that?

joabakk:
Aha, there is no test/data/sails.js anymore. I have merged and squashed it down to one commit now

PR #371, opened 2017-07-07
Add server information to discovery

sbender9:
Add details like name, version and capabilities to the discovery information.

### DIFF
--- a/schemas/definitions.json
+++ b/schemas/definitions.json
@@ -506,6 +506,17 @@
       ]
     },
 
+    "serverCapabilityEnum": {
+      "enum": [
+        "stream",
+        "subscription",
+        "auth",
+        "meta",
+        "ratelimit",
+        "command"
+      ]
+    },
+
     "meta": {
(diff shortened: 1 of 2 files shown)


sbender9:
Should we make capabilities an enum or leave it open?

sbender9:
I changed to an enum an started with values from @timmathews comment on slack:

```
"serverCapabilityEnum": {
      "enum": [
        "stream",
        "subscription",
        "auth",
        "meta",
        "ratelimit",
        "command"
      ]
    },
```

tkurki:
I think that `id` and `version` would be ready to merge, as they are pretty well defined server metadata, but `capabilities` is far from ready to merge. Maybe split that off to a separate PR?

sbender9:
Agreed

tkurki:
Could you please update the PR? I think that the server information is ready to be added, just remove the capabilities.

sbender9:
Re-creating because I deleted my branch...

PR #372, opened 2017-07-14
Improve, remove logic error in prop slip description

joabakk:


tkurki:
>actual distance travelled/propeller pitch distance). 0<=ratio<=1, 0 is 0% slip (ideal), 1 is 100% slip

I am confused. If actual distance travelled is 0.09 m and propeller pitch distance is 0.10 then ratio as described _actual distance travelled/propeller pitch distance_ is 0.9, not 0.1 as _0 is 0% slip (ideal)_ would lead me to believe.

joabakk:
You are right. I changed it to 1 - y/x.

PR #377, opened 2017-08-22
Add server information to discovery

sbender9:


tkurki:
The example at http://signalk.org/specification/master/urls_etc.html needs to be updated as well.

tkurki:
Ping

tkurki:
Closed in #390

PR #384, opened 2017-10-14
fix: change aisShipType and atonType

sbender9:
I think this is a much better option for aisShipType and atonType and we should do it before v1.

```
      "aisShipType": {
          "id": 36,
          "name": "Sailing"
        }
```

```
      "atonType": {
        "id": 6,
        "name": "Light, With Sectors"
      }
```


tkurki:
I really don't like enconding numerical id's and human oriented value names into a string field - hey, this is json, let's use objects. So this is definitely a step in the right direction.

However the schema does not enforce or inform what valid values are - you are free to mix and match id and value, if implemented like this.

I took a stab at using object enums here. Object enums work in json schema, but I could not find a way to combine enums for the values and metadata like timestamp and $source in the same object.

But if we push the enum object value one level down, in the `value` property, then we can use enum validation, list possible values in the schema and avoid the confusion related to _what is a value_ a k a _value or not_ .  

See  https://github.com/SignalK/specification/commit/91be402dbf38340a6f5ab0906e994590db8c33bf#diff-5747c9829ac12c3fc4f32be614b4c930 for a start in this direction.

sbender9:
Replaced by #386

PR #386, opened 2017-10-14
fix: better aisShipType and atonType

sbender9:
I think this is a much better option for aisShipType and atonType and we should do it before v1.

```
      "aisShipType": {
          "value": {
            "id": 36,
            "name": "Sailing"
          }
       }
```

```
      "atonType": {
        "value": {
          "id": 6,
          "name": "Light, With Sectors"
        }
      }
```


rob42:
yes.

PR #392, opened 2017-10-28
spec: remove sparse format

tkurki:
I would like to remove sparse format from the specification.

I don't think it is a format on its own like full and delta are: it is just a partially populated full model. The less surface area the spec has the better.

I know that node server has no special support for sparse format, nor am I aware of any client code that relies on data being sparse.


timmathews:
Sounds good to me

PR #404, opened 2017-12-03
feature: remove unsupported i18n

tkurki:
Remove the unsupported, somewhat out of date i18n stuff.
Name and description in schema files and available in
the rest api can be used in stead and once we need
translations we can add proper i18n support, with
actual translations.


timmathews:
Do we need to make any changes to Instrument Panel before this is merged?

tkurki:
Good point.

@signalk/signalk-schema is pinned there at 0.0.1-7, so it won't break, but should be updated.

fabdrol:
@tkurki is this ready to be merged? I see no objection if instrument panel uses a pinned version for now, that can be corrected later.

tkurki:
Did not think of that option. Works for me.

fabdrol:
Okay, I that case let's merge and I'll create an issue on instrument panel referencing this.

fabdrol:
@tkurki github indicated that this merge was fine, but I get e-mails that it's failing.. any idea what that's about?

tkurki:
Maybe take a look at the build yourself?

bkp7:
I've just created PR #466 to fix this problem

ieb:
where did shortName go ? (old UI clients are broken without the same fixes that had to be applied to instrument panel.)

Is there a way of getting these terms without having to refactor the code to poll and cache from the server or should UI clients that don't want to do that, maintain their own mappings for path->shortName ?

AWS, AWA etc are terms that are generally universal from an i18n point of view and used to be part of the schema, admittedly under the i18n subtree, however that may not have been the correct place.

ieb:
Sorry noticed it was closed. I'll open a new issue.

PR #413, opened 2018-01-07
Fixed: bug in messages tests

bkp7:
The validSubscribeMessage and validUnsubscribeMessage were never being called when the tests ran, and so the message tests would always return valid even if there were errors. The actual 2 messages being tested (signalk-subscribe.json and signalk-unsubscribe.json) were in fact OK so no changes are needed to them.


tkurki:
Thanks!

PR #414, opened 2018-01-08
Add the samples to test scripts

bkp7:
Further to #320 samples included in the tests by way of the new samples.js which will automatically includes any files in the samples directory. 

Removed the following scripts which were testing a few of the samples but are now included within 

- samples.js:
- vessel-time.js
- treeValidation.js
- messageValidation.js

The following samples were not schema compliant and have been modified:

- electrical-ac-full.json (this required major modification to comply with the schema)
- signalk-depth-alarm.json
- signalk-depth-meta-attr.json
- 0183-RMC-export-min.json
- 0183-RMC-export.json
- 0183-RMC-full.json
- vessel-time.json

The following samples have been added to ensure the sample testing works properly:

- mob-alarm.json
- MOB-alarm-delta.json
- discovery.json

The signalk-attr.json sample file was removed as it was only a snippet and there is a sample with attributes elsewhere.

Finally, to make it easy for developers and the samples.js code to identify which sample is which, the samples have been split into 5 subdirectories representing the schema they are based on:

- full
- delta
- subscribe
- unsubscribe
- delta

All tests pass and there are no changes to schemas.


samples/full/0183-RMC-export.json
```diff
@@ -1,14 +1,15 @@
-{
+{ "self":"urn:mrn:imo:mmsi:366982330",
   "vessels": {
     "urn:mrn:imo:mmsi:366982330": {
       "mmsi": "230099999",
```
tkurki: Discrepancy, could you please fix it at the same time, even this was already there?

test/samples.js
```diff
@@ -0,0 +1,62 @@
+﻿var chai = require('chai');
+chai.Should();
+chai.use(require('../dist/').chaiModule);
+
+describe('Samples', function() {
+  
+  describe('full', function() {
+	  getFiles('./samples/full').forEach(function(file) {
+			it(file, function() {
```
tkurki: Please indent with 2 spaces

bkp7:
Fixed mmsi and spacing, rebased and checked tests all OK.

PR #419, opened 2018-01-16
fix: modify package.json to make it cross platform compatible.

bkp7:
Added new dependency cross-var which allows npm environment variables to be included in scripts in a cross-platform way.

Substituted 'replace-in-file' in place of 'replace' package which is not cross platform compatible and is flagged as 'Unmaintained' by the author with no activity for over 3 years.

Tested on Windows 10 Version 1709 (build 16299.192) and Ubuntu xenial 16.04.3 LTS

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,12 +22,14 @@
     "ansi-regex": {
       "version": "2.1.1",
       "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-2.1.1.tgz",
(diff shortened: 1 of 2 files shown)


bkp7:
Any thoughts or feedback? Can this be committed?

tkurki:
Could cross-var and replace-in-file be devDepencies, like replace was previously?

bkp7:
You're quite right, sorry. Now moved.

Issue #423, opened 2018-01-18
is value property required?


bkp7:
Where there are multiple sources of the same data we can have values eg:  (valid against vessel schema)
```json
{ "uuid": "urn:mrn:signalk:uuid:c0d79334-4e25-4245-8892-54e8ccc8021d",
  "propulsion": {
    "instance0": {
      "revolutions": {
	"timestamp": "2014-08-15T19:00:15.402Z",
        "$source": "source.a",
        "value": 1280,
        "values": {
	  "source.a": {
	    "value": 1280,
	    "timestamp": "2014-08-15T19:00:15.402Z"
	  },
	  "source.b": {
	    "value": 1281,
	    "timestamp": "2014-08-15T19:00:15.731Z"
}}}}}}
```
The schema currently allows neither, either, or both: value and values. Is this correct?

Should it be valid without either?
Should it be valid with only values?

NB. values are set up consistently in the schema, so the above is equally true for all value types.

Some guidance or a reference to understand whether or not this is correct would be appreciated.

tkurki:
Value is requires. It is the oocation of the _default_ value for that data item. Default is up to server policy, but anyway a safe path for a client to retrieve a value, as there must always be something there.

Values signals the existence of multiple values and is required if there are multiple values. Its existence is not defined for the single value case.

bkp7:
Great, that's what I would have expected. See PR #425

bkp7:
The requirement to have a value has been implemented in PR #425 at least for items utilising commonValueFields, as part of the schema cleanup project.

The issue regarding whether values should be optional for all data points is transferred to issue #427.

Therefore I am closing this issue.

PR #425, opened 2018-01-19
fix: 'value' made a required property for anything using commonValueFields in schema

bkp7:
Ref issue #423.

This changes the schema in a non backwards compatible way, so will bump the version to 1.1.x. Do you want to create a separate branch for changes which will bump the schema version? I will have more to come next week as I'm intending to review the schema once I've completed the testing and document refactoring.

### DIFF
--- a/schemas/definitions.json
+++ b/schemas/definitions.json
@@ -303,7 +303,7 @@
 
     "commonValueFields": {
       "type": "object",
-      "required": ["timestamp", "$source"],
+      "required": ["timestamp", "$source", "value"],
       "properties": {
         "timestamp": {
           "$ref": "#/definitions/timestamp"


rob42:
#427

fabdrol:
@bkp7 @tkurki @rob42 how should we handle this - as it is a breaking change? My feeling is: 

- Use master as the development line for any non-breaking work, i.e. on the current major and minor version level
- Set up a new branch (development?) in which we can merge breaking stuff like this (or perhaps a branch name like 1.1.x - although that may backfire if for whatever reason we want to delay this until 1.2.x). 

Only downside may be that such a branch needs rebasing on master every now and then. Thoughts?

bkp7:
I have been working on changes required to lose the TV4 dependency, and it would be very time consuming to do that work and retain bugs like this and #427 (with a view to subsequently fixing the bugs). It will be much easier to fix the bugs as we work through the changes required for TV4. There are some areas where there are a fair number of anomalies.

I propose a branch 'schema clean up' or similar with the purpose of dealing with these issues and the TV4 dependency. From my work so far I believe the clean up may require going to version 2. I am happy to get this work done, but it would be easier to have a specific branch for the purpose so everyone can comment/contribute if required.

fabdrol:
@bkp7 that makes sense, I'll set it up.

Issue #406, opened 2018-01-07
metadata: overall gauge limits are not included in metadata


bkp7:
Whilst zones are defined in the meta data the upper and lower limits of the gauge itself are not defined. Existing implementations have to make assumptions about the gauge limits eg. some implementations assume 10 knots as the maximum for STW which is not useful for faster vessels.

Possible solution:
Add a lower and upper gauge limit to the meta properties. I considered whether we could use a whole of gauge zone, but since that would be a zone with unique properties (ie. both upper and lower limits required) I have discounted it. Also there is no way that the gauge 'zone' could be considered an alarm zone. One benefit of specifically defining gauge limits as opposed to trying to infer gauge limits from warning zone limits, is that alarm conditions can be independent of the gauge. For example a typical coolant temperature gauge might display from 40 to 120 Celcius with warning zones at the higher temps. It might be useful to have a warning zone if for example the coolant temperature were below freezing in order to warn against engine start, even though that warning zone would not be depicted on the gauge.

tkurki:
Yes, how about adding something lilke
```
"displayScale": {
  "lower": 10
  "upper": 15
}
```

bkp7:
Or:
```
"gaugeLower": 10,
"gaugeUpper": 15
```
or:
```
"gaugeScale": {
  "lower": 10,
  "upper": 15
}
```
At first I didn't like yours because 'scale' meant to me some sort of sizing. However on reflection I think it may be better not least because it could also include whether or not the scale on the gauge should be logarithmic. eg. depth, luminosity (I know it's a separate point and I've not raised an issue for it yet).

I'm not keen on display though. To me one display or instrument panel can have multiple gauges, see similar usage in subscription_protocol.md (line 90-112).

If we can decide on the terminology I will implement it.

tkurki:
I think we have used the word display elsewhere in the schema. To me it is more related to displaying the value”, more a verb than a noun.

bkp7:
You're quite right, 'displayName' is used already in metadata in the context of one gauge and could be interpreted as either a verb or noun.

In that context 'displayScale' sits alongside nicely, so unless these are any other thoughts I'll put it in as you suggested.

In passing 'display' is also used in the context of units definitions (eg "display": "Hz") which I don't think bears on this at all.

PR #436, opened 2018-02-20
Add displayScale lower and upper limits to metadata

bkp7:
Resolves issue #406.

Schema modified so that 'displayScale' is optional (all existing tests pass unaltered). For versioning this constitutes an ADDITION so bumps the schema by x.x.+1.

If displayScale is present it must have both 'lower' and 'upper' as having just one is pointless:
- Tests added:
    - vessel-valid/meta-with displayScale.json
    - vessel-invalid/meta-with displayScale lower only.json
    - vessel-invalid/meta-with displayScale upper only.json

Both Upper and Lower must be numeric:
- Tests added:
    - vessel-invalid/meta-with displayScale non numeric lower.json
    - vessel-invalid/meta-with displayScale non numeric upper.json

No other properties allowed within displayScale
- Test added:
    -  vessel-invalid/meta-with displayScale and additional properties.json


tkurki:
>If displayScale is present it must have both 'lower' and 'upper' as having just one is pointless:

I stopped to think about this a bit. Think about a graphical scrolling echosounder display: it is effectively an autoranging gauge that has "lower" set to 0 and it scales upper to a "nice" value that contains all samples in the time window.

Should we relax this so that either is mandatory, if the property exists?

bkp7:
OK, yes and the autoranging, time series or sparkline type display could be used for all sorts of data.

However, the metadata/displayScale is applicable to a single discrete value, and a gauge (dial) cannot exist without having an upper limit. I am still minded that if present there should be both a lower and upper value set and sent by the server.

Perhaps we need to find a form of words which states that the consumer display must be able to show values within the range specified by lower and upper displayScale. Autoranging would meet that criteria. That looser wording would also allow a gauge to show a larger display scale. eg. a tacho with 0-4200 lower-upper might look better in some designs if the gauge went to 4500 or 5000. This would be a decision for the gauge designer.

tkurki:
I think we are thinking too far ahead here and too much about a single detail / use case. 

> states that the consumer display must be able to show values within the range specified by lower and upper displayScale

The schema should have well known "homes" for different data items that are needed to support the real world use cases. Here the case is "we need a place to handle the min and max values for displaying a gauge". I think we are going in the wrong direction if we try to specify exactly how the user interface must behave.

bkp7:
Agreed with the fundamental issue: "we need a place to handle the min and max values for displaying a gauge".

Apart from a few gauge designs, both values are required to render a gauge, that is why I propose that both must be present in the data (if either is). That doesn't mean they both have to be used by the consumer. If you allow only one value to be present you are making assumptions at the sensor/server end about what type of gauge will be rendered by the consumer.

However, if the view is that it should be changed to allow either I'll make the changes so we can put this one to bed.

tkurki:
We are arguing a pretty fine point here, I'm good with both being required.

I'd prefer no spaces in filenames.

I read through the titles and descriptions and thought about my argumentation above. I'd prefer wording that talks less about a gauge and more about the display preferences, be it a dial gauge or for example an axis of a graph. But we can move forward with this as well.

bkp7:
Spaces removed from filenames. We have previously added a test for spaces [here](https://github.com/SignalK/specification/blob/7b1338dcc9fb2940992ec13b4927f5fef74c5685/test/jsonSamplesAndTestFiles.js#L10) but this PR predated the addition of that test.

Reworded to refer to display rather than gauge and fixed typo.

Issue #444, opened 2018-03-01
bug: environment inside zones not extensible


bkp7:
The only Inside Zones currently allowed within environment are: engineRoom, mainCabin, refrigerator, freezer and heating. There is no way to include eg. portEngineRoom, aftCabin, galley, etc.

The schema needs to be changed to allow patternProperties as with propulsion, etc.

PR #445, opened 2018-03-01
Fix Environment Schema and add tests

bkp7:
Fixes issue #444 

Removes hard coded inside zones eg. mainCabin, engineRoom and replaces with user definable to allow eg. portEngineRoom, aftCabin, etc.

All existing tests pass unchanged. Added 1 valid and 2 invalid tests.


bkp7:
Any reasons this can't be merged?

PR #447, opened 2018-03-01
feature: add heatIndexTemperature to environment inside zones

### LINKED ISSUE #446: enhancement: add heatIndexTemperature to inside environment
heatIndexTemperature is available for the outside 'zone' but not inside zones and should be added to the environment schema.

bkp7:
fixes #446.

Adds heatIndexTemperature to inside zones to be consistent with outside.

Also added a valid and invalid test.

### DIFF
--- a/schemas/groups/environment.json
+++ b/schemas/groups/environment.json
@@ -13,6 +13,11 @@
           "$ref": "../definitions.json#/definitions/numberValue",
           "units": "K"
         },
+        "heatIndexTemperature": {
+          "description": "Current heat index temperature in zone",
+          "$ref": "../definitions.json#/definitions/numberValue",
+          "units": "K"
+        },
         "pressure": {
           "description": "Pressure in zone",
           "units": "Pa",
--- a/test/data/vessel-invalid/environment-with_inside_heatIndex.json
+++ b/test/data/vessel-invalid/environment-with_inside_heatIndex.json
@@ -1,0 +1,24 @@
+{
+  "uuid": "urn:mrn:signalk:uuid:c0d79334-4e25-4245-8892-54e8ccc8021d",
+  "environment": {
+    "inside": {
+      "mainCabin": {
+        "temperature": {
+          "value": 288.1,
+          "timestamp": "2018-02-28T09:46:55.302Z",
+          "$source": "foo.bar"
+        },
+        "relativeHumidity": {
+          "value": 0.86,
+          "timestamp": "2018-02-28T09:46:55.302Z",
+          "$source": "foo.bar"
+        },
+        "heatIndex": {
+          "value": 289.2,
+          "timestamp": "2018-02-28T09:46:55.5102Z",
+          "$source": "foo.bar"
+        }
+      }
+    }
+  }
+}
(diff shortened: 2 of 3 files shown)


schemas/groups/environment.json
```diff
@@ -13,6 +13,11 @@
           "$ref": "../definitions.json#/definitions/numberValue",
           "units": "K"
         },
+        "heatIndexTemperature": {
+          "description": "Current outside heat index temperature",
```
tkurki: Outside?

test/data/vessel-valid/environment-with inside heatIndexTemperature.json
```diff
@@ -0,0 +1,24 @@
+{
```
tkurki: I just noticed that the earlier test data reorganisation added lots of files with spaces in their names. They are sort of a pain to work with on the command line, adding cruft to how they look, and I would be much happier with the spaces replaced with underscores, here and elsewhere.

bkp7: Fair call. See PR #449 which replaces all the spaces with underscores

bkp7:
@tkurki sorry, by flattening the additional changes you requested, I seem to have made your comments on this PR disappear.

A new PR #449 removes spaces from all existing filenames and also modifies the tests to ensure there's no recurrence.

I have modified the description wording in environment.json, and changed the filenames on this PR to remove the spaces.

tkurki:
Github shows them ok to me, as _outdated diffs_ that can be expanded.

I don't think the invalid example adds any value here: somebody might stumble on that and treat as an example, not an invalid one. Unless there is something I am missing?

BTW  my pet peeve is commit messages (and PR titles) in the imperative mood per https://chris.beams.io/posts/git-commit/#imperative. 

And while not a rule I have found [Angular commit message conventions'](https://gist.github.com/stephenparish/9941e89d80e2bc58a153#allowed-type) <type> helpful most of the time.

bkp7:
The invalid test is there for 2 purposes:
1. to prove that the schema fails when an invalid property is present within a zone
2. to ensure that heatIndex property explicitly fails as I felt this was actually a more intuitive name than heatIndexTemperature and it should be a test to ensure that the more intuitive name doesn't get put in in future, and also if someone does look through invalid tests they see that heatIndex is invalid by design.

I agree the second part of the test is of secondary importance to the first which ensures the schema is doing its job of failing invalid json. I can rename the test to make it clearer it's meant to fail or would you prefer I remove the secondary part of the test and change the property to an obviously spurious name?

Re the PR and commit naming, I'm certainly with you on keeping things consistent and standardised. I had looked at previous commit messages and it looked like everyone was doing their own thing.

I have added on new issue #450 to document some guidelines.

bkp7:
@timmathews, could I just clarify what the 'After v1' milestone is for please.

v1.0.0 was released on 1 Jan 2018, the current release is 1.0.3 released 8 Jan 2018.

If there's a freeze on new features whilst the bugs in v1 are ironed out that's great. It would be useful to be able to tag those bugs so we can get them cleared up.

Thanks

timmathews:
@bkp7, "After v1" is essentially a feature freeze pending the official Signal K v1.0.0" announcement. Should be soon ....

Issue #298, opened 2016-11-29
Replace N2K src with CAN NAME


jboynes:
Something has been bugging me with the way we handle N2K sources based on the 'src' address field. I'm not privy to the spec so may be off base here, but digging into the N2K data stream from my boat I see devices, primarily my Raymarine autopilot, claim different addresses over time.

I see references to PGN 060928 which [is publicly documented](http://www.nmea.org/Assets/20140710%20nmea-2000-060928%20iso%20address%20claim%20pgn%20corrigendum.pdf) as an address claim per ISO 11783-5 which associates the address with the actual NAME of the device (i.e. its 64 bit unique id).

From what I'm seeing on my network this address appears ephemeral and can vary each time the N2K CAN network boots. Now it seems to settle into a repeatable state but I think that's just the result of the devices I have connected and not any address reservation scheme so may change if I add/remove a device. NMEA hints at that with a [TC](http://www.nmea.org/Assets/20160209%20nmea%202000%20te%20200020160208%20address%20to%20name%20association%20errata.pdf) that states each device is required to build a table that "provides the means to associate a parameter group with a specific NMEA 2000 certified device even though the device’s address may change at any time."

It seems to me that our use of `src` to contain the CAN source address should be replaced with use of the NAME that has claimed that address at the time the message was sent. If someone more familiar with the requirements can verify this behaviour I will put together a PR.

tkurki:
I believe this is something that should be done implementation first, so that we don't end up with a spec that is unimplementable or with no open source implementation.

Some provision should be made for playback of n2k capture files, where no bus interrogation is possible.

pod909:
An investigation into the NAME(s) of a device connected to 2 busses might be wise.

sumps:
I'm happy with the implementation as it is now.

jboynes:
I have [a capture](https://github.com/SignalK/specification/files/619556/actisense.txt) of my boat's startup taken by actisense-serial from an Actisense NGT-1. Running it through canboat analyzer, these are the 60928 PGNs I see:
```
2016-10-08T21:04:45.078Z 6   5 255  60928 ISO Address Claim:  Unique Number = 0x9000e; Manufacturer Code = Raymarine; Device Instance Lower = 0; Device Instance Upper = 0; Device Function = 190; Device Class = Instrumentation/general systems; System Instance = 0; Industry Group = Marine
[...]

PR #456, opened 2018-03-23
[WIP] Add `name` property to `source` object

timmathews:
This adds a `name` property to the `source` object so that each NMEA 2000 (or other CAN bus protocol built on J1939) source has a consistent, static, and unique identifier even if the network is reconfigured or in the event of address contention.

See SignalK/n2k-signalk#16 and https://github.com/SignalK/signalk-server-node/tree/canname for more background.


tkurki:
The original specification issue is #298.

fabdrol:
@timmathews is this still WIP?

fabdrol:
@timmathews I'm closing this for lack of activity. Please re-submit a PR when you have time again!

PR #461, opened 2018-05-03
feature: add type to meta/displayScale

### LINKED ISSUE #418: metadata: no way to describe the type of scale for a gauge (logarithmic/linear)
Some values really need to be shown on a logarithmic rather than linear scale eg. luminosity, depth, signal strength. There are also some which work better on a squareroot scale eg. depth, windspeed.

Suggest within meta we add:
```
"gaugeScale": {
  "lower": 10,
  "upper": 15,
  "type": "linear"
}
```
where type can be linear, logarithmic, squareroot or power with a default of linear if not present. See also #406 where this layout is suggested for gauge limits but not implemented at present. If power is specified then an additional property must be present to define the power:

```
"gaugeScale": {
  "lower": 10,
  "upper": 15,
  "type": "power",
  "power": 2
}
```
Note that a power of 0.5 is equivalent to squareroot, and a power of 1 is equivalent to linear.

In using these scales the type just defines the function which is applied to all values in order to calculate % deflection of the pointer/needle.

For linear:          % deflection = (V - L) / (U - L)
For logarithmic: % deflection = (log(V) - log(L)) / (log(U) - log(L))
For squareroot: % deflection = (√V - √L) / (√U - √L) 
For power:         % deflection = (Vᴾ - Lᴾ) / (Uᴾ - Lᴾ)
Where: V = value, L = lower bound of the gauge, U = upper bound of the gauge and P = power

Note that on a logarithmic scale neither L nor U can be zero.

Are there any others?

bkp7:
Add `type` within `displayScale` object to allow for non linear scales, ie. logarithmic, squareroot and power.

Tests added, sample extended and documentation updated.

closes #418

### DIFF
--- a/schemas/definitions.json
+++ b/schemas/definitions.json
@@ -569,7 +569,6 @@
           "type": "object",
           "title": "Scale to display.",
           "description": "Gives details of the display scale against which the value should be displayed",
-          "required": ["lower", "upper"],
           "properties": {
             "lower": {
               "type": "number",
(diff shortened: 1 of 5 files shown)


schemas/definitions.json
```diff
@@ -582,9 +581,26 @@
...
+            },
+            "power": {
+              "type": "number",
+              "title": "Selected power for display scale",
+              "description": "The power to use when the displayScale/type is set to 'power'. Can be any numeric value except zero.",
+              "example": 2
             }
           },
-          "additionalProperties": false
+          "oneOf": [
```
tkurki: Please format the json also here.

rob42:
Vote to merge

bkp7:
@fabdrol, yes, this is good to merge.

PR #462, opened 2018-05-03
fix: modify schema to enforce `description` and `units`

### LINKED ISSUE #439: clarification: are 'units' and 'description' required properties in 'metadata'
The Documentation refers to [Implicit Metadata](http://signalk.org/specification/1.0.0/doc/data_model_metadata.html#implicit-metadata). Does this mean that these 2 properties are always required within metadata? If not always, under what circumstances are they required?

bkp7:
Schema modified to enforce the written specification that description and units must be present in meta.

Also added possibility for meta to include an enum list of possible values where the value is non numeric. This is not mandatory as at 1.0.x but the documentation notes that it is likely to become mandatory in future (see issue #439).

Samples and tests corrected to comply with the requirement.

Additional tests added to test the schema modifications.

Although the schema is not backwards compatible (making `description` mandatory where it wasn't previously), this change is just placing into the schema a requirement that was previously written in the specification. As such it can be viewed as an ADDITION (a schema change that is compatible with all historical data).

closes #439

### DIFF
--- a/schemas/definitions.json
+++ b/schemas/definitions.json
@@ -346,6 +346,9 @@
                 "$ref": "#/definitions/valuesNumberValue"
               }
             }
+          },
+          "meta": {
+            "required": ["units"]
           }
         }
       }]
@@ -529,6 +532,7 @@
       "type": "object",
       "title": "Meta schema.",
       "description": "Provides meta data to enable alarm and display configuration.",
+      "required": ["description"],
       "properties": {
         "displayName": {
           "type": "string",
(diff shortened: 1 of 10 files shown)


schemas/definitions.json
```diff
@@ -346,6 +346,9 @@
                 "$ref": "#/definitions/valuesNumberValue"
               }
             }
+          },
+          "meta": {
```
tkurki: I don't understand how this would work - a property named `meta` with nothing but a `required` field - this does not reference the meta definition at all?

tkurki: Oh never mind, I think I got it: commonValueFields pulls in the meta def.

PR #463, opened 2018-05-03
feature: add 'nominal' state to meta zones

### LINKED ISSUE #407: metadata: gauge zone overlap
There is no explanation as to what should happen if a value is in 2 zones simultaneously. Which should take precedence? We cannot rely on the order they appear in the json array. Also there is no explanation or requirement that zones are contiguous. What is the position where the value is not within any zone? The examples given in the documentation all have overlapping zones because the upper and lower limits are by schema definition within the zone.

Possible solution:
Either force every zone to have both an upper and lower limit and make sure they are contiguous, or explain that zones should be considered by their level of severity with higher severity overriding lower ones. My preference is the second option as it requires no changes to the schema or usage. Also it would seem sensible to explain that anything not within a zone is considered normal, therefore any normal zones supplied would have no effect.

### LINKED ISSUE #411: metadata: no provision for gauge null zones
Some gauge styles have 'null zones', typically near the centre of the normal band within which the indicator doesn't move at all. It enables gauges to be designed so that all needles even across different measurements be aligned during normal operations making identification of any trends away from normal very easy to spot. This is not currently supported in SignalK at all.

Possible Solution: Add an additional zone type of 'dead' or 'null' which can be used to indicate such a zone if relevant to the gauge design in question. An alternative would be to use 'normal', which assuming the proposed solution to issue #407 were adopted, would otherwise have no meaning (because anything which was not specifically a warning is now considered 'normal'). The only issue then would be backward compatibility.

bkp7:
Add `nominal` state to zones within meta and update documentation accordingly.

closes #411 and resolves #407


gitbook-docs/data_model_metadata.md
```diff
@@ -44,18 +44,50 @@ In the example `meta` object above, a definition is provided for an analog RPM g
...
+
+The `upper` and `lower` values in the zones do not need to be contiguous, they don't have to both be present in a zone, nor do
+they need to be within the bounds of the `upper` and `lower` specified in `displayScale`. When they are outside of the
+`displayScale` range they will still give rise to alerts. Both `upper` and `lower` values are considered to be inclusive.
+
+If zones overlap each other the state/zone with the highest severity will take precedence. This is true for both alerts and
+gauge/display rendering. Any part of the range which is not explicitly within a zone is considered to be `normal` (the default).
+As such zones with a state of `normal` have no effect and their remove would result in no changes to either displays or alerts.
+
+The can be multiple zones with the same `state`, for example if a different message is required, or if they are on different parts of the scale.
```
tkurki: Should be "There"

gitbook-docs/data_model_metadata.md
```diff
@@ -44,18 +44,50 @@ In the example `meta` object above, a definition is provided for an analog RPM g
...
+###`zones`
 The last property in the `meta` object is the `zones` array. This provides a series of hints to the consumer which can
-be used to properly set a range on a display gauge and also color sectors of the gauge to indicate nominal or dangerous
+be used to properly set a range on a display gauge and also color sectors of a gauge to indicate normal or dangerous
 operating conditions. It also tells the consumer which state the data is in for a given range. Combined with the alert
 method properties, all Signal K consumers can react the same way to a given state.
 
-It is also possible for a Signal K server to use this information to monitor any data which has a `meta` object and
+The possible states in ascending order of severity are:
+
```
tkurki: Nice table & order, but I am not too crazy about going so far into specifying UI specifics as to state colors that should be used. My vote is on removing the third column. There are so many ways you can make readings more or less prominent (color, size, blinking, movement, outlines,  opacity to list a few off the top of my head) that I feel we should leave all of that to the UI designer and implementor.

Capitalize the heading texts?

tkurki:
This looks good to me, should I merge?

rob42:
Good to merge.

fabdrol:
@bkp7 could you look at the conflict? Then we can merge!

bkp7:
@fabdrol conflicts sorted. This is now good to merge.

PR #469, opened 2018-06-08
chore: mark gaugeType as deprecated, update documentation and samples

### LINKED ISSUE #409: metadata: gaugeType not well defined
There is a suggested "gaugeType" but no indication as to possible/permissible values. The only 2 examples I can find in the docs are analog or sparkline. Is it not expected that other gaugeTypes be implemented? It is described in the schema as "the gauge necessary to display this data", but the documentation describes it as the "recommended display format".

Solution: Make the documentation consistent with the schema description. Also consider whether it should be an enum, or there should be a suggested list. Without some guidance the gaugeType will only work where server and client are from the same manufacturer.

bkp7:
resolves #409


fabdrol:
Please see my comments at #409

PR #470, opened 2018-06-13
chore: remove deprecated gaugeType

### LINKED ISSUE #409: metadata: gaugeType not well defined

bkp7:
Remove gauge type from meta.

closes #409

### DIFF
--- a/gitbook-docs/data_model_metadata.md
+++ b/gitbook-docs/data_model_metadata.md
@@ -45,6 +45,8 @@ In the example `meta` object above, a definition is provided for an analog RPM g
 few different options for the consumer to use to display the name of the measurement and explicitly calls out the unit
 of measure. It also specifies a recommended display format via `gaugeType`.
 
+###`timeout`
+
 The `timeout` property tells the consumer how long it should consider the value valid. This value is specified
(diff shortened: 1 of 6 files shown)


fabdrol:
@bkp7 we should probably keep an eye on `schema-cleanup`, and make sure that it gets rebased on master every now and again. Here's what I propose we do:

- Merge the removal of `gaugeType` into schema cleanup
- Close the other PR and the issue
- Create a [WIP] issue for `schema-cleanup`, that it needs to be merged onto master at some point

Since cleanup is one of those tasks that one can do forever, let's make a todo list of things that really badly need doing. That'll be the scope for the `schema-cleanup` branch and once that list is done, we rebase it on master and create a PR for the branch. Further cleanup can be done in a new branch.

Q: do you have a list of things you feel should be in Schema Cleanup?

PR #475, opened 2018-06-23
docs: clarify websockets hello message

### LINKED ISSUE #473: Connection Hello response for streaming is not valid signalk json
The specification requires a hello message to be sent from the server upon successful subscription to streaming. The example in the [documentation](https://github.com/SignalK/specification/blob/master/gitbook-docs/streaming_api.md#connection-hello) being:
```json
{
  "version": "1.1.2",
  "timestamp": "2015-04-13T01:13:50.524Z",
  "self": "123456789"
}
```

This is neither a valid full nor delta signalk json message.

- Is this supposed to be a totally separate format to the deltas otherwise received via streaming?
- Should it be here, or is it needed at all? I assume the version is required because deltas don't include the version. Would it be better to include the version in every delta?

Assuming it should remain and the production of a new schema is required for this:
- Is `self` mandatory? Doesn't seem relevant to an off vessel upstream server.
- Is `timestamp` mandatory? Does every server have to have a reliable time source?

bkp7:
Documentation improved.

Add new json schema file to be used to validate the hello message.

Add sample and test json files to illustrate and test usage.

resolves #473 

@tkurki be grateful if you could take a look at this. In particular there appears to be some commonality between the websocket hello message, the rest api, and the streaming api. None of which have to be implemented but I'm assuming if more than one is implemented the values (server name, signalK version, etc) should be the same in each place. I've assumed this and added to the documentation accordingly.

### DIFF
--- a/schemas/hello.json
+++ b/schemas/hello.json
@@ -1,0 +1,60 @@
+{
+  "type": "object",
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "https://signalk.org/specification/1.0.0/schemas/delta.json#",
+  "title": "SignalK Websockets Hello message schema",
+  "description": "Schema for defining the hello message passed from the server to a client following succesful websocket connection",
+  "required": [
+    "version",
+    "roles"
+  ],
+  "properties": {
+    "version": {
+      "description": "Version of the schema and APIs that this data is using in canonical format i.e. 1.0.0.",
+      "$ref": "definitions.json#/definitions/version"
+    },
+    "name": {
+      "type": "string",
+      "description": "The name of the Signal K server software",
+      "example": "iKommunicate"
+    },
+    "timestamp": {
+      "$ref": "./definitions.json#/definitions/timestamp"
+    },
+    "self": {
+      "type": "string",
+      "description": "This holds the context (prefix + UUID, MMSI or URL in dot notation) of the server's self object.",
+      "example": "vessels.urn:mrn:signalk:uuid:6b0e776f-811a-4b35-980e-b93405371bc5",
+      "oneOf": [
+        {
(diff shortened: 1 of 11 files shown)


schemas/hello.json
```diff
@@ -0,0 +1,60 @@
+{
+  "type": "object",
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "https://signalk.org/specification/1.0.0/schemas/delta.json#",
```
fabdrol: I think this `id` needs to change

schemas/hello.json
```diff
@@ -0,0 +1,60 @@
...
+      "oneOf": [
+        {
+          "pattern": "^vessels.(urn:mrn:(imo:mmsi:[2-7][0-9]{8}$|signalk:uuid:[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-4[0-9A-Fa-f]{3}-[89ABab][0-9A-Fa-f]{3}-[0-9A-Fa-f]{12}$))|(http(s?):.*|mailto:.*|tel:(\\+?)[0-9]{4,})$"
+        },
+        {
+          "pattern": "^aircraft.(urn:mrn:(imo:mmsi:1[0-9]{8}$|signalk:uuid:[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-4[0-9A-Fa-f]{3}-[89ABab][0-9A-Fa-f]{3}-[0-9A-Fa-f]{12}$))|(http(s?):.*|mailto:.*|tel:(\\+?)[0-9]{4,})$"
+        },
+        {
+          "pattern": "^aton.(urn:mrn:(imo:mmsi:99[0-9]{7}$|signalk:uuid:[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-4[0-9A-Fa-f]{3}-[89ABab][0-9A-Fa-f]{3}-[0-9A-Fa-f]{12}$))|(http(s?):.*|mailto:.*|tel:(\\+?)[0-9]{4,})$"
+        },{
```
fabdrol: Please check formatting

PR #476, opened 2018-06-23
fix: discrepancy between docs and sample json

bkp7:
Documentation discrepancy between the meta documentation and the sample file populating it. Also minor fix in the sample to ensure meta/description agrees with the value in the schema. This was introduced in PR #461 by me, sorry.


fabdrol:
Looks fine to me, anyone else care to comment? Otherwise I'll merge.

PR #492, opened 2018-07-21
fix: remove additionalProperties from meta deltas

bkp7:


fabdrol:
@bkp7 how does this relate to #454? should it be merged?

fabdrol:
Ping @bkp7? I'll close if not active

PR #493, opened 2018-07-21
style: modify and reformat json test files

bkp7:


fabdrol:
@bkp7 can we merge this - as it's only test files?

PR #494, opened 2018-07-23
docs: clarify discovery schema and usage

### LINKED ISSUE #482: No way to permit secure interfaces other than by redirect
Currently the only permitted endpoints are:
- signalk-http
- signalk-ws
- signalk-tcp

Should these be permitted:
- signalk-https
- signalk-wss
- signalk-udp

Any others ?

Redirect doesn't seem like a great idea, it effectively means assuming all consumers support SSL/TLS

### LINKED ISSUE #485: Ambiguous documentation for REST discovery
The discovery.json schema and supporting [documentation](http://signalk.org/specification/1.0.4/doc/rest_api.html) are ambiguous as regards the following questions:

Q: is the format of group of endpoints eg. `"v1"` fixed or is it freetext?
Q: is the path to the service fixed `.../signalk/v1/api/` ie. are `api` and `stream` mandated and is the use of the `v1` in a fixed format?
Q: how does a server present both IPv4 and IPv6 interfaces?

In light of the above is the following allowed/expected? (it does validate against the schema). If not how should these services be advertised?

```
{
  "endpoints": {
    "v1-alpha": {
      "version": "1.1.0-alpha1",
      "signalk-http": "http://192.168.1.5/signalk/alpha/rest/",
      "signalk-http": "http://[21DA:D3:0:2F3B:2AA:FF:FE28:9C5A]/signalk/alpha/rest/",
      ...
    },
    "v1": {
      "version": "1.0.4",
      "signalk-http": "http://192.168.1.5/signalk/stable/rest/",
      ...
    }
  }
}
```

bkp7:
Improve documentation:
- clarify that https can be used within 'signalk-http'
- clarify that current spec only covers 'v1' within discovery schema
- clarify that endpoint locations are fixed in v1
- clarify how to deal with IPv6 endpoints
- clarify that insecure and secure endpoints cannot co-exist on a v1 server (ie. you can't have both http and https unless one is on IPv4 and the other on IPv6

closes #482 and closes #485

### DIFF
--- a/gitbook-docs/rest_api.md
+++ b/gitbook-docs/rest_api.md
@@ -9,21 +9,17 @@ configuration changes. As specified in the previous section, all URLs for intera
 Making a `GET` request to `/signalk` returns a JSON object which specifies the available Signal K endpoints and some
 information about the server. Also see [versioning](versioning.md) for details about `version` strings.
 
-[>]: # (mdpInsert ```json fsnip ../samples/discovery/docs-rest_api.json)
+[>]: # (mdpInsert ```json fsnip ../samples/discovery/docs-rest_api.json --ellipsify v3)
 ```json
 {
   "endpoints": {
     "v1": {
       "version": "1.0.0-alpha1",
-      "signalk-http": "http://localhost:3000/signalk/v1/api/",
-      "signalk-ws": "ws://localhost:3000/signalk/v1/stream"
-    },
-    "v3": {
-      "version": "3.0.0",
-      "signalk-http": "http://localhost/signalk/v3/api/",
-      "signalk-ws": "ws://localhost/signalk/v3/stream",
+      "signalk-http": "https://localhost:3000/signalk/v1/api/",
+      "signalk-ws": "ws://localhost:3000/signalk/v1/stream",
       "signalk-tcp": "tcp://localhost:8367"
-    }
+    },
+    "v3": {...}
   },
   "server": {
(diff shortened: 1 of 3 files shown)


gitbook-docs/rest_api.md
```diff
@@ -48,15 +44,24 @@ of the specification and there is no registry for id values. If providfed, the `
 as `swname` and `swvers` within the [DNS-SD advertisement](connection.md) (if implemented), and also the `id` MUST
 provide the same value as `name` within the [Websocket hello message](streaming_api.md) (if implemented).
 
-## /signalk/«version»/api/
+# v1 specification
+
+If a REST API is provided it MUST be located at `http[s]://<<host>>:[port]/signalk/v1/api/`
+If a streaming websocket is provided it MUST be located at `ws[s]://<<host>>:[port]/signalk/v1/stream`
+
+The discovery json document published by an IPv4 address should advertise IPv4 endpoints and similarly the discovery json at an IPv6 address should advertise IPv6 endpoints.
```
tkurki: I don't think this should be stated in the spec. If a server wants it can advertise IPv6 over IPv4 - dunno how much sense it makes, but nevertheless I don't think the spec should make restrictions here.

fabdrol:
@bkp7 this still needs resolving, could you look at @tkurki's comment?

fabdrol:
@bkp7 ping?

PR #497, opened 2018-08-12
doc: add MAY re: http/2

tkurki:
Add a mention of optional support for http/2.


gitbook-docs/rest_api.md
```diff
@@ -4,6 +4,8 @@ Signal K producers MAY implement an HTTP API which consumers can use to self-con
 configuration changes. As specified in the previous section, all URLs for interacting with Signal K are rooted at
 `/signalk`.
 
+A Signal K server implementing the HTTP API MAY support http/2.
```
JonRowe: Prehaps italics or bold would be better for `MAY` given that `HTTP API` is an acronym

tkurki: If we do that here then we should do it all over the spec. I'll just remove the capitalization.

Issue #437, opened 2018-02-20
tests: there is no check in place for schema validity.


bkp7:
There are currently no automated tests in place for schema validity and these should be added to prevent recurrence of invalid schemas (issues #434 and #320).

The current json validator in use (TV4) does not check for schema validity as part of its tests.

rob42:
Needs http access to the different versions of the schema (v1.0.0 differs from v1.0.4 ), currently I only see v1.0.0.
I'm very short on time, so I dont want to start on this if there is another solution underway, should I go ahead?

webmasterkai:
https://ajv.js.org might be a useful tool in validating.

bkp7:
@rob42, sorry perhaps my post was slightly misleading. There is no problem in getting the existing TV4 validator to validate the schema, it just that such a test hasn't been implemented yet.

All other json validation tools are useless to us at present as none support the TV4 'banUnknownProperties' flag which our schemas require in order to work properly (see issue #246).

We already have an online validator tool at [ http://node-master.signalk.org/@signalk/playground/dist/ ]( http://node-master.signalk.org/@signalk/playground/dist/ ) which uses TV4.

The current proposal is to update the existing schema to be draft-04 compliant thus removing the dependency on TV4. At that point any json validator can be used for validation and the proposal is to move to AJV. This will also give us coverage stats for our tests.

rob42:
ah - ok. I'll leave it then

PR #498, opened 2018-08-15
chore: remove dependency on tv4-formats

sarfata:
TV4-Format provides some JSON validation helper for common string
formats. It was loaded only when validating the full model (not delta)
and all the tests run without it so I propose we remove it.

Reason for me looking into this is that this library causes problem with
webpack because of a heuristic they use to require one of their
dependency. This is a known issue that they have fixed in a later
version and I have confirmed that upgrading the package fixes the issue
for us too. However, if we do not use the library I think it's better to
completely remove it.

For more info: https://github.com/ikr/tv4-formats/commit/61450dc7499440980169a67bc51d554a67c5c30d

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3937,7 +3937,8 @@
     "moment": {
       "version": "2.20.1",
       "resolved": "https://registry.npmjs.org/moment/-/moment-2.20.1.tgz",
-      "integrity": "sha512-Yh9y73JRljxW5QxN08Fner68eFLxM5ynNOAw2LbIB1YAGeQzZT8QFSUvkAz609Zf+IHhhaUxqZK8dG3W/+HEvg=="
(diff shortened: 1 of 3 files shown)


sarfata:
ping? @tkurki are you ok merging this?

tkurki:
_all the tests run without it_ does not mean that it is not used: try mangling a timestamp in one of the test full files and run the tests - they will fail. This dependency adds validation of `date-time` JSON Schema data type that is in use in the schema.

Options on how to move forward:
- update to the current version of tv4 & formats
- switch to a better validation tool #246 that [supports checking for unknown properties](https://github.com/SignalK/specification/issues/437#issuecomment-386426798)

Then there is the issue of separating the schema js API #435 (including validation) from the actual schema data itself, versioning them separately. Then we would have the option two npm packages, one with schema data and js access to it and another for validation, versioned independently.

sarfata:
@tjkurki Sorry about that. My bad.

Pushed #501 instead. Hopefully this works for you.

PR #499, opened 2018-08-16
fix: make package usable from git

sarfata:
## Description of the problem

Right now it is very hard to work with SignalK/specification from a git repository because the package is not usable when it's installed from git.

For example:
```
"dependencies": {
  "@signalk/nmea0183-utilities": "^0.6.0",
  "@signalk/signalk-schema": "git://github.com/sarfata/specification#b3502bf",
```
Leads to:
```
 Module not found: Error: Can't resolve '@signalk/signalk-schema' in '/home/circleci/repo/node_modules/@signalk/n2k-signalk'
```

That is because most of the useful files, including the `main` file `dist/index.js` are generated, and installing from a git dependency does not generate the files.

You can test this for yourself:
 - Clone a fresh copy of the repository
 - Run `npm install --production`
 - Observe that `dist` is empty

## How I fixed it

To fix this we need to:
 - Run `npm prepublish` when the package is installed
 - Make sure that the files we need are not removed by npm (see changes in `.npmignore`)
 - Make sure `npm run docs:keys` is called before `npm run js:dist` because otherwise the file `keyswithmetadata.json` is not copied to the `dist` folder.

## Discussion

Biggest problem with this change is that everyone who depends on signalk-schema will now depend on `babel` as well. We are making the dependency much bigger. However, I do not see a way around this if we care about the package being usable from git.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -7,17 +7,11 @@ pids
 *.pid
 *.seed
 
-# source files (transpiled under dist)
-src
-
 # samples
 samples
 
 # test
 test
-
-#scripts
-scripts
 
 # Compiled binary addons (http://nodejs.org/api/addons.html)
 build/Release
@@ -37,6 +31,3 @@ thumbs.db
 *.swp
 *~
 .vscode
-
-#Gitbook source files, original and generated
-gitbook-docs
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@signalk/signalk-schema",
-  "version": "1.0.3",
+  "version": "1.0.4",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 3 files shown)


tkurki:
I do my cross repo work with separate clones and `npm linl`. Do yoy really need git dependencies? For CI you do.

sarfata:
> Do you really need git dependencies? For CI you do.

Yes exactly. My problem is that I need the package somewhere for my CI to build, test and deploy.

Another solution would be to produce a `.tar.gz` and store it somewhere until the PRs are merged but it's manual and not ideal either. 

I guess it's really a philosophical question of whether a package should be usable via git. I have not found any really authoritative answer on that.

sarfata:
Closing this. I do not think we want to merge it. Agree that making babel a dependency is not worth it here.

PR #502, opened 2018-09-13
feature: remove old docson schema viewer

tkurki:
The schema viewer has not been functional for quite long
while and has been largely surpassed by the keys reference in the
Gitbook documentation, so remove it and replace /specification/
with a redirect to a 'well known path', /latest/, that should in
turn do the redirect to the current latest version.


timmathews:
I'm 90% sure that we can remove [schemas/def_loader.json](/SignalK/specification/blob/master/schemas/def_loader.json) and [schemas/def_loader_pre.json](/SignalK/specification/blob/master/schemas/def_loader_pre.json) as well. These were hacks to get docson to work.

Issue #442, opened 2018-02-28
RFC 0010: Actions (PUT/POST)


sbender9:
RFC 0010: Actions (PUT/POST)

# Summary
[summary]: #summary

We do not currently detail the way to take actions to control things on a boat. There should be a well defined way to do things like turn lights on and off and control an autopilot, etc.

# Motivation
[motivation]: #motivation

This is currently being done in proprietary ways making it difficult for app developers to add support. 'update' deltas are one way this could be done and that has been proposed in the past. That does not work because displays and devices need to be able to tell the difference between a request to change the state of thing and an update of the current state.

# Detailed design
[design]: #detailed-design

There has been some discussion already about the exact methods used to do this. PUT/POST over HTTP and/or put deltas over ws. I'm just going to use the word PUT for now here.

## Making a request to change a value
To change a value, a PUT request should be sent via HTTP or using a SignalK 'put' delta. 

The "source" field is optional. If a request is sent without the source and their is more than one source for the value, that will result in a  400 HTTP error response.

### via http
```
PUT http://localhost:3000/signalk/v1/api/vessels/self/steering/autopilot/target/headingTrue
{
  "value" = 1.52,
  "source": "actisense.204",
}
```

### via a delta
```
{
  "context": "vessels.self",
  "put": {
    "path": "steering.autopilot.target.headingTrue",
    "source": "actisense.204",
    "value": 1.52
  }
}
```

## The response to a request to change a value

The possible responses the server can make to this request.
* Permission denied
* The request is not supported
* There was an error processing the request
* The request was processed and the value has been changed
* The request has been received and is being worked on asynchronously

I will cover these for HTTP methods since there is no defined way to do request/response over ws or other protocols.

### HTTP response for permission denied

HTTP response code 403 (Forbidden)

### HTTP response when the request is not supported

HTTP response code 405 (Method Not Allowed)

### HTTP response when there is an error processing the request

HTTP response codes:
 400 (something wrong with the client's request) 
 502 (something went wrong carrying out the request on the server side)
 504 (timeout on the server side trying to carry out the request

JSON response body:
```
{
  "state": "COMPLETED",
  "message": "Unable to reach device"
}
```

### HTTP Response to a successful PUT request

HTTP response code 200 (OK)
JSON response body:
```
{
  "state": "COMPLETED",
}
```

### HTTP Response to a request that is being worked on asynchronously

The response in this case includes an optional `href` that can be used to check the status of the request.

HTTP response code 202 (Accepted)
[...]

PR #507, opened 2018-09-20
feature: PUT Requests

sbender9:
See https://github.com/SignalK/specification/issues/442 for past discussions.

This, like #506 and #505 brings up the requirement to define the semantics for request/response over WS and other non HTTP protocols.

### DIFF
--- a/gitbook-docs/put.md
+++ b/gitbook-docs/put.md
@@ -1,0 +1,141 @@
+# PUT Requests
+
+PUT requestes are sent to a server to request a change to a value. For example, a client would use PUT to switch the anchor light on or off, change the heading of the autopilot, or set position of the anchor.
+
+## Making a request to change a value
+To change a value, a PUT request should be sent via HTTP or using a SignalK 'put' delta. 
+
+The "source" field is optional. If a request is sent without the source and their is more than one source for the value, that will result in a  400 HTTP error response.
+
+### via http
+```
+PUT http://localhost:3000/signalk/v1/api/vessels/self/steering/autopilot/target/headingTrue
+{
+  "value" = 1.52,
+  "source": "actisense.204",
+}
+```
+
+### via a delta
+```
+{
+  "context": "vessels.self",
+  "put": {
+    "path": "steering.autopilot.target.headingTrue",
+    "source": "actisense.204",
+    "value": 1.52
+  }
+}
+```
+
+## The response to a request to change a value
+
+The possible responses the server can make to this request.
+* Permission denied
+* The request is not supported
+* There was an error processing the request
+* The request was processed and the value has been changed
+* The request has been received and is being worked on asynchronously
+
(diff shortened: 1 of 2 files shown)


rob42:
I think this needs some changes. 

`correlationId` should be `requestId`.

When processing an incoming message its easy to look for `update`,`put`, `subscribe` etc in the body and process that accordingly.
The format here means thats more complex, having to look at the requestId and decide if its a reply etc.

I would use something like:

PUT Request
```
{
  "context": "vessels.self",
  "requestId": "184743-434373-348483",
  "put": {
    "path": "steering.autopilot.target.headingTrue",
    "source": "actisense.204",
    "value": 1.52
  }
}
```
Response
```
{
   "context" : "vessels.self",
   "requestId": "184743-434373-348483",
   "state": "COMPLETED"
   "result" : 200
   "message|percentComplete|etc"
}
```
The client should already know what they requested, and any changes will have been propagated by updates any way, in fact the following should be valid too (it is in artemis):

```
{
	"context": "vessels.self",
	"requestId": "184743-434373-348483",
	"state": "COMPLETED",
	"result": 200,
	"updates": [{
		"timestamp": "2013-10-08T15:47:28.263Z",
		"source": "actisense.204",
		"values": [{
			"path": "steering.autopilot.target.headingTrue",
			"value": 1.52
		}]
	}]
}
```

sbender9:
@rob42 agreed. I removed the details about the put request from the response.

fabdrol:
See my notes on https://github.com/SignalK/specification/pull/508

gitbook-docs/put.md
```diff
@@ -0,0 +1,97 @@
+# PUT Requests
+
+PUT requestes are sent to a server to request a change to a value. For example, a client would use PUT to switch the anchor light on or off, change the heading of the autopilot, or set position of the anchor.
```
neftaly: spelling :)

gitbook-docs/put.md
```diff
@@ -0,0 +1,97 @@
...
+}
+```
+
+#### Response to `/signalk/v1/api/actions/12567` when the request is pending
+(note that percentComplete is optional)
+```
+{
+   "context" : "vessels.self",
+   "state": "PENDING",
+   "percentComplete": 0.45,
```
neftaly: Worth calling it `progress`, seeing as it isn't an integer percentage?

sbender9: In Signal K percentages are specified as a ratio, a number between 0 and 1.

tkurki: Percent is misleading. ratioCompleted, completedRatio?

sbender9: I actually like `progress` better. It's more like the rest of spec. We don't use `humidityRatio`, just `humidity`

sbender9: Document has been updated to use `percentComplete`

sbender9:
Unless there are an objections, I think this is ready to go.

PR #508, opened 2018-09-20
feature: define request reponse semantics

sbender9:


rob42:
"The response when a server successfully processes a login request synchronously"
The response should actually be a login object (which is undefined yet.)

If we use this as a generic async message correlation mechanism, then the  _delta_ format is better described as the _envelope_ . The header can have any of `context`, `token`, `requestId`, `status`, `result`, and  one  _payload_ object (PUT, LIST, GET, ACCESS, LOGIN, and others later). 

A request will be 
```
{
  "context": "vessels.self",
  "token": "gdggdggdggdggdgdgdggdggd",
  "requestId": "1234567890",
  "put": {
    "path": "steering.autopilot.target.headingTrue",
    "source": "actisense.204",
    "value": 1.52
  }
}

```
a reply will be
```
{
  "context": "vessels.self",
  "token": "gdggdggdggdggdgdgdggdggd",
  "requestId": "1234567890",
  "state": "COMPLETED",
  "result": "SUCCESS"
}
```
To query a PENDING request we would use
```
{
  "context": "vessels.self",
  "token": "gdggdggdggdggdgdgdggdggd",
  "requestId": "1234567890",
  "query": true
}
```
The same query in http could be made via a well-know api call, /signalk/v1/query/request/{requestId}

rob42:
While its a small point we might be better to simply return an HTTP code in `state`. Its easier/faster to check a numeric value, and saves having to define mappings for other http codes that may prove useful.

Since the use of http codes are well understood its also one less thing to learn for newbies

rob42:
But how long do you wait on an update to confirm your request was accepted/processed? 
The `query` allows you to specifically see the current state of the request.

Also we have had people request a mechanism to request an action, then recieve a confirmation it had happened. This fulfils that need too.

sbender9:
I don’t see what that buys you, if you have not received a response from the server then a query is not going to give you anything new. Same as if you don’t get an update on wind speed, querying for it won’t help you.

rob42:
What about the case where a request goes PENDING, then FAIL. How will you learn about that if no update ever comes out because nothing changed?

sbender9:
An update would go out when it fails.

sbender9:
Let me add a WS example to the document, should make it more clear...

tkurki:
How does a client know which paths it should use a `put` message for? Like if I want to create a virtual switch panel web app how do I retrieve the list of switches and their potential states, eg. the values that I can `put`?

rob42:
> An response would go out when it fails. And state would change to COMPLETED
The response can be sent over ws, but for REST the client connection is gone so the REST client checks the action url. `query` is the ws equivalent. Its useful for ws when you want to know current state, eg you may have missed the response message, or it seems to be taking too long.

rob42:
> How does a client know which paths it should use a `put` message for? Like if I want to create a virtual switch panel web app how do I retrieve the list of switches and their potential states, eg. the values that I can `put`?

I think PUT should be valid for any path, just as UPDATES are. The difference is PUT is a one-off change to a path, with success/pending/fail semantics (like TCP - a reliable transaction). An UPDATE is a notification that a value has changed, (like UDP - fire and forget).

tkurki:
Are you guys absolutely certain that we do need the message-based version of request-response at this stage? I honestly think that we should stick to http for this. If the connection is ws then the client could as well use http for the requests. Why would a client use request-response over WebSocket as opposed to http? Ws implementation makes server implementation needlessly complex for very little added benefit. What is the case that is considerably better implemented with request-response over ws, if the other option is http?

As for the _we need to eventually support MQTT/STOMP/plain tcp_ argument I don't buy it. I don't think considering a connection oriented protocol like WebSocket will provide solutions that will work over other paradigms. Like how exactly will a MQTT client get the response back from the server so that the response reaches only that particular client, not all connected clients? I know request response can be done over MQTT, but it involves different things than the protocol that works over ws.

As for in band authentication: why not use the authentication on the protocol level? MQTT, STOMP and CoAP all include authentication mechanisms.

I don't think it makes sense to consider plain tcp for anything more complex. What would be the situation where tcp was the preferred method, over http+ws or CoAP or MQTT? There are multiple different protocols built over tcp that provide the qualities that we need. Why would you want to do the same things, probably less than perfectly and in a way that is specific to Signal K?

I think it would be interesting to map SK things properly to MQTT and CoAP and figure out where each protocol excels in the SK problem space. But I think we should do that by starting with the protocol's mechanisms, not from the abstract.

Sorry if this comes off as a rant, I just want to seriously consider this from different angles, including "do we really need this".

rob42:
Its a valid question. Yes we do need it. It fulfils two use cases in particular:
1) tcp clients without an http stack. Also applies to bluetooth, LO-RA, serial or one-wire connections and whatever comes next year. MQTT and COAP request/response mechanisms are aimed at the current message only, not the signalk action problem. eg like in http if you want to check on an async app you need to send another query message. 

2) We get regular issues about being able to perform an action and get confirmation the action occurred or not. This solves that. 

Assuming that all clients that will want action/confirmation semantics will always have an http(s) stack is short-sighted. What about safety issues like MOB, or auto SOS button, where you want more  than a fire and forget update.

Its not a complex or burdensome change. Some additional optional message headers, and an action object definition.

sbender9:
@rob42 I've made some changes to incorporate your suggestions. Please review.

sbender9:
> How does a client know which paths it should use a `put` message for? Like if I want to create a virtual switch panel web app how do I retrieve the list of switches and their potential states, eg. the values that I can `put`?

@tkurki I don'y think we should cover that here. We can get into that in the docs on the specific requests, like the 'PUT' document.

rob42:
line 26: "result wil one of the following and standard HTTP result codes are used." The `result` will be any standard HTTP code including the folowing:

Line 37: Needs more detail. "The response object may contain other objects depending on the specific request being made. For example, a response to auth request could contain an auth object."
```
{
  "context": "vessels.self",
  "requestId": "123345-23232-232323",
  "state": "COMPLETED",
  "result": 200,
  "auth": {
       "token": "....." 
   }
}
```
Sorry the above is not a great example but I cant find the login/logout issue, 

Line 71: "The state of a request can also be by sending the following:" The state of a request can also be found by sending the following:

sbender9:
@rob42 I've updated the doc as you requested above.

I've also added a `correlationId`. This is added because the server cannot guarantee the uniqueness of client generated id's, but still needs to be able to uniquely identify requests.

The `correlationId` is only used by the client for matching up asynchronous messages.

rob42:
If we define `requestId` as a real uuid rather than a random string then collisions should never happen and we can avoid also having `correlationId`?

rob42:
I dont mind having both. But it also actually highlights a different issue. 

If two devices use the same bad behaviour, or a client sends the same message twice, how should the server respond? 

Should we perform the action twice, ignore the second, or do we respond with error `invalid requestId`.

If we specify that `requestId` is a uuid and must be unique for its lifecycle, we can test for uniqueness (within currently running requestIds) then we should not have this problem

rob42:
Also in the REST interface how should a query a completed id behave?

sbender9:
Yes, I was thinking requestId is a uuid, I'll add that to the document.

sbender9:
> Also in the REST interface how should a query a completed id behave?

I just updated it to reflect changes in the put document, but it's the last example in the document.

sbender9:
> Yes, I was thinking requestId is a uuid, I'll add that to the document.

Actually, this is a server implementation detail, I don't see any reason to put this in the spec.

sbender9:
> Also in the REST interface how should a query on a completed id behave? eg one completed 48 hrs ago? Do we need to keep endless history...

I think we can leave this up to the server. Some servers might want to persist this info for ever in a db. Good info for auditing, etc. Other's might choose a reasonable timeout to prune the info.

sbender9:
@rob42 after some discussions with @tkurki, I've been convinced to remove correlationId. I've gone back to `requestId` and specified that  it must be a valid v4 uuid.

rob42:
Im happy with that, possibly a note to say a server may refuse a request if the requestId is not unique...maybe that inherent in the reply options anyway.

tkurki:
Why is context here - is it part of the request response protocol? If so what problem does it solve and what values could it have?

I think context is just a part of the request payload, for example a put request, and not request response semantics.

tkurki:
I would structure the examples so that the requestId is always the first property and separate the payload with a newline.

tkurki:
Add `401` to results? For the case where a client is not authenticated on a server where read only ws is allowed without authentication.

>- 401 - missing or bad authentication

gitbook-docs/request_response.md
```diff
@@ -0,0 +1,130 @@
+# Request/Response
+
+## WebSockets and similar transports
```
fabdrol: `similar transports` is ambiguous: WS as an extension to HTTP is much more similar to WS than it is to it's distant forefather, TCP. I thinks we should define this such as:

`WebSockets and other full-duplex streaming message-based protocols; e.g. TCP streams, UDP, 1-wire, etc`

fabdrol: My English sentence could use some TLC

gitbook-docs/request_response.md
```diff
@@ -0,0 +1,130 @@
...
+- 502 - something went wrong carrying out the request on the server side
+- 400 - something is wrong with the client's request
+- 504 - timeout on the server side trying to carry out the request
+- 405 - the server does not support the request
+- 401 - the request has not been applied because it lacks valid authentication credentials
+- 403 - the client does not have permission to make the request
+
+The message can optionally contain a `message`.
+
+The response object may contain other objects depending on the specific request being made. For example, a response to auth request could contain a `login` object.
```
fabdrol: Wouldn't it be easier for implementors to have a single, optional key `payload` which can be used to convey information such as this example? That way the client doesn't need to use a different key every single time, which is cumbersome in my experience.

gitbook-docs/request_response.md
```diff
@@ -0,0 +1,130 @@
...
+}
+```
+
+When processing is done, but it was not succesfull:
+```json
+{
+  "context": "vessels.self",
+  "requestId": "123345-23232-232323",
+  "state": "COMPLETED",
+  "result": 502,
```
fabdrol: Result, here and in the other places, as HTTP code, makes sense to me. The key name `result` does not however. `code` is an oft-used convention when sending errors (e.g. `code: 401, message: invalid credentials`)

gitbook-docs/request_response.md
```diff
@@ -0,0 +1,130 @@
...
+
+The contents of the response message when checking the status will include the values defined above for the `result` object and may also include extra information related to the request.
+
+For example, the result of a PUT request:
+```json
+{
+   "state": "COMPLETED",
+   "result": 200
+}
+```
```
fabdrol: Shouldn't these last two contain `requestId`, so that the envelope is consistent across the board?

fabdrol:
I put some comments in the code, but in general:

- Generally this looks good to me, although I would like to suggest some small convention-changes (see my code comments)
- The envelope for the request and response messages is slightly different across situations. It makes sense to me to standardise this to an envelope with these keys: `requestId`, `context`, `state`, `code` (result), `message` (optional), `payload` (optional). Any request/response-specific data is then put in the payload.

gitbook-docs/request_response.md
JonRowe: How about `WebSockets and other full-duplex protocols`

gitbook-docs/request_response.md
sbender9: How about `resultCode`, I don't really like just `code` because it's not verbose enough, IMHO.

gitbook-docs/request_response.md
sbender9: I'm not sure that `context` is always relevant. For example, I don't think makes sense in the device access request stuff. 

For HTTP, the client never needs the `requestId`, is I have left it out.

tkurki: See my question above - I don't think `context` is part of the request response protocol, but part of the payload.

fabdrol: @tkurki that makes sense to me

gitbook-docs/request_response.md
fabdrol: I guess both our arguments are 80% preferential; let's see if anyone else has an opinion here and if not we'll stick with what you have.

gitbook-docs/request_response.md
fabdrol: Works for me

gitbook-docs/request_response.md
sbender9: removed

gitbook-docs/request_response.md
sbender9: I initially had `response` instead of payload, but @rob42 requested that we have keys for specific requests.

gitbook-docs/request_response.md
timmathews: How about `statusCode` since that's what they're called in the [RFC](https://tools.ietf.org/html/rfc7231#section-6) which defines them.

sbender9: I'm fine with that.

gitbook-docs/request_response.md
```diff
@@ -0,0 +1,125 @@
...
+
+A server may respond to a request multiple times depending on how it processes the request.
+
+When a server cannot process the request immediately, it will respond with the `state` PENDING:
+```json
+{
+  "requestId": "123345-23232-232323",
+  "state": "PENDING"
```
timmathews: To be consistent with the HTTP semantics below, this ought to include `"statusCode": 202`, otherwise it feels like a special case.

gitbook-docs/request_response.md
```diff
@@ -0,0 +1,137 @@
...
+}
+```
+
+This will result in a reply like the examples above.
+
+## HTTP
+
+HTTP requests use REST API semantics and the responses are similar to the `response` object used above.
+
+One difference is that the `statusCode` value is sent as the HTTP response code.
```
timmathews: I think @rob42 brought this up elsewhere, but `statusCode` should be included in HTTP responses for consistency. Likely all of the other transports will need to have them in the body of the response, so it seems a little odd to special-case HTTP here.

gitbook-docs/request_response.md
```diff
@@ -0,0 +1,137 @@
...
+
+The response when a server successfully processes a login request synchronously:
+
+HTTP response code 200
+
+```json
+{
+  "state": "COMPLETED",
+  "token": "eyJhbGciOiJIUzI1NiI...aQ8sN1XBAP8bt3tNBT1WiIttm3qM"
```
timmathews: These are only examples, but above `token` was wrapped in a `login` object. Should it not be the same here?

gitbook-docs/request_response.md
```diff
@@ -0,0 +1,137 @@
...
+object and may also include extra information related to the request.
+
+For example, the result of a PUT request:
+
+```json
+{
+   "state": "COMPLETED",
+   "statusCode": 200
+}
+```
```
timmathews: Why isn't the `requestId` included in the HTTP case?

gitbook-docs/request_response.md
```diff
@@ -0,0 +1,137 @@
...
+  "context": "vessels.self",
+  "requestId": "123345-23232-232323",
+  "put": {
+    "path": "electrical.switches.anchorLight.state",
+    "value": 1
+  }
+}
+```
+
+The server will respond with a message which includes the `requestId`, `correlationId` if provided, and a `state`.
```
timmathews: `correlationId` is never mentioned again. What's it for?

gitbook-docs/request_response.md
sbender9: fixed

gitbook-docs/request_response.md
sbender9: That was left over from a previous version of the document. I removed it.

gitbook-docs/request_response.md
sbender9: It's because if's not needed by the client. But as with statusCode, we should just be consistent. I will add it.

sbender9:
> I put some comments in the code, but in general:
> 
> * Generally this looks good to me, although I would like to suggest some small convention-changes (see my code comments)
> * The envelope for the request and response messages is slightly different across situations. It makes sense to me to standardise this to an envelope with these keys: `requestId`, `context`, `state`, `code` (result), `message` (optional), `payload` (optional). Any request/response-specific data is then put in the payload.

I think this issue about the `payload` is the only problem remaining here. 

@rob42 @fabdrol I don't care, either way is fine with me. Can we come to an agreement?

rob42:
I prefer not to use `payload` to hold the response. The actions (`put`,`authenticate`, etc) are already unique, I think `payload` will become the holder for arbitrary data (eg images) that we want to transfer - I had a use case for that recently. But different issue to this.

gitbook-docs/request_response.md
```diff
@@ -0,0 +1,142 @@
+# Request/Response
+
+## WebSockets and Other Full-duplex Protocols
+
+The exact format of the message for a specific request is defined elsewhere in the specification.
+
+A request should include the `context` when appropriate and must include a client generated `requestId`. The
```
tkurki: I think I asked this earlier, so sorry if I am repeating myself here: does `context` have a function in relation to request/response mechanism? If it is related to the individual, specific request then it is confusing to have it mentioned here in the request/response section.

gitbook-docs/request_response.md
```diff
@@ -0,0 +1,142 @@
+# Request/Response
```
tkurki: Do we need a short intro, raison d'etre for req/resp?

gitbook-docs/request_response.md
```diff
@@ -0,0 +1,142 @@
...
+
+- 200 - the request was successful
+- 400 - something is wrong with the client's request
+- 401 - the request has not been applied because it lacks valid authentication credentials
+- 403 - the client does not have permission to make the request
+- 405 - the server does not support the request
+- 502 - something went wrong carrying out the request on the server side
+- 504 - timeout on the server side trying to carry out the request
+
+The message can optionally contain a `message`.
```
tkurki: How about _...contain an informational, human oriented `message`_.

gitbook-docs/request_response.md
```diff
@@ -0,0 +1,142 @@
...
+
+The response object may contain other objects depending on the specific request being made. For example, a response to
```
tkurki: Maybe _other objects_  => _other response data_?

gitbook-docs/request_response.md
sbender9: I took a shot at it. Seems like it needs more?

sbender9:
Unless there are any objections, I think this is ready to go

Issue #363, opened 2017-06-18
Track API


tkurki:
I added track handling as part of the http API to the Node server InfluxDb plugin. 

it creates a `/track` endpoint under self vessel, but I don't see why the same API could not be used for any vessel.

The endpoint accepts two query parameters, `timespan` and `resolution`, each in in xxxY format, where `xxx` is a _Number_ and `Y` one of 
- s  (seconds)
- m  (minutes)
- h  (hours)
- d  (days)
- w  (weeks)
Duration expression is a direct copy from InfluxDb's duration, just omitting ms and us.

The response is a GeoJSON MultiLineString with proper mime type `application/vnd.geo+json`.

The big difference here is that **this is real API endpoint** as opposed to just adding stuff to the Signal K full document and the equivalent RESTish paths. I don't think it makes sense to include this data in the full response, as the data is different in nature (historical time series) versus latest values.

Per the current implementation the track is just a MultiLineString: a list of position lists. There is no time data in the response. I am not aware of any standard way to represent tracks with time information. I think it would be a good idea to define either a separate mime type (application/json comes to mind) or an optional query string that would allow you to retrieve  the same MultiLineString plus the corresponding timestamp data in two top level properties:
```json
{
  timestamp: [ [...], [...],...]
  geojson: {..MultiLineString. }
}
```

tkurki:
https://github.com/tkurki/signalk-to-influxdb/pull/9

tkurki:
https://github.com/SignalK/specification/issues/89

tkurki:
Another addition would be the ability to specify start and end timestamps.

tkurki:
Another alternative way to retrieve track data is by bounding box.

rob42:
Nice addition. I agree its an api call, not an extension to the data model. Geojson is a great way to transport the track as it works most web-mapping (like freeboard-sk) and is a defined std. I think form memory there is the ability to add arbitrary properties into the geojson object. 

Then we could define a specific one for the timestamp.

rob42:
Same api call technique can apply to grabbing data for any timespan or  bounding box. Vessels, routes, features, charts, notes...

tkurki:
I partially rewrote the Node implementation based on geohashes in SQLite and actual data in InfluxDb. This version supports just bounding box searching, no time related searching.

The response is now a GeoJSON FeatureCollection with name, starttime and endtime metadata in track feature's properties.

https://github.com/tkurki/signalk-to-influxdb/pull/12

[...]

PR #510, opened 2018-10-02
History playback and snapshot retrieval APIs

tkurki:
Add query parameters to stream historical data and retrieve historical snapshot values.


tkurki:
Ping @panaaj

tkurki:
No longer WIP.

tkurki:
It was WIP, no longer is and should be reviewed.

samples/hello/docs-hello-history.json
```diff
@@ -0,0 +1,11 @@
...
+    "name": "foobar marine server",
+    "version": "1.1.4",
+    "startTime": "2018-08-24T15:19:09Z",
+    "playbackRate": 1,
+    "self": "vessels.urn:mrn:signalk:uuid:c0d79334-4e25-4245-8892-54e8ccc8021d",
+    "roles": [
+        "master",
+        "main"
+    ]
+}
```
JonRowe: There's a line ending error here.

gitbook-docs/streaming_api.md
```diff
@@ -50,3 +50,35 @@ The server MAY provide:
...
+    ]
+}
+```
+[<]: #
+
+Since this is an addition to the 1.0 Signal K the client can not reliably detect if the server supports history playback streaming from http status code. A server MAY respond with `501 Not Implemented` status code if it does not support history playback and with `400 Bad Request` if it does not have data to play back for the given time period.
+
+
+
+
```
JonRowe: Is it necessary to commit a bunch of blank lines here? Would be tidier if they were removed.

rob42:
See swagger definition at https://github.com/SignalK/signalk-java/blob/artemis/signalk-static/docs/openapi.json
When https://github.com/SignalK/signalk-java/tree/artemis is installed  the swaggerUI is available  at http://localhost:8080/docs/

sarfata:
I think an important use-case is grabbing a big chunk of data over time in one swoop, for example to draw a trace on a map or to graph speed over time. 

With the current proposal, this would be very hard to do: you would have to either make a lot of HTTP calls to the snapshot retrieval API - or to the websocket API with playbackRate=100000.

Suggestion of changes to the websocket endpoint:
 - Add optional `endTime` parameter. If specified, the server will disconnect when replay (or realtime) has reached this point in time.
 - If playbackRate is not specified, the server will send all the data it has between `startTime` and `endTime` (or now if endTime is not specified) as fast as possible.

I think we should also support a way to let clients know what data is available but that could come separately.

tkurki:
You are talking about history and/or track api. This is only for playback/snapshot retrieval.

https://github.com/SignalK/specification/issues/363
https://github.com/SignalK/specification/issues/89

gitbook-docs/rest_api.md
```diff
@@ -65,3 +65,9 @@ of the data under the `vessels` container in JSON format. Likewise, `GET
 /signalk/v1/api/vessels/urn:mrn:signalk:uuid:c0d79334-4e25-4245-8892-54e8ccc8021d` should return data for one specific
 vessel. In other words, the full Signal K data model SHOULD be traversable by any client making GET requests to an
 arbitrary depth.
+
+## History snapshot retrieval
+
+A server MAY support retrieving historical data. The client specifies the requested timestamp with request parameter `time`, for example `https://localhost:3443/signalk/v1/api/vessels/self?time=2018-08-24T15:19:09Z`. The server will attempt to create the request part of the full model at the requested time.
+
+Since this is an addition to the 1.0 Signal K the client can not reliably detect if the server supports history snapshot retrieval from http status code, but should check the timestamps in the returned data. A server MAY respond with `501 Not Implemented` status code if it does not support history snapshot retrieval and with `400 Bad Request` if it does not have data for the requested timestamp.
```
sarfata: Why don't we make this a 1.1 feature?

gitbook-docs/rest_api.md
```diff
@@ -65,3 +65,9 @@ of the data under the `vessels` container in JSON format. Likewise, `GET
 /signalk/v1/api/vessels/urn:mrn:signalk:uuid:c0d79334-4e25-4245-8892-54e8ccc8021d` should return data for one specific
 vessel. In other words, the full Signal K data model SHOULD be traversable by any client making GET requests to an
 arbitrary depth.
+
+## History snapshot retrieval
+
+A server MAY support retrieving historical data. The client specifies the requested timestamp with request parameter `time`, for example `https://localhost:3443/signalk/v1/api/vessels/self?time=2018-08-24T15:19:09Z`. The server will attempt to create the request part of the full model at the requested time.
```
sarfata: I have noticed Rob42 mentioning that some server support a `*` in the path. I am not sure we define this anywhere in the spec, so I assume the request here is for one context only and the response is an object, not a list.

gitbook-docs/streaming_api.md
```diff
@@ -50,3 +50,31 @@ The server MAY provide:
 `name`, `self` and `roles` MUST return the same values as provided in the `swname`, `self` and `roles` properties within the [DNS-SD advertisement](connection.md) (if implemented).
 
 `version` MUST be the same value as `version` within the associated endpoints list provided by the http `GET` request to `/signalk` within the [REST API](rest_api.md) (if implemented).
+
+## History playback
+
+The server MAY support history playback from a certain point in time with a specified rate.
+
+To create a WebSocket connection that plays back data the client uses the request parameter `startTime` to specify the start timestamp and the optional request parameter `playbackRate` to specify the rate. Rate value parameter is a floating point value with value `1` equal to real time playback and for example `0.5` to half the real time rate and `5` to five times real time rate. Omitting the `playbackRate` will result in real time playback.
```
sarfata: I still think no `playbackRate` should return the entire set of data available and then close the connection.

gitbook-docs/streaming_api.md
```diff
@@ -50,3 +50,31 @@ The server MAY provide:
...
+
+Example url for history playback streaming: `wss://localhost:3443/signalk/v1/stream?subscribe=self&startTime=2018-08-24T15:19:09Z&playbackRate=5`.
+
+The hello message for a history playback stream MUST NOT contain the `timestamp` property and MUST include the properties `startTime` and `playbackRate`. The delta stream format for history playback is the normal streaming format except for the . Timestamps will indicate the time data was originally captured.
+
```
sarfata: I think we should add: `When the server has finished playing back available data, the connection is closed.`

gitbook-docs/streaming_api.md
```diff
@@ -50,3 +50,31 @@ The server MAY provide:
...
 
 `version` MUST be the same value as `version` within the associated endpoints list provided by the http `GET` request to `/signalk` within the [REST API](rest_api.md) (if implemented).
+
+## History playback
+
+The server MAY support history playback from a certain point in time with a specified rate.
+
+To create a WebSocket connection that plays back data the client uses the request parameter `startTime` to specify the start timestamp and the optional request parameter `playbackRate` to specify the rate. Rate value parameter is a floating point value with value `1` equal to real time playback and for example `0.5` to half the real time rate and `5` to five times real time rate. Omitting the `playbackRate` will result in real time playback.
+
+Example url for history playback streaming: `wss://localhost:3443/signalk/v1/stream?subscribe=self&startTime=2018-08-24T15:19:09Z&playbackRate=5`.
```
sarfata: Why don't we make this a separate URL endpoint? This could be a `v1/playback`. This would make it very easy to detect if the server supports it.

gitbook-docs/rest_api.md
sarfata: > The server will attempt to create the request part of the full model at the requested time.

Considering that we have multiple server implementation, maybe we could be more specific:
 - The server will reconstruct the SignalK full state as it would have been served at this time.
 - All the values returned by the server will have a timestamp equal or small to the requested snapshot time.

gitbook-docs/rest_api.md
```diff
@@ -65,3 +65,9 @@ of the data under the `vessels` container in JSON format. Likewise, `GET
 /signalk/v1/api/vessels/urn:mrn:signalk:uuid:c0d79334-4e25-4245-8892-54e8ccc8021d` should return data for one specific
 vessel. In other words, the full Signal K data model SHOULD be traversable by any client making GET requests to an
 arbitrary depth.
+
+## History snapshot retrieval
```
sarfata: As a suggestion to debate, how about making the URL contain the timestamp, this way it would be a bit more RESTful:

```
GET /signalk/v1/snapshot/2018-08-24T15:19:09Z/vessels/self
```

gitbook-docs/streaming_api.md
tkurki: I don't think WebSocket is the way to serve this use case: browser does not do anything with the output, so downloading your raw log data is not really feasible via ws. Command line tools don't work/exist and probably all other client libs need a bit of extra effort, compared to http.

Your use case makes perfect sense, just that I don't think this is way to solve it.

gitbook-docs/streaming_api.md
tkurki: Why? When is the available data exhausted? Like if I sail some, stop for lunch x minutes and turn instruments off, then sail some should the playback stop?

gitbook-docs/rest_api.md
tkurki: Did not think about this option. Using the same url as normally with an extra parameter makes client implementation maybe more straightforward and the query parameter can be thought of as a qualifier: 'i want resource X, with this twist'. A separate prefix may be more explicit and cleaner in that way.

Any other APIs we could draw experience from?

gitbook-docs/streaming_api.md
tkurki: That would certainly be more explicit. Then again using the same url with extra parameters is pretty simple and straightforward imho.

@sbender9 ?

sbender9: For me, it makes the client code much simpler to keep the urls the same and add a parameter.

gitbook-docs/rest_api.md
sbender9: This makes things more complicated on the client side. Much easier to add a parameter than having to much with the path.

gitbook-docs/streaming_api.md
sbender9: I like having a separate endPoint. I currently don't really have a way to tell if the server supports this.

gitbook-docs/rest_api.md
sarfata: You are right, the word 'reconstruct' is probably enough here.

The other way I thought it could be interpreted is: "the server will grab all the 'updates' with the requested timestamp and serve them", which could have meant that if you asked for data at 10:00:00 but the gps update with position and sog was at 9:59:59.800 you would not get it.

gitbook-docs/streaming_api.md
tkurki: What is the end and what is just a hole? Ihave data over several years with different size holes. 

We can leave this for the server to decide of course. One way is just to keep the connection hanging and skip this.

gitbook-docs/rest_api.md
sbender9: True, and existing server will probably just ignore the query parameter and return he current data, so it would look like they support it.

tkurki:
So, separate endpoints then?

gitbook-docs/rest_api.md
rob42: Another solution is to add `/signalk/v1/snapshot`, which has the advantage of being naturally 404 when unimplemented. If `time` is supported on `/signalk/v1/api` we just redirect. So its easy to check if its supported, and if so you can just use the std api?

Or we could start a `/signalk/v1.1/api` that included the new functionality. Trouble is we will get a zillion `/v1.x/` variants as we add to it, so it wont be nice.

rob42: In fact if we make a `/signalk/v1/feature/xxx` structure, we can add features as we please. The API users should then code so they test `/signak/v1/feature/xxx` before using a feature.

tkurki: Do we really need the `/feature/` part in the path - why not just `/signalk/v1/snapshot` like you suggested?

gitbook-docs/rest_api.md
tkurki: Resolved in a5c4d504aa86b867be730977dae35a97dc2e59f7

tkurki:
Github swallowed the conversations, but a5c4d504aa86b867be730977dae35a97dc2e59f7 and dae3fb2c3ba21af85f12515057765001d7d9ab3e use now separate endpoints.

Timestamp in the query string makes more sense to me in the api/snapshot full model case than inserting it in the path string, restful or not.

samples/hello/docs-hello-history.json
```diff
@@ -0,0 +1,11 @@
+{
```
sarfata: nitpicking again but we might want to rename this to `docs-hello-playback.json` just in case we do add a `history` endpoint in the future.

gitbook-docs/rest_api.md
rob42: I meant that if we dont use a separate `/signalk/v1/snapshot` path then we could use a `/signalk/v1/feature/snapshot` url to test if the server has the snapshot capability. It would 404 for a server version that did not support it, so the client would know not to try `/signalk/v1/api?time-=...` as it would just return now(). Moot if we use separate paths which is cleaner anyway.

gitbook-docs/streaming_api.md
```diff
@@ -50,3 +50,31 @@ The server MAY provide:
...
 
 `version` MUST be the same value as `version` within the associated endpoints list provided by the http `GET` request to `/signalk` within the [REST API](rest_api.md) (if implemented).
+
+## History playback
+
+The server MAY support history playback from a certain point in time with a specified rate.
+
+To create a WebSocket connection that plays back data the client uses the request parameter `startTime` to specify the start timestamp and the optional request parameter `playbackRate` to specify the rate. Rate value parameter is a floating point value with value `1` equal to real time playback and for example `0.5` to half the real time rate and `5` to five times real time rate. Omitting the `playbackRate` will result in real time playback.
+
+The playback api is located at `/signalk/v1/playback`. An example url for history playback streaming: `wss://localhost:3443/signalk/v1/playback?subscribe=self&startTime=2018-08-24T15:19:09Z&playbackRate=5`.
```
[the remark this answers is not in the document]
rob42: It should also have an endTime, so you can select a specific time period, rather than playing till you reach now(). Also `/signalk/v1/playback/` is equivalent to `/signalk/v1/stream/` not `/signalk/v1/api` isnt it?, eg /stream does a ws endpoint, /api does an http rest reply.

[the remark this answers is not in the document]
tkurki: The client can drop the connection whenever it wants, so endTime is not strictly needed.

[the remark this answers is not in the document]
tkurki: Sorry, missed your point regarding the paths?

gitbook-docs/rest_api.md
```diff
@@ -65,3 +65,9 @@ of the data under the `vessels` container in JSON format. Likewise, `GET
...
+
+A server MAY support retrieving historical data. The history snapshot retrieval endpoint is `/signalk/v1/snapshot` and functions like the full model endpoint at `/signalk/v1/api`. The client specifies the requested timestamp with request parameter `time`, for example `https://localhost:3443/signalk/v1/snapshot/vessels/self?time=2018-08-24T15:19:09Z`. The server will attempt to create the request part of the full model at the requested time.
+
+A server MAY respond with `501 Not Implemented` status code if it does not support history snapshot retrieval and with `400 Bad Request` if it does not have data for the requested timestamp. A `400 Not Found` response is also acceptable to be backwards compatible.
```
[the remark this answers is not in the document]
tkurki: 400 as it says, or am I missing something?

PR #512, opened 2018-10-05
fix: units should not be a required field on metadata

sarfata:
`n2k-signalk` tests do not pass with `signalk-specification` v1.0.4.

I have done some research and found that the issue started when we merged https://github.com/SignalK/specification/pull/462. In this PR, we have made `units` a required field on `meta` object.

However, when you use `FullSignalK.addDelta(delta)` and then `.retrieve()` we automatically add some meta tags based on information from the schema but that information is incomplete: we are missing some units which cause tests to fail. 

I have added a test here that demonstrates the issue.

```
/vessels/urn:mrn:imo:mmsi:230099999/navigation/gnss/satellites/meta:Missing required property: units
/vessels/urn:mrn:imo:mmsi:230099999/navigation/gnss/horizontalDilution/meta:Missing required property: units
/vessels/urn:mrn:imo:mmsi:230099999/navigation/gnss/positionDilution/meta:Missing required property: units
/vessels/urn:mrn:imo:mmsi:230099999/navigation/gnss/geoidalSeparation/meta:Missing required property: units
/vessels/urn:mrn:imo:mmsi:230099999/navigation/gnss/differentialReference/meta:Missing required property: units
```


sarfata:
We can fix this by adding units in the schema but most of these values are unitless. For example, the count of visible satellites or the [horizontal dilution](https://en.wikipedia.org/wiki/Dilution_of_precision_(navigation)).

Maybe the fix here is to make `units` optional again?

ping @tkurki @sbender9 @bkp7

sarfata:
Besides making `units` optional in the `meta` section (which I think is the right thing to do), another solution would be to edit [`fullsignalk.js`](https://github.com/SignalK/specification/blob/master/src/fullsignalk.js#L198) and change:

```
            previous[pathPart].meta = meta;
```

to:
```
          // units is a required field in SignalK since 1.0.4
          if (Object.keys(meta).includes('units')) {
            previous[pathPart].meta = meta;
          }
```
I have tested and confirmed it works but really that seems like a worse solution to me.

sarfata:
The description that was changed when we made `units` a required field, said:

> If a client requests the `meta` property for a valid Signal K key via the HTTP REST interface, the server must return
> the `description` and, if applicable, `units`, even if no value has ever been generated for that key.

Therefore, I think `units` should never have been made required and I have pushed a commit to roll this back.

PR #515, opened 2018-10-19
fix: remove wildcard properties in groups/sources

az0uz:
Having a wildcard `patternProperties` in parallel of `ais`, `talker`, `sentences`, `n2k` won't correctly validate compliance.

Any object that would define a wrongly typed message inside of the above properties will then fallback on matching the wildcard property which won't return an error on validation.

example `sources`:
```
{
  "label1": {
    "talker1": {
      "ais": {
        "aisType": "long range"
      }
    }
  }
}
```

## Actual Behavior
The validator will try to match the nested object with the `ais` property which won't work given that the `aisType` is a string and not a number here *but* the validator will then fallback to match `ais` with the wildcard property in the second nested object which will *succeed* the validation test.

## Expected Behavior
The validator should *fail* given that `aisType` is in the wrong format (`string` instead of `number`)

## Remark
json-schema already let's you have properties outside of the defined list in `anyOf` unless you specify `otherProperties` to false which is not the case here.

## Solution
Just removing the wildcard property should makes the definition work as expected

### DIFF
--- a/schemas/groups/sources.json
+++ b/schemas/groups/sources.json
@@ -115,7 +115,12 @@
                 }
               }
             }
-          ]
+          ],
+          "patternProperties": {
+            ".*": {
+              "properties": {}
+            }
+          }
         }
       },
       "properties": {
--- a/schemas/groups/sources.json
+++ b/schemas/groups/sources.json
@@ -114,13 +114,6 @@
                   }
                 }
               }
-            },
-            {
-              "patternProperties": {
-                ".*": {
-                  "properties": {}
-                }
-              }
             }
           ]
         }


az0uz:
I did a second commit to push the wildcard property outside of the AnyOf,
I notice it was only failing to work on others json-schema validators like python-jsonschema and Ajv but not on tv4.

with the wildcard outside of AnyOf, the test `test/data/full-invalid/ais-source0.jso` is invalid on all three validators when it was valid on python-jsonschema and Ajv before.

fabdrol:
@az0uz what's the latest on this? Do you recommend this change still, given your latest comment?

az0uz:
Hi @fabdrol,
I think you can close this issue,
it's actually a much wider issue because you are using some out of spec feature and I feel like tv4 is not respecting well the jsonSchema standard and the specs would need to be rewritten using a better validator like Ajv

fabdrol:
Thanks @az0uz - is there an issue to track this `wider issue`?

PR #526, opened 2018-11-23
Fix: reducedState as object

### LINKED ISSUE #525: sails.json
Hey,

in sails.json (https://github.com/SignalK/specification/blob/master/schemas/groups/sails.json) on line 63, 64 there is the definition for reducedStat. 

` "reducedState": {
              "properties": {
                "reduced": {
                  "type": "boolean",`

I think `"type": "object"` is missing before properties.

Best,
Martin

joabakk:
fixes #525

### DIFF
--- a/schemas/groups/sails.json
+++ b/schemas/groups/sails.json
@@ -61,6 +61,8 @@
               "default": 666
             },
             "reducedState": {
+              "type": "object",
+              "description": "An object describing reduction of sail area",
               "properties": {
                 "reduced": {
                   "type": "boolean",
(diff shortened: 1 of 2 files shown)


joabakk:
I would be happy to, when time permits. Rather surprisingly, the test never failed. Any idea how to make a test that is more general than just checking if this specific one is an object?

tkurki:
Which test never failed?

joabakk:
This one: https://github.com/SignalK/specification/blob/master/test/data/vessel-valid/sails-sample.json
I'm just trying to figure out what to test for. First of all, should all paths in the tree have a description and a type? I don't want the tests to be too specific for one PR

tkurki:
See https://github.com/heiamoss/specification/pull/1 for adding a test that detects if the schema detects invalid reduced state.

joabakk:
Thanks @tkurki you did it for me again. That was not my intention. Now how do we get travis to look at this again?

PR #528, opened 2018-12-01
Fix message schemas

tkurki:
The delta schema is clearly buggy / inconsistent, because we have never really exercised anything but the updates mechanism in it. Going forward I think it would make sense to split the current delta.json to separate schemas for

- delta proper (updates etc)
- subscribe/unsubscribe
- request/response with put (add others here when documented properly)

and create a superschema for "stream message".

Previously we have had subscribe/unsubscribe specified twice, in delta.json and under messages separately.

This PR is an attempt to make the messages schema consistent and in line with the rest of the documentation.

See also #523.


schemas/delta.json
```diff
@@ -67,155 +67,6 @@
           }
         }
       }
-    },
-    "subscribe": {
```
sarfata: So is this completely gone? I thought from the description that you were moving things around. I do not see where this went.

Completely agree with the intent and direction of this PR though.

tkurki: (un)subscribe was is in master in delta.json and then at https://github.com/SignalK/specification/tree/c31e54f047d5e50c961d24b9898f13d5c165ed2e/schemas/messages

fabdrol:
This can be merged, no?

PR #529, opened 2018-12-02
chore: update to lodash 4

sarfata:
Lodash is a fairly large dependency. Version 4 came out 3 years ago and
is what most packages use today. By upgrading to v4 we help people who
use signalk-specification: they will not need to bundle two versions of lodash.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1962,7 +1962,8 @@
           "version": "1.1.0",
           "resolved": "https://registry.npmjs.org/code-point-at/-/code-point-at-1.1.0.tgz",
           "integrity": "sha1-DQcLTQQ6W+ozovGkDi7bPZpMz3c=",
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "combined-stream": {
           "version": "1.0.5",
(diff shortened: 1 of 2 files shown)


sarfata:
(I had forgotten to update the package-lock.json - force-pushed with it - that will teach me *not* to use yarn for everything!)

tkurki:
Or should we go even step further and only include what we use? or remove the usage in some cases.

```
src/fullsignalk.js
80:  var context = _.get(root, contextPath);
83:    _.set(root, contextPath, context);
179:  if (_.isUndefined(pathValue.path) || _.isUndefined(pathValue.value)) {
185:    _.merge(context, pathValue.value)
238:  _.assign(toLeaf, _.omit(fromLeaf, ['$source', 'timestamp', 'meta']));

src/index.js
213:            _.set(result, pathValue.path, pathValue.value);
215:            _.set(result, pathValue.path + '.value', pathValue.value);
217:          _.set(result, pathValue.path + '.timestamp', update.timestamp);
220:              _.set(result, pathValue.path + '.pgn', update.source.pgn);
222:            if (!_.isUndefined(update.source.label) && update.source.src) {
223:              _.set(result, pathValue.path + "['$source']", update.source.label + '.' + update.source.src);
226:          _.set(result, pathValue.path + '.timestamp', update.timestamp);
307:_.forIn(module.exports.metadata, (value, key) => {

src/delta.js
38:      source: _.clone(source),
```

sarfata:
I looked but since there are tons of package that use it, it will end up included for most people anyways.

I think you are feature-creeping the scope of this pull request - that could be done later ;)

tkurki:
I am!

Thanks!

PR #531, opened 2018-12-21
Added headingRaw, and magneticDeviation, adjust descriptions to be clearer

rob42:
See #242

### DIFF
--- a/schemas/groups/navigation.json
+++ b/schemas/groups/navigation.json
@@ -476,17 +476,17 @@
     },
     "headingMagnetic": {
       "$ref": "../definitions.json#/definitions/numberValue",
-      "description": "Current magnetic heading of the vessel, equals 'headingRaw adjusted for magneticDeviation'",
+      "description": "Current magnetic heading of the vessel, equals 'headingCompass adjusted for magneticDeviation'",
       "units": "rad"
     },
     "magneticDeviation": {
       "$ref": "../definitions.json#/definitions/numberValue",
-      "description": "Magnetic deviation of the compass at the current headingRaw",
+      "description": "Magnetic deviation of the compass at the current headingCompass",
       "units": "rad"
     },
-    "headingRaw": {
+    "headingCompass": {
       "$ref": "../definitions.json#/definitions/numberValue",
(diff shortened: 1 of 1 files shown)


schemas/groups/navigation.json
```diff
@@ -476,12 +476,22 @@
     },
     "headingMagnetic": {
       "$ref": "../definitions.json#/definitions/numberValue",
-      "description": "Current magnetic heading of the vessel",
+      "description": "Current magnetic heading of the vessel, equals 'headingRaw adjusted for magneticDeviation'",
+      "units": "rad"
+    },
+    "magneticDeviation": {
+      "$ref": "../definitions.json#/definitions/numberValue",
+      "description": "Magnetic deviation of the compass at the current headingRaw (if known)",
```
tkurki: I would leave ’if known’ out. It applies to a lot of items in the schema. Or is there a particular reason to add it here and not all over the place?

schemas/groups/navigation.json
```diff
@@ -476,12 +476,22 @@
...
+      "units": "rad"
+    },
+    "magneticDeviation": {
+      "$ref": "../definitions.json#/definitions/numberValue",
+      "description": "Magnetic deviation of the compass at the current headingRaw",
+      "units": "rad"
+    },
+    "headingRaw": {
+      "$ref": "../definitions.json#/definitions/numberValue",
+      "description": "Current magnetic heading received from the compass.",
```
tkurki: I would write explicitly that this is *Current magnetic heading of the vessel received from the compass prior to any adjustment for deviation or declination* to tell make the relation to the other headings clear.

To me `headingRaw` is too ambiguous in this context, I think `headingCompass` would make much more sense here. 

To me this is about capturing the traditional three heading values & their companion calibration/adjustment values in the schema. If we need a generic raw/calibrated distinction mechanism I think we need something better than appending *raw* all over the place. query/subscription parameter for example. But I feel that is another issue, as I said.

rob42: I prefer `headingCompass` as well, and I agree with @timmathews  reservations about `headingRaw` as a solution.

PR #534, opened 2019-01-21
Update electrical.json

dmdelorme:
Could we add this to DC qualities. It could also use max warn and max fault.


fabdrol:
@dmdelorme please see my comments. Needs some work I think

tkurki:
Closing as stalled.

PR #535, opened 2019-01-24
Update security doc with full functionality and support for other transports

rob42:
Updated and completed security doc to reflect a full set of functionality for  web and other transports


fabdrol:
PR looks good to me, but it may make sense to get @tkurki and/or @sbender9 opinion on this too, as they did a lot of work on security in the node server. 

Something else I noticed, but not necessarily an issue with this commit:

Early in the document, it is stated that tokens can be of any type: `The tokens can be of any type, but are typically JWT tokens.See https://jwt.io/`. In the initial authentication requests, a server responds to the login by sending a response that contains the token type and the actual token. Later on, however, there are various examples of request authentication to the server that include a token, but no type (with the exception of the Authorization header approach). 

How would a server know, e.g. from a request with the `SK_COOKIE` set, what kind of token it's looking it? JWT tokens are fairly easy to recognise, but other formats may not be. I realise that this is not such a big issue in practise, as most servers will only implement a single type of token and in most cases the token will come from that server in the first place. However, it seems like an oversight to me.

gitbook-docs/security.md
```diff
@@ -112,6 +127,93 @@ Clients using other kinds of protocols can include the `token` in the Signal K m
...
+#### Reply Messages
+
+In both cases the reply will be
+
+On success:
+```
+{
+  "requestId": "1234-45653-343454",
+  "logout": {
+    "token": "eyJhbGciOiJIUzI1NiIsI...ibtv41fOnJObT4RdOyZ_UI9is8"
```
tkurki: The token is now meaningless after logout, why include it here?

gitbook-docs/security.md
```diff
@@ -112,6 +127,93 @@ Clients using other kinds of protocols can include the `token` in the Signal K m
 }
 ```
 
+### Token Validation
+
+Tokens may have a short expiry time and need to be renewed periodically, or a token's current validity may be unknown.
```
tkurki: How does a client now how often it should validate the token?

gitbook-docs/security.md
```diff
@@ -80,25 +89,31 @@ If the login fails, the server will send a response like the following:
...
 
 #### WebSockets Clients
 
-Clients can include the authentication cookie with the initial request.
+Clients can include the authentication cookie with the initial request. 
 
 Clients can include the `Authorization` HTTP header with the initial connect request. The format of the header should
 be `{type} {token}`, for example `Authorization: JWT eyJhbGciOiJIUzI1NiIsI...ibtv41fOnJObT4RdOyZ_UI9is8`
 
+They must still send the token in every message.
```
tkurki: Why would you like to add this requirement?

This is not backwards compatible.

gitbook-docs/security.md
fabdrol: @tkurki I was wondering that whilst implementing security in a client. Does Node server currently send an expiration, or do they simply not expire until server is restarted?

rob42: In my case the server handles this automatically for http/ws, so the client doesnt need to do anything. For other transports the token is validated on every incoming message, but that wont cover a long  running subscription.
The expiry is in the token, but I had brain fade, the client cant read that. 
Need to send 'expiry' in the login reply.

rob42:
@fabdrol using 'Bearer' instead of 'JWT' seems much better. Removes the need for token 'type', which should not matter to the client anyway?

gitbook-docs/security.md
rob42: It standardises the message across streaming protocols. I guess it could be injected into messages by the ws layer, but thats just more work at the server end.
If you process messages in an async streaming model the token must accompany each message for key level security to be applied.

tkurki: Ws connection can use cookie authentication, I don’t think this makes sense in that context.

How a server processes the stream is not really relevant for the client or the protocol.

For connectionful transports passing the token once is enough imho.

JonRowe: It does because a WS stream is a single request, not multiple, and a reconnecting web socket will resend its original headers (wether that be an authorization header or the cookie).

gitbook-docs/security.md
tkurki: If we define a standard way to validate we should define a standard way to inform the client about the need.

A client may not have a real time clock, so it should be time to live, not a timestamp?

gitbook-docs/security.md
JonRowe: How about `s/They must still send the token in every message./When using HTTP headers to authorize requests you must remember to send the token with every HTTP request made` then?

tkurki: The paragraphs above this one talk about cookie and header auth in ws. Why add a third one that states something that is should be fairly obvious?

JonRowe: It might not be obvious to novices

fabdrol:
@rob42 unless a server, for some reason, needs to be able to accept multiple token types (federated ID server integration or whatever). So I’m not against the `type` lest we implement it across the board (I.e. in every auth method).

@sbender9 what do you reckon? Remove everywhere for now and go with default Bearer?

sbender9:
Maybe we should should drop `type` completely? We're just making things overly complicated for clients. Let's keep it simple, if the server needs to include `type` in the token it can. We can just state that the token is a string, there's no need to specify what it looks like.

sbender9:
Too many unrelated changes in this PR, don't see how we could come to consensus on all of them.

gitbook-docs/security.md
rob42: Re sending token in every request. 

HTTP REST requests are quite different from a signalk message. Due to REST conventions we find the `token` in the header or cookie, the `context` and `path` in the url, and we expect the reply status to be in the header and follow http semantics. Since a single HTTP call is synchronous, eg the client sends and waits/timeout expecting a reply,  we can be sure that a signalk message is identified with unique `token`, `context`, and `path`. So Its a special case and does not need `token` is a request.

Ws, tcp, udp, serial, etc are simple streaming protocols. While the setup of a ws connection can include a lot of HTTP semantics, the actual connection is just a tcp pipe.

At the server end its just a stream of messages. They could be multiplexed from another server or proxy, they could be direct. To process them reliably each message should be stateless, hence we include `token`, `context`, and sometimes `requestId`. Its not safe to assume that all messages have the same auth, hence `token`, plus sending a message over a streaming protocol is asynchronous, there is no  inherent reply, so we need to correlate messages with `requestId` when appropriate.

Basically when we send a signalk message over any streaming protocol it should contain the `token` if one is required. I have no objection to using a token in ws setup, but the messages must include the token.

fabdrol: Reading through this thread, I have a couple of thoughts. 

I tend to agree with Rob's point about standardising the message between the messaging protocols (i.e. WS, TCP, etc) - but not because it makes it easier for the server (i.e. one backend or middleware to handle all the messages). @tkurki makes a good point that that is all irrelevant to our effort here. However, since we're standardising an API I feel that the more conventions the better. In my mind the solution is to make _handling the token key of the message by the server_ an optional requirement for WS. I.e.: if I implement a server that _only_ offers HTTP REST and WS transports, I will simply rely on the Authorization header or cookie (I personally prefer the former) and ignore the token in the message. Requiring that I client sends it to be spec compliant isn't such a big issue IMHO. 

That train of thought yields the following conclusions:

- Clients sending _messages_ to a server, regardless of transport, must send along the token (sibling of context)
- Servers must implement HTTP security (Authorization header and cookie) for the REST and WS transports
- Servers may implement parsing of the token inside the message, and this is required if the server supports non-HTTP messaging protocols.

tkurki: Let me reiterate: requiring that tokens are included in every message is not backwards compatible. Current clients that authenticate with header or cookie will break. @rob42 and @fabdrol, do you disagree with that?

rob42: Ok, I see what you mean now. The spec does not define `token` as a possible field in messages, but https://signalk.org/specification/1.3.0/doc/security.html says `it can be included` etc. So it must be optional ( I will update put,get,*subscribe).

It should say something like _Clients using auth tokens should also send the token in every message in the 'token' key_

tkurki: >They could be multiplexed from another server or proxy,
>Its not safe to assume that all messages have the same auth, hence token

I think you are coming up with new requirements. Being able to discern the origin of individual messages is a valid use case, but not one that can be solved with simple tokens. If Fabian's boat gets a token from Rob's server in exchange for presenting username/password, then sends a message to Rob's server with the proper token, then Rob's server forwards the message to my server I have no way to know where the message came from, just that the token fails validation on my server.

I am totally ok with working out a solution for  arbitrary multiplexing and forwarding of messages from multiple authors, but let's do it properly and try to make the token something it is not.

gitbook-docs/security.md
```diff
@@ -8,8 +8,16 @@ with authentication and access control mechanisms in place.
 
 ## Authentication
 
-Authentication for Signal K REST and WebSockets connections is based on HTTP cookies or tokens carried in the HTTP
-header.
+Authentication for Signal K connections is based on a token carried in the message, or in a cookie for a HTTP
+request. The tokens can be of any type, but are typically JWT tokens.See https://jwt.io/
```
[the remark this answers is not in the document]
rob42: removed refs

gitbook-docs/security.md
```diff
@@ -8,8 +8,16 @@ with authentication and access control mechanisms in place.
...
+
+There are 3  authentication actions:
+
+* authenticate - login and obtain a token
+* logout - invalidate a token
+* validate - validate a token with auto-renewal if valid. 
+
+All 3 actions can be done via HTTP REST semantics for web based clients, or by sending signalk messages for others.
```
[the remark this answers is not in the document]
rob42: done

gitbook-docs/security.md
```diff
@@ -8,8 +8,16 @@ with authentication and access control mechanisms in place.
 
 ## Authentication
 
-Authentication for Signal K REST and WebSockets connections is based on HTTP cookies or tokens carried in the HTTP
-header.
+Authentication for Signal K connections is based on a token carried in the message, or in a cookie for a HTTP
```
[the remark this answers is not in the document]
rob42: done

gitbook-docs/security.md
```diff
@@ -80,25 +89,31 @@ If the login fails, the server will send a response like the following:
   "result": 401
 }
 ```
+The `result` codes are the same as normally used in HTTP
```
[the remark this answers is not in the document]
rob42: done

gitbook-docs/security.md
```diff
@@ -80,25 +89,31 @@ If the login fails, the server will send a response like the following:
...
 
 #### Web Based Clients
 
 Web based clients should be sure to include the cookie set in the authentication response in all subsequent requests.
 
-To logout, a web based client should send an HTTP PUT request to `/signalk/«version»/auth/logout`.
+```
+Cookie: SK_TOKEN=eyJhbGciOiJIUzUxMiJ9.eyJyb2xlcyI6IltcInNraXBwZXJcIl0iLCJpYXQiOjE1ND...
+
+```
```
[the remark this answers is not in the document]
rob42: Just added clarity.

gitbook-docs/security.md
```diff
@@ -80,25 +89,31 @@ If the login fails, the server will send a response like the following:
...
 
 #### WebSockets Clients
 
-Clients can include the authentication cookie with the initial request.
+Clients can include the authentication cookie with the initial request.
```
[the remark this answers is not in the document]
rob42: Ah no, cant remove that, its critical to the deeper meaning :-)

gitbook-docs/security.md
```diff
@@ -112,6 +127,93 @@ Clients using other kinds of protocols can include the `token` in the Signal K m
...
+The reply message will be returned in any case.
+
+#### Other Clients
+
+Clients using other kinds of protocols can send the following message.
+
+```json
+{
+  "requestId": "1234-45653-343454",
+  "login": {
```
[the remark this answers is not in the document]
rob42: changed to `validate`

gitbook-docs/security.md
tkurki: >the actual connection is just a tcp pipe

A WebSocket connection is not just a tcp pipe, it has additional semantics and mechanisms, for example the frame/message structure and per message compression. 

BTW we still have not defined a message delimiter for streaming deltas over byte streams and there is no checksumming in the spec for unreliable transports like serial.

So serial, tcp and ws are not quite so similar as you claim they are.

gitbook-docs/security.md
```diff
@@ -112,6 +127,93 @@ Clients using other kinds of protocols can include the `token` in the Signal K m
...
+```
+
+#### Reply Messages
+
+In both cases the reply will be
+
+On success:
+```
+{
+  "requestId": "1234-45653-343454",
```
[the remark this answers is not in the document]
rob42: fixed

gitbook-docs/security.md
tkurki: Apart from 
- parity with other protocols
- multiplexing, proxying and routing (better served with some kind of PKI infrastructure)
- easier server implementation in some cases
why exactly do you guys want the token in the message? What is a practical use case that is not possible or not very practical without it?

tkurki:
>We can just state that the token is a string, there's no need to specify what it looks like.

+1

gitbook-docs/security.md
```diff
@@ -112,6 +127,93 @@ Clients using other kinds of protocols can include the `token` in the Signal K m
...
+  "state": "COMPLETED",
+  "result": 401
+}
+```
+
+### Logout
+
+#### Web Clients
+To logout, a web based client should send an HTTP PUT request to `/signalk/«version»/auth/logout` with the token in the cookie.
```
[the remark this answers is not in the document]
rob42: done

gitbook-docs/security.md
rob42: Added `expiry`, in seconds

gitbook-docs/security.md
rob42: Modified

rob42:
Need to add the new messages to the schema.

gitbook-docs/security.md
rob42: @tkurki Sorry, had an old version of the page, didnt see those last replies.

Tcp, udp, serial, and ws are all the same above the transport layer. Basically you get/send a stream of signalk messages. Sure they are different underneath, but that should be transparent to the messages, and hence the processing. Compare that to the HTTP REST interface, thats quite different.

By not including the the token in the WS messages it makes that a special protocol too, and it doesnt have to be, eg their is no benefit. By making it different its more complex to deal with, and more difficult to learn signalk.

rob42: Having the token in the message makes processing stateless, which is a benefit for any sort of processing chain.

tkurki: The primary use case for WebSockets is browser based applications. There the vast majority of applications use cookie or header based authentication, not in-band tokens. WebSocket libraries have typically support for dealing with these mechanisms, both client and server side. I find it more likely that the developer is familiar with http-based authentication mechanisms than that she is coming from tcp/udp/serial background. I have the opposite opinion re: complexity and difficulty in learning: having to send the same token over and over again over a channel that was already 

There is also value in not needing to verify each and every message once the connection is established. 

Sending the token in band over unsecure channels makes it more vulnerable.

I know you're going to disagree with me, but all connectionful protocols could technically work like ws: for example in tcp you could send the token once and all subsequent messages would use that authorization.

BTW we have not defined what happens if an in-band token is not valid. Is the connection broken? Can a client keep streaming data with invalid tokens for ever? Update messages have no response, so there is no mechanism to inform the client (sender) that the token is not accepted.

What does `should` mean in the spec here? Does it work without the in band token  it or not? If the token is optional I don't think `should` makes sense and I would like to remove this additional statement.

gitbook-docs/security.md
```diff
@@ -66,6 +74,7 @@ If the login is successful, the server will send a response like the following:
   "state": "COMPLETED",
   "result": 200,
   "login": {
+  	"type": "JWT",
```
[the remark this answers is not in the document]
rob42: Its removed now

gitbook-docs/security.md
rob42: >Sending the token in band over unsecure channels makes it more vulnerable
Im not suggesting that. And if the channel is insecure then reading the http cookie or header is easy too.

>There is also value in not needing to verify each and every message once the connection is established.
When you implement per key security you will find you need the key deep in the server. If you are processing messages in a stateless async pattern you will either have to inject the token or roles into messages, or provide some sort of lookup back to the ws session. Everything is much simpler when the token is in the message

>all connectionful protocols could technically work like ws.
You assume that a single connection equals a single client. Thats a simplistic assumption based on a web browser centric view. It may well be the case for ws, but definitely not for tcp. Even in ws and a single browser I can see the case for a low-permission token for most stuff, and a short-lived high-permission token for admin changes. A bit like `sudo` in linux so you dont  have to login as root for everyday tasks.

gitbook-docs/security.md
```diff
@@ -44,9 +52,9 @@ In response to invalid login information the server must return HTTP error code
 
 If the server does not implement this authentication mechanism it must return HTTP error code 501 (Not Implemented).
 
-### Authentication via WebSockets and Similar Transports
+### Authentication via WebSockets, Tcp, and Similar Transports
```
mairas: Nitpicking: TCP is all capital acronym.

gitbook-docs/security.md
mairas: Disclaimer: I might've missed an argument or two.

I don't think including tokens in each message would be a good idea. That would mix the metadata and the actual payload.

One practical consideration would be that the servers would need to know when to drop the token from the messages. If a read-only client receives a message with a token, it could just reuse the token and start sending messages to the server.

gitbook-docs/security.md
```diff
@@ -89,27 +89,23 @@ If the login fails, the server will send a response like the following:
...
+Web based clients should be sure to include the cookie or `Authorization` HTTP header obtained from the authentication response in all subsequent requests.
 
 #### WebSockets Clients
 
-Clients can include the authentication cookie with the initial request. 
+Clients can include the authentication cookie with the initial request.
 
 Clients can include the `Authorization` HTTP header with the initial connect request. The format of the header should
-be `{type} {token}`, for example `Authorization: JWT eyJhbGciOiJIUzI1NiIsI...ibtv41fOnJObT4RdOyZ_UI9is8`
+be `BEARER {token}`, for example `Authorization: BEARER eyJhbGciOiJIUzI1NiIsI...ibtv41fOnJObT4RdOyZ_UI9is8`
```
mairas: Only the first letter of `Bearer` should be capitalized.

mairas: Also, this is a breaking change wrt. current implementation but I think the current implementation is non-conforming, so I don't have an issue with that.

gitbook-docs/security.md
rob42: A signalk _message_ is stateless/asynchronous, hence the _message_ needs the token so that it can be processed asynchronously. A signalk http/ws connection is stateful/synchronous, so it can be established with a header/token and the token can be injected at the server. 

If the server only supports connection level security then sending the token once is enough as its held at the connection, where security is applied. If you only consider simple direct connect networks (http/ws/tcp) then the token works at a connection level. But it doesnt work for other communications. What about a proxy gateway, MQTT, UDP, messages aggregated by a mux and forwarded, and many other scenarios. 

When security is at the key level (aka artemis server) each processing step needs reference to the token to apply security.

gitbook-docs/security.md
rob42: nits picked

gitbook-docs/security.md
rob42: Changed

gitbook-docs/security.md
rob42: > One practical consideration would be that the servers would need to know when to drop the token from the messages. If a read-only client receives a message with a token, it could just reuse the token and start sending messages to the server.

The server should never simply echo messages as that opens all sort of exploits. At the very least it should remove or replace the token.  So thats really an implementation issue, but worth mentioning for clarity.

rob42:
Are we ready to commit?

sbender9:
My only issue is the requirement to include the token in every websocket message.

rob42:
I removed that for ws.

gitbook-docs/security.md
```diff
@@ -112,6 +126,92 @@ Clients using other kinds of protocols can include the `token` in the Signal K m
 }
 ```
 
+### Token Validation
+
+Tokens may have a short expiry time and need to be renewed periodically, or a token's current validity may be unknown. 
+
+#### Web Clients
+To validate a token, a web based client should send an HTTP GET request to `/signalk/«version»/auth/validate` with the token in the cookie.
+If the token is valid, it will be renewed, and the new cookie set in the response.
```
tkurki: _it will be renewed_ sounds like the existing token will be renewed.

gitbook-docs/security.md
```diff
@@ -112,6 +126,92 @@ Clients using other kinds of protocols can include the `token` in the Signal K m
 }
 ```
 
+### Token Validation
+
+Tokens may have a short expiry time and need to be renewed periodically, or a token's current validity may be unknown. 
+
+#### Web Clients
+To validate a token, a web based client should send an HTTP GET request to `/signalk/«version»/auth/validate` with the token in the cookie.
```
tkurki: We should allow for token in header here as well, not just cookie.

gitbook-docs/security.md
```diff
@@ -112,6 +126,92 @@ Clients using other kinds of protocols can include the `token` in the Signal K m
 }
 ```
 
+### Token Validation
+
+Tokens may have a short expiry time and need to be renewed periodically, or a token's current validity may be unknown.
```
tkurki: `expiry` is now available, so in what case is the validity unknown?

gitbook-docs/security.md
```diff
@@ -66,6 +77,7 @@ If the login is successful, the server will send a response like the following:
   "state": "COMPLETED",
   "result": 200,
   "login": {
+  	"expiry": 86400,
```
tkurki: How about `timetolive` or something along that line? Would more intuitivite to me.

gitbook-docs/security.md
```diff
@@ -8,8 +8,19 @@ with authentication and access control mechanisms in place.
...
+__Note__: a signalk server should never simply echo or redistribute a message received without removing or replacing the token. 
+That would result in A's token being sent to B, which allows the B to impersonate A.  
+
+There are 3  authentication actions:
+
+* authenticate - login and obtain a token
+* logout - invalidate a token
+* validate - validate a token with auto-renewal if valid. 
+
+All 3 actions can be done via HTTP REST semantics for web based clients, or by sending Signal K messages for others.
```
tkurki: Clients may be sensors and still use http, no web involved, and there isn't anything particularly REST about the mechanism. I suggest _All 3 actions can be done via HTTP requests or by sending Signal K messages for non HTTP clients._

gitbook-docs/security.md
```diff
@@ -112,6 +126,92 @@ Clients using other kinds of protocols can include the `token` in the Signal K m
...
+
+The reply message will be returned in any case.
+
+#### Other Clients
+
+Clients using other kinds of protocols can send the following message.
+
+```json
+{
+  "requestId": "1234-45653-343454",
```
tkurki: No `requestId` in http responses, right?

gitbook-docs/security.md
```diff
@@ -112,6 +126,92 @@ Clients using other kinds of protocols can include the `token` in the Signal K m
...
+  "state": "COMPLETED",
+  "result": 401
+}
+```
+
+### Logout
+
+#### Web Clients
```
tkurki: HTTP Clients

gitbook-docs/security.md
```diff
@@ -80,25 +92,27 @@ If the login fails, the server will send a response like the following:
...
 
 #### WebSockets Clients
 
 Clients can include the authentication cookie with the initial request.
 
 Clients can include the `Authorization` HTTP header with the initial connect request. The format of the header should
-be `{type} {token}`, for example `Authorization: JWT eyJhbGciOiJIUzI1NiIsI...ibtv41fOnJObT4RdOyZ_UI9is8`
+be `Bearer {token}`, for example `Authorization: Bearer eyJhbGciOiJIUzI1NiIsI...ibtv41fOnJObT4RdOyZ_UI9is8`
+
+Note that clients with a token can also include the `token` key in every signalk message sent over the websocket connection
```
tkurki: Sorry about being a pest, but what does _can include a token_ exactly mean? A client _can_ include all kinds of data in the messages, but they won't have any effect unless specified here.

gitbook-docs/security.md
rob42: Changed

gitbook-docs/security.md
rob42: changed

gitbook-docs/security.md
rob42: Its possible for a token with valid expiry to be invalid for server-side reasons. eg The artemis server invalidates all tokens after a reboot, because the signing key is regenerated each boot.

gitbook-docs/security.md
rob42: now `timeToLive`

gitbook-docs/security.md
rob42: changed

gitbook-docs/security.md
rob42: No, the request/response semantics return the requestId in http too (I thnk) Then the http client can check request status for long running requests that are PENDING

gitbook-docs/security.md
rob42: done

gitbook-docs/security.md
rob42: Removed, so we can get this merged.

gitbook-docs/security.md
```diff
@@ -112,6 +124,93 @@ Clients using other kinds of protocols can include the `token` in the Signal K m
 }
 ```
 
+### Token Validation
+
+Tokens may have a short expiry time and need to be renewed periodically, or a token's current validity may be unknown. 
+
+#### HTTP Clients
+
+To validate a token, a web based client should send an HTTP GET request to `/signalk/«version»/auth/validate` with the token in the cookie, or in the header.
```
tkurki: Rereading this one more time...shouldn't this be a POST?

gitbook-docs/security.md
```diff
@@ -31,11 +42,11 @@ The client may send the login request with a `Content-Type` of `application/json
 ```
 
 In response to a valid login, the server shall respond with a 200 (OK) status, set an HTTP session cookie and include
-the token type and the token value in the body of the response. The response `Content-Type` must be `application/json`.
+the token expiry in seconds and the token value in the body of the response. The response `Content-Type` must be `application/json`.
 
 ```json
 {
-  "type": "JWT",
+  "expiry": 86400,
```
tkurki: Should be timeToLive.

gitbook-docs/security.md
tkurki: Leave the unknown validity out, as it does not have any bearing on either server or client implementation as far as I can tell?

tkurki:
@rob42 the token type change is not backwards compatible with the security spec in 1.3.0. I think it should be rewritten so that it is.

gitbook-docs/security.md
rob42: There is no body since the data is in the header hence a I used GET. Is POST the correct way?

gitbook-docs/security.md
rob42: It is in commit 6047ff6 above

tkurki: That changes line 80, this is line 49...

gitbook-docs/security.md
tkurki: IMHO security related requests should not be GETs so that they are less susceptible to XSRF.

fabdrol:
@rob42 there are some more changes required here. Could you have a look and update?

fabdrol:
Looks like all changes have been made, except the GET thing? Is it correct to remove the type of token and default to `Bearer` - because then I need to modify that in `signalk-js-client`.

rob42:
I thought I changed GET to POST? Let me check my commits

rob42:
Yes its changed. Its in commit d7dd6f0 above

fabdrol:
Okay, @tkurki does this resolve your review? Please comment and merge if yes :)

tkurki:
Yay, progress!

The schema file still uses `expiry` and none of the test samples use `timeToLive` anywhere, so maybe these should be corrected?

fabdrol:
Looks like it's resolved, is this ready to merge @rob42 ?

PR #536, opened 2019-01-25
Fix PUT default context and warn about paths

rob42:
Fix PUT default context and warn about paths.  The default context of `vessels.self` made it very difficult to distinguish between `resources', `sources`, and other root level branches we may grow. While it can be done at the server end it creates complexity and is error prone.

The default is removed, always use full path


rob42:
Oh damn - sucked in the unmerged swagger changes....

sbender9:
I think this PR was merged prematurely. I disagree with the removal of the default. And it also contains unrelated changes. I assumed you were going to remove those commits since you clearly knew they were in there.

rob42:
Just realised the initial openapi.json commit was there too. I thought it was just the update, which was simply an improvement. 
The removal of the default produced no comments, so i merged it.
I will revert the pull and create a new one

PR #538, opened 2019-01-31
Attempt 2-Fix PUT default context and warn about paths

rob42:
See #536 #537 
Created a clean pull.
Fix PUT default context and warn about paths. The default context of vessels.self made it very difficult to distinguish between resources', sources`, and other root level branches we may grow. While it can be done at the server end it creates complexity and is error prone.

The default is removed, always use full path, which is also consistent with GET


rob42:
Any objections to a merge this time?

tkurki:
I do not understand how context defaulting to `vessels.self` makes things difficult. Everything you can do with a required context you can do with a missing context property defaulting to vessels.self.

What am I missing here?

tkurki:
To expand on my previous answer: even if omitted context defaults to `vessels.self` you can explicitly use an empty context value to effect resources under the root.

rob42:
Set context to blank? How do you do that exactly in a url?
In an HTTP PUT the url contains the path-to-leaf, and the body is the leaf. So if I want to set `/signalk/v1/api/resources/waypoints/...` with a PUT, where exactly do I set the blank context?

Even if I could send or set it, or if I use lots of clever code to test/correct it at the server end, why make PUT behave in such an arcane way, when we dont do it anywhere else in signalk, eg GET, etc? Its confusing, error prone, and adds complexity for no benefit that I can see

The path problem is completely solved and consistent with the rest of signalk by adding `vessels.self` into the PUT path, which is trivially simple to do.

tkurki:
>In an HTTP PUT the url contains the path-to-leaf, and the body is the leaf. So if I want to set `/signalk/v1/api/resources/waypoints/...` with a PUT, where exactly do I set the blank context?

Nowhere. `context` and `path` are relevant only in delta, not when you do http PUT.

rob42:
This change is about the removing the default `vessels/self` in an HTTP PUT, not in the PUT over ws.
Ref Scotts comment "I disagree with the removal of the default." in #536

tkurki:
This is how `put.md` looks like in the current master. 
image

There is no `context` in http PUT, because everything is specified in the URL, so there is no default to remove.

Your change removes the only paragraph that talks about the default and it is under WS.

rob42:
Hmm, put.md in master is correct. thats what I intended. Somehow got confused with this during  https://github.com/SignalK/freeboard-sk/issues/70, and #536.

PR #541, opened 2019-03-14
Add Resources section

panaaj:
As per issue#540, the specification could benefit from the addition of a Resources section to describe the way in which to work with these data items. 

The initial content outlines the proposed creation, updating and deletion of resources and also proposes some basic filters to use when retrieving resources to limit the number of records returned.

As there is no documented way of supplying parameters to a Delta message I have taken a leap and proposed a `params` attribute for a `get` delta message.

I have referenced and aligned with other parts of the specification where possible.
Also tried to align POST and PUT requests with common use patterns.

There probably needs to be some text around security and accessing individual resource records as well.

### DIFF
--- a/gitbook-docs/resources.md
+++ b/gitbook-docs/resources.md
@@ -1,0 +1,346 @@
+# Resources
+
+Resources are collections of objects that are used to represent data that serves as additional information to aid with navigation, etc. 
+
+Resources are:
+- Usually persisted in a non-volatile data store _(i.e. not lost when server re-starts)_
+- Potentially large in number and / or record size
+- Able to be created, updated and deleted by both applications and server processes
+- A top level path in the Signal K API: `/signalk/v1/api/resources/*`
+
+Resource entries:
+- Of a specific type are grouped under a specific path _i.e. `/signalk/v1/api/resources/waypoints`_
+
+- Are identified by a __uuid__. _e.g. `urn:mrn:signalk:uuid:36f9b6b5-959f-46a1-8a68-82159742aadd`_
+
+- Originating from other vessels will have a `$source` attribute that contains the id of the source vessel.
+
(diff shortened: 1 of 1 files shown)


panaaj:
Resources can link / reference to other resources e.g. Route `start` and `end` link to `waypoint` records.
If a route is deleted should the associated a `start` and `end` waypoints be deleted as well?

rob42:
> Resources can link / reference to other resources e.g. Route `start` and `end` link to `waypoint` records.
> If a route is deleted should the associated a `start` and `end` waypoints be deleted as well?

I think this is implementation specific. Obviously you should not delete a resource that is referenced from elsewhere, but I dont think thats a part of the signalk specification. The spec should say that " deletion is attempted and may fail, eg if the resource is referenced from elsewhere, or permissions etc. See request/response"

panaaj:
Should then the http parameters format align to make them similar to stream api?
e.g. ?filter=geo{radius:1000}

rob42:
Is that valid http?  Probably `geo.radius=1000&geo.etc` would be more std

panaaj:
So parameters will likely refer to both generic filters e.g. geo and resource attributes e.g. note.region.
Where filtering on a resources attribute is required is t worth namespacing these with attr (e.g. attr.region=xxx) as not all resources have a consistent set of attributes?

fabdrol:
@panaaj @rob42 what's the status of this, more work required? If so @rob42 can you start a review?

panaaj:
Given the intent of this PR was to have resources represented in the spec I have removed the section that talked about the ability to filter returned records as this needs more consideration.

There are clearly implementation considerations as well as specification additions to be made to add entry filtering, delta parameters, etc so rather than hold up this PR I have removed them.

@fabdrol @rob42

PR #542, opened 2019-03-23
Add Apps API page

rob42:
Documenting the `signalk/v1/apps` API

### DIFF
--- a/gitbook-docs/apps_api.md
+++ b/gitbook-docs/apps_api.md
@@ -11,7 +11,7 @@ and a connection to a Signal K server which supplies json data via the Signal K 
 
 ## API root
 
-The Aps API root is `signalk/v1/apps/`
+The Apps API root is `signalk/v1/apps/`
 
 ## List
 
@@ -19,11 +19,12 @@ The Aps API root is `signalk/v1/apps/`
 It returns a json array in the form:
 ```
 [{
-//mandatory
+//mandatory.
+// [server_apps_root] is the location of installed apps from server html root.
 	"name": "Kip",
 	"description": "Really cool instrument app!!",
-	"location": "../@mxtommy/kip/package/public/index.html",
-	"_location": "../@mxtommy/kip/package/public/index.html", //DEPRECATED
+	"location": "[server_apps_root]/@mxtommy/kip/package/public/index.html",
+	"_location": "[server_apps_root]/@mxtommy/kip/package/public/index.html", //DEPRECATED
 // if available include manifest.json
 	"manifest": {..content of manifest.json..}
 //recommended
@@ -115,7 +116,7 @@ It returns a json array in the form:
 `signalk/v1/apps/install?appName=@signalk/freeboard-sk&appVersion=1.3.1`, with params selected from above. This will install the app into
 the preferred location on the Signal K server, from which the `list` call can be used to generate the app menu.
 
-# Remove
+# Uninstall
 
-`signalk/v1/apps/remove?appName=@signalk/freeboard-sk`, with params selected from `list`. This will remove the app.
+`signalk/v1/apps/uninstall?appName=@signalk/freeboard-sk`, with params selected from `list`. This will uninstall the app.
 
--- a/gitbook-docs/apps_api.md
+++ b/gitbook-docs/apps_api.md
@@ -24,6 +24,8 @@ It returns a json array in the form:
 	"description": "Really cool instrument app!!",
 	"location": "../@mxtommy/kip/package/public/index.html",
 	"_location": "../@mxtommy/kip/package/public/index.html", //DEPRECATED
+// if available include manifest.json
+	"manifest": {..contect on manifest.json..}
 //recommended
 	"licence": "",
 	"author": "Thomas St.Pierre <thomas@samoht.ca>",
--- a/gitbook-docs/apps_api.md
+++ b/gitbook-docs/apps_api.md
@@ -1,0 +1,119 @@
+# Apps API
+
+When a Signal K server provides webapps (rather than just data) it SHOULD implement the Apps API. This provides a standard way to list installed webapps, 
+search for new webapps, and install or remove them. 
+
+This allows different server implementations to share webapps, which gives webapps wider markets, and reduces 'brand' dependency. Basically any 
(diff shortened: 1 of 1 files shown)


rob42:
Freeboard-sk loads your choice of instruments in a sidepanel. It needs to be able to get name, url. I would like to see KIP do the same in reverse, so you create an MFD by inserting apps into panels.

I included search, install, remove so that they could potentially offer options that were not currently installed.

sbender9:
Got you. Like it! And now I want to do this in WilhelmSK ;)

sbender9:
Can we add some kind of “status” to the data from /list? Maybe with values like `available`, `installing`, ‘installation failed’?

Maybe `installed/restart required` in case the server need restarted after installation? Or maybe I just need to fix node server so restarts on required after install :(

rob42:
That supposes the install process has a well-defined process and outcome and the wrapper can store data on that somewhere. Outside of node/npm, install is just download the archive, unpack, and search for the index.html (all over the place..). No real way to know if its ok. 

I create the `list` output by traversing the install dirs and parsing package.json. So if its there its listed. I create the `search` list by a search and then mark the installed ones, which allows me to add a 'Remove' button. But thats server specific really, adding extra data to the reply json

rob42:
given the mess of current webapps package formats I think that needs sorting first

panaaj:
Some added value to consider (not mandatory though) is meta-data that can allow inclusion of app icons, etc to be displayed when listing apps... for example [#629](https://github.com/SignalK/signalk-server-node/issues/629)

rob42:
Added the manifest.json if it exists
`"manifest": {..content of manifest.json..}`

sbender9:
@rob42 I don’t see the issue with status. Node server provides this already. Maybe make it optional in case the server can’t provide it?

rob42:
Theres no problem sending additional data thats implementation dependent. Its especially useful for a servers app management ui. When the install is just unzip, if its there status=INSTALLED, there is no FAILED state,  just a failed install, eg there is no way to tell its installed but not working. So I dont see how it would be useful.

rob42:
> Maybe installed/restart
Since a webapp is just a bunch of static files  a reboot seems a bit extreme 😮

sbender9:
Lots of things could fail when installing one of these. No internet. Missing dependencies. No disk space.

I think letting the user know that the install failed would be very useful. And like we do with node server, actually show the user information as to why it failed would be great too. 

I think providing a way for users to install web apps and not giving them any way to see the progress and results of the install is short sighted.

rob42:
I think you misunderstand me. Definitely things can fail in the install, and showing progress is useful, but I see that as part of the install process, which is implementation dependant. This api is just to find installed apps, search for apps, and being able to request an install or remove is just a conveinience

How would you structure feedback from the install?

rob42:
@sbender9 ?

fabdrol:
Would it be out of scope to add to this that a server _may_ provide the apps via the various HTML5 discovery protocols of the MFDs (i.e. Garmin OneHelm & Navico's implementation)?

gitbook-docs/apps_api.md
```diff
@@ -0,0 +1,121 @@
...
+## List
+
+`signalk/v1/apps/list` lists the currently install webapps. It is typically used to create app menus, and manage installed app lists.
+It returns a json array in the form:
+```
+[{
+//mandatory
+	"name": "Kip",
+	"description": "Really cool instrument app!!",
+	"location": "../@mxtommy/kip/package/public/index.html",
```
fabdrol: @rob42 when implementing a client I'd prefer an absolute path from the root of the server. @sbender9 your thoughts from Wilhelm SK perspective?

gitbook-docs/apps_api.md
```diff
@@ -0,0 +1,121 @@
...
+....
+]
+```
+
+## Search
+
+`signalk/v1/apps/search?keyword=signalk-webapp` finds the currently available webapps from npm. It is typically used to list available apps for installation.
+By default it makes a call to `https://api.npms.io/v2/search?size=250&q=keywords:signalk-webapp`
+It returns a json array in the form:
+```
```
fabdrol: Shouldn't we include pagination here as part of the search request? I.e. `signalk/v1/apps/search?keyword=foo&skip=0&limit=25`? If things grow that may be necessary

gitbook-docs/apps_api.md
```diff
@@ -0,0 +1,121 @@
...
+
+# Install
+
+`signalk/v1/apps/install?appName=@signalk/freeboard-sk&appVersion=1.3.1`, with params selected from above. This will install the app into
+the preferred location on the Signal K server, from which the `list` call can be used to generate the app menu.
+
+# Remove
+
+`signalk/v1/apps/remove?appName=@signalk/freeboard-sk`, with params selected from `list`. This will remove the app.
```
fabdrol: `uninstall` makes more sense than remove (as we use `install` rather than `add`)

rob42:
> Would it be out of scope to add to this that a server _may_ provide the apps via the various HTML5 discovery protocols of the MFDs (i.e. Garmin OneHelm & Navico's implementation)?

I think that falls under implementation. The implementation is free to do whatever, this is about  _signalk_ webapps.

gitbook-docs/apps_api.md
rob42: done

gitbook-docs/apps_api.md
rob42: Lets add that when we need it. If we add pagination etc it should be generic to all the signalk API's. That includes pagination, ordering,  search hits, max size, lots of stuff etc

fabdrol:
> I think that falls under implementation

@rob42 I agree that the _how_ falls under implementation, but way could promote this ourselves. This is an area where Signal K could really come to its own...

tkurki:
I agree with Rob. 

I don't quite see how saying that a server may support manufacturers' propietary system X (where each manufacturer has their own prop way of doing things) is useful, even if X is something that Signal K does / could do well.

A server may support all kinds of things, the spec should concentrate on Signal K.

fabdrol:
Okay, in that case shall we merge?

rob42:
Yes

Issue #539, opened 2019-03-04
Issue with Notes


irandrews:
We have a cloud database of around 15,000 'notes' (growing by 3,500 a year). Our 6,000 members share these notes in real time using an existing App and Website. We would like to (actually have done!) make these notes available via SignalK so that the range of clients can be expanded. 

We can get the system to work (see the demo) but the current notes schema has a couple of limitations for us, and, I suspect, others. Specifically:

**Location**. Typically our notes are about Harbours and Anchorages. The first person to make a note about somewhere picks an arbitrary GPS point somewhere in the harbour and creates a note. Other users can add their own views to this location and these are all displayed on the screen at the same time. The current schema has no way to relate notes to each other - and specifically for the client to tell the server whether the note is about an existing location, or a new one. Unfortunately the 'name' field (which we use as the port name) is not unique, and the 'position' field only has to have a minor resolution difference for it to become a new location. Currently we are overloading the uuid field to provide this info, but this is against the schema. One solution might be for each note to have a pointer (like it does for 'region') to a named 'location' - aka a Waypoint - but we don't want to overload Waypoints as these can be used for different purposes and appear differently (i.e. in a different colour) on a screen. One possibility would be to create a new class of entity: 'location' (which has a GPS reference) or another would be to add a 'type' tag to the Waypoint entity so the client could determine what it relates to. I wouldn't want to remove 'position' from the note though. 

**Privileges**. In our world users can edit/delete their own notes, but not one created by someone else. The current schema has no way for the server to tell the client what is editable and what is not, so that the client can put the relevant buttons on the screen. I favour something like a 'readonly' tag added to the note schema.

rob42:
Re Location:
You can use one of 3 options to 'locate' a note:
```
           "region": {
              "type": "string",
              "description": "Region related to note. A pointer to a region UUID. Alternative to position or geohash"
            },
            "position": {
              "description": "Position related to note. Alternative to region or geohash",
              "$ref": "../definitions.json#/definitions/position"
            },
            "geohash": {
              "description": "Position related to note. Alternative to region or position",
              "$ref": "../definitions.json#/definitions/geohash"
            },
```
A position is a specific lat/lon, and as you say, a few decimals and its a different entry. This is where  a geohash is useful. See https://en.wikipedia.org/wiki/Geohash. Basically the geohash is a single string that identifies a rectangle, the more chars the smaller area. So a search for less chars is like a zoom-out. 
[...]

PR #544, opened 2019-04-08
Feature: Additional Note attributes

panaaj:
This PR looks to address Notes schema limitations outlined in Issue #539 (@irandrews)

The current Notes schema facilitates the following:

- Create a Note at a geographic location
- Create a Note at a geohash
- Attach Note to a Region uuid in `/resources/regions`

Although `position`, `region` and `geohash` are not defined as `oneOf` they are described as alternatives to each other. This implies that one can not define a Note at a specific location within a `region`. 

Therefore the scenario where the area of a region is:

1. Large enough to contain multiple hazards, at various locations within the region AND 
2. One wants to place different Notes on specific hazards 

is not specifically supported with the current definitions.

Additionally there is no way to:
- Group Notes together regardless of position or region. 
- Define the author of a Note.
- Define whether the contents of the Note can be edited / changed.
- Define whether the note is available only on the host Signal K server or is shared with other servers / services.

__To facilitate these scenarios the following changes are proposed:__
1. Change the wording of the definitions of:
- `region` to be defined as _additional_ to `position` or `geohash`
- `position` and `geohash` to be defined as only being alternatives to each other

2. Add a `group` attribute to allow the definition of a group or collection to which the Note belongs

3. Add an `author` attribute

4. Add a `readOnly` attribute to show whether the content of the Note is permitted to be changed.

5. Add a `published` attribute, to show whether a Note is available to other Signal K servers or services.

---

Adding these additional attributes would:
- Facilitate the retrieval of notes that are part of a specific `group` by supplying the relevant parameter to `/resources/notes?group=<group_name>` _(as proposed in PR #541)_.

- Identify the author of a Note.

- Identify a Note as non-editable.

- Identify a Note as available outside of the host Signal K server.

### DIFF
--- a/schemas/groups/resources.json
+++ b/schemas/groups/resources.json
@@ -202,14 +202,14 @@
             },
             "region": {
               "type": "string",
-              "description": "Region related to note. A pointer to a region UUID. Alternative to position or geohash"
+              "description": "Region related to note. A pointer to a region UUID. Can be defined in additon to position or geohash"
             },
             "position": {
-              "description": "Position related to note. Alternative to region or geohash",
+              "description": "Position related to note. Alternative to geohash",
               "$ref": "../definitions.json#/definitions/position"
             },
             "geohash": {
-              "description": "Position related to note. Alternative to region or position",
+              "description": "Position related to note. Alternative to position",
               "$ref": "../definitions.json#/definitions/geohash"
             },
             "mimeType": {
@@ -220,6 +220,14 @@
               "type": "string",
               "description": "Location of the note"
             },
+            "group": {
+                "type": "string",
+                "description": "Name of a group or collection to which the note is attached"
+            }, 
+            "author": {
+                "type": "string",
+                "description": "Name of the creator of the note"
+            },                         
             "timestamp": {
               "description": "Timestamp of the last update to this data",
               "$ref": "../definitions.json#/definitions/timestamp"
--- a/schemas/groups/resources.json
+++ b/schemas/groups/resources.json
@@ -227,7 +227,15 @@
             "author": {
                 "type": "string",
                 "description": "Name of the creator of the note"
-            },                         
+            }, 
+            "readOnly": {
+                "type": "boolean",
+                "description": "Indicates whether the whether the content of the Note is permitted to be changed."
+            },
+            "published": {
+                "type": "boolean",
+                "description": "Indicates whether the note is available only for use on this server or shared with others."
+            },                                                
             "timestamp": {
               "description": "Timestamp of the last update to this data",
               "$ref": "../definitions.json#/definitions/timestamp"


schemas/groups/resources.json
```diff
@@ -202,14 +202,14 @@
             },
             "region": {
               "type": "string",
-              "description": "Region related to note. A pointer to a region UUID. Alternative to position or geohash"
+              "description": "Region related to note. A pointer to a region UUID. Can be defined in additon to position or geohash"
```
tkurki: typo in _addition_

schemas/groups/resources.json
```diff
@@ -220,6 +220,22 @@
...
+                "type": "string",
+                "description": "Name of a group or collection to which the note is attached"
+            }, 
+            "author": {
+                "type": "string",
+                "description": "Name of the creator of the note"
+            }, 
+            "readOnly": {
+                "type": "boolean",
+                "description": "Indicates whether the whether the content of the Note is permitted to be changed."
```
tkurki: double whether

schemas/groups/resources.json
```diff
@@ -220,6 +220,22 @@
               "type": "string",
               "description": "Location of the note"
             },
+            "group": {
+                "type": "string",
+                "description": "Name of a group or collection to which the note is attached"
+            }, 
+            "author": {
```
tkurki: Here I would copy the `contributors` structure from package.json:
- `authors` to allow multiple
- either objects with name/email/url properties (anyOf) or strings with suggested structure

https://flaviocopes.com/package-json/#contributors

Even if the current use case is note per author I have a hunch allowing multiple would be useful.

schemas/groups/resources.json
```diff
@@ -220,6 +220,22 @@
...
+                "type": "string",
+                "description": "Name of the creator of the note"
+            }, 
+            "readOnly": {
+                "type": "boolean",
+                "description": "Indicates whether the whether the content of the Note is permitted to be changed."
+            },
+            "published": {
+                "type": "boolean",
+                "description": "Indicates whether the note is available only for use on this server or shared with others."
```
tkurki: Can you explain this a bit more. The name of the property and the description don't quite connect for me. `Published` means that somebody published (emphasis on the past tense) this somehow somewhere. Do you mean that the note is `local=true` to this server?

schemas/groups/resources.json
```diff
@@ -202,14 +202,14 @@
...
+              "description": "Region related to note. A pointer to a region UUID. Can be defined in additon to position or geohash"
             },
             "position": {
-              "description": "Position related to note. Alternative to region or geohash",
+              "description": "Position related to note. Alternative to geohash",
               "$ref": "../definitions.json#/definitions/position"
             },
             "geohash": {
-              "description": "Position related to note. Alternative to region or position",
+              "description": "Position related to note. Alternative to position",
```
tkurki: As region (arbitrary geographic bounds), geohash (covering some area, small or large) and a precise location are different things I would be happy to change the wording to allow all of them.

schemas/groups/resources.json
tkurki: A more future proof way to implement this would be to add an optional enum property `accessType` with one value, `READONLY`. This way it is easier to extend.

rob42: Keep in mind that read/write/delete can only be controlled on the local server. Once a note (or any signalk object) is exported to another client/server you can no longer control editing. 

If the server provides per-key security the note can have permissions set the same as you would a file in linux, eg -rw-r---r--- (644) which achieves a more flexible result and is standard across all keys. Likewise -rw-r------ (640) would stop other users seeing the note unless in the proper group.

This can be communicated to the client via `meta._attr`, however any disallowed attempt to view/save the note would fail anyway. 

So I dont think we even need `accessType`

schemas/groups/resources.json
rob42: Same as above, better to control this withe per-key security. That gives fine-grain control over who can read/write the note.

schemas/groups/resources.json
rob42: Yes, allow all of them. They are basically all different ways of filtering by location, having all of them is a bonus. A server might even auto-calculate the others, eg a geohash for a location, or enclosing a region.

schemas/groups/resources.json
panaaj: I believe the intent of this attribute is to support the use case where a note is being drafted (incomplete) and is saved to the server but not ready for publication / upload / sharing.
I think `local=true` works here.

schemas/groups/resources.json
panaaj: If the client can get read/write status via a means like `meta._attr` then this would work. 
Is this a documented approach for developers to follow?

schemas/groups/resources.json
panaaj: Fixed

schemas/groups/resources.json
panaaj: Updated to indicate all are allowed.

schemas/groups/resources.json
panaaj: Updated property name to `authors` and aligned structure to package.json `contributors`.

schemas/groups/resources.json
rob42: hmm..it might be better to have a `validity` attribute, eg from and to dates. That would allow publishing notes in the future and removing them, eg 'Notice to Mariners - logs in water around xxx' or weather warnings etc

panaaj: So something like ..
```
validity: {
    start: <datetime>,
    end: <datetime>
}
```
where:
- NO **start** and  NO **end**= note is valid from time of creation
- DEFINED **start** and NO **end**= note becomes valid after **start** date/time has elapsed and will remain valid indefinitely
- NO **start** and DEFINED **end**= note is valid only until **end** date/time has elapsed
- DEFINED **start** and DEFINED **end**= note is valid only between the **start**  and **end** date/times

rob42: yes (`startTime` ,`endTime` for consistency maybe), and use it instead of `readOnly` and `published`. eg if its before `startTime` only the author can see/edit it. Use key-level security to control access when published. 

A background cron job can then auto publish/remove notes to suit.

irandrews: Unfortunately this doesn't support the 'save as Draft' functionality that `published` was intended to provide. The User doesn't want his note made valid until he physically goes in and marks it as so.

schemas/groups/resources.json
irandrews: The 'real' privileges for something like a note entity would be CRUD, not Linux permissions, so I favour the enum solution. However I can live with anything that allows the client to get read/write status from the server, preferably using a documented approach.

irandrews:
I am very keen for this PR to be adopted, as its absence is currently holding up my project. I can live with all the changes in the current PR. There are two open suggestions above which I’ve commented on.

schemas/groups/resources.json
rob42: The user could set a date in the future, even 09/09/2999 so the note would not be published. That could be triggered by a 'Draft/Publish' button ?

I'm not against the `published` key, just dont want unnecessary duplicated functionality, especially one-off for a specific key (`notes`). 
Implementing a server is complex enough now, the more we can use existing generic features the better.

schemas/groups/resources.json
rob42: Linux permissions basically provide CRUD per owner/group/others. See below - I am trying to avoid adding one-off solutions so that server implementations and security can be applied easily across all current and future keys.
The UNIX/Linux solution isnt perfect but for something that was invented 40 years ago, and is still viable in the internet age suggests its a pretty good design.

rob42:
So two remaining questions:
1) `validity`, using far future dates, does this suffice for draft/published/deleted ?
2) Any objections to replacing `accessType` with use of `meta._attr`?

panaaj:
With regards to `meta._attr` I have looked through the spec and I can't seem to find a reference to it.
the `data_model_metadata.md` document talks about the metadata attributes and setting them up globally but there is no mention of permissions meta data.
In node-server `./<some path>/meta._attr` is not recognised ( not sure about Artemis) but it begs the question how do we deal with an immediate use case in the short term?

tkurki:
>`validity`, using far future dates, does this suffice for draft/published/deleted 

Not to me. `draft` status related to publishing is semantically different from `validity`. Why would we want to overload validity with this special meaning? For example if I want to show notes in draft status differently in the UI or provide a list of notes in draft status we would need to define a magic value to represent drafts.

tkurki:
Something to consider: are some of the use cases and additions here specific to this @irandrews 's  use case or are they generic? Not everything has to be in the specification from day zero, we can write software that extends Signal K, figure out the details with field use, maybe redo some of the stuff and then expand the specification. Naturally there is no specification for everybody to refer to. But often additions created from a purely theoretical standpoint don't stand the test of real use, or need to be extended anyway to be usable.

schemas/groups/resources.json
irandrews: I take the point about duplication, and agree that a date range can be useful, but using a random future date for this is something of a distortion of the spec, since the functionality isn't implicit.
Also, the user will want to retrieve all 'draft' notes from the server, and this implies the client 'knowing' the magic date.  If the date is simply in the future, how does the client know the difference between 'draft' notes and 'finished' notes to be published at some point in the future?

panaaj:
For "use case" specific attributes is it maybe worth considering a mechanism to provide `extended attributes` in a similar way to gpx (they define <ext> to hold application / use case specific attributes) or GeoJSON  (they have the `properties` object)?

Since GeoJSON is used in a number of resources it's  kind of already in place.

tkurki:
Having a generic, non note specific method to represent access is something to consider, but we don't have one at this stage.

I disagree with adopting the Unix access model wholesale. Some of the features are not relevant without further specification: 
- what does execute mean? 
- there is no concept of `group` in SK spec at the moment
- what's a directory in SK data model? can you retrieve the access information for a key that you don't have any access to (no read/write)? does write access mean that you can remove or just empty a path?
- do we want to implement special cases like having write but no read access?

Even if we'd adopt the model as a whole we would still have to define the format to present this data in the API. I don't think encoding this in numbers or strings the way Unix-like utilities intended for human use makes sense, as programmatic access would require an extra step to interpret what the value means.

Again even if we'd adopt the model in the future as a whole there is nothing stopping us from providing an application developer friendly,  backwards compatible way to this information: what if we add `meta.accessType` enum with values `READ` and `WRITE`? This would allow an application to be informed about the access type the current user has for this resource. Even if we expand the access model in the future we could keep this API with very little overhead.

tkurki:
I was about to add something about generic key-value pairs for use case specific values, but Adrian beat me to it. I think it would be a good idea. [Already discussed](https://github.com/SignalK/specification/issues/539#issuecomment-471725779) in #539. Call them `properties` to be in line with GeoJSON.

rob42:
Definition for `_attr` is in https://github.com/SignalK/specification/blob/master/schemas/definitions.json
```
"_attr": {
      "type": "object",
      "title": "_attr schema.",
      "description": "Filesystem specific data, e.g. security, possibly more later.",
      "properties": {
        "_mode": {
          "type": "integer",
          "title": "_mode schema.",
          "description": "Unix style permissions, often written in `owner:group:other` form, `-rw-r--r--`",
          "default": 644
        },
        "_owner": {
          "type": "string",
          "title": "_owner schema.",
          "description": "The owner of this resource.",
          "default": "self"
        },
        "_group": {
          "type": "string",
          "title": "_group schema.",
          "description": "The group owning this resource.",
          "default": "self"
        }
      }
    },
```

rob42:
As Ive said before unix style permissions arent perfect but they have stood the test of time and are already in the spec. As always it can be implemented by a server any way that suits, this is just a rendition of the permissions for sharing. 

AFAIK key level security is not implemented in node server, but it is in the artemis server. I learned a lot about the issues and problems involved, and parsing them at high speed. Simply put, _it works_ for our needs. In application it works exactly the same as a Linux filesystem in what you can see and do. The use of `groups` is also an implementation issue. It just provides a way to identify a collection of users which is more than owner, and less than all.

BTW its `your.key.meta._attr`, so would be `...your/key/meta/_attr` in a GET.

rob42:
Seems we need a 'published' flag. Should that be  `status`, with ENUM `draft,published`

panaaj:
I have made the following changes.....
Replaced `draft` and `readOnly` attributes with a `properties` object to hold additional key/value pairs required by specific use cases.

So now `draft` and `readOnly` (if the use case requires them) can be placed in the `properties` object.

irandrews:
I think the permissions issue is so important it deserves more thought and its own PR. owner/group/world may have a role to play in the context of, say, SignalK being used to crowdshare weather or navigation data, (with other boats or chart/met suppliers) but I don't think a discussion about notes is the right place to make the decision.
For this PR, I think the `properties` block is a good solution.

schemas/groups/resources.json
panaaj: The inclusion of the `properties` block should cater for this.

schemas/groups/resources.json
panaaj: As per comments below, permissions should have a separate PR.

rob42:
`published` seems to have value, lets add it.  

Security was argued over endlessly in the past, the _attr is already in the spec. It doesnt need revisiting. Security and access to paths is critical to a viable server (at scale) but you _cannot_ control the data once it leaves your servers. Look at the whole DRM mess, it just doesnt work. Remember the implementation of security is  internal to the server, you can do it anyway you want.

So the only need for `_attr` is to provide a generic method to _signal_ permissions _beyond_ your servers, or between them, or for users to set key permissions in a generic way (from an app). The UNIX perms model is well understood, and works, we dont need to re-invent a custom version. 

`readOnly` has no practical use. Its equivalent to a 10yr old putting a "No burglars" sign on his bedroom door.

Before you flame me yet again consider how the `readOnly` flag will work. Most of you use the node server so AFAIK it only has web based global login. No key level security, aka no real security beyond a simple web app model at all.

I login to your server, load the note, see the `readOnly` flag and laugh, edit it anyway and PUT it back. What happens? Do you quickly add a bit of code for notes that checks `readOnly` and refuses to write? Excellent, the author can no longer write either. Another quick fix...cool a custom one off security for one aspect of notes. Now, about charts, autopilot, position...or if I inject a PUT message over ws or tcp..

Try this using the artemis server and the author (`owner`) sets -`rw-r--r--`  your PUT will fail because you dont have write permission. It wont work on charts or autopilot either. The same generic security works across all transports for all keys. Its there now - try implementing it in node and you will understand. The servers implementation is not important, just that it applies the permissions requested by the author.

If you really cant see past `readOnly` use it in your own implementation, but dont add it to the spec and expect others to implement it.

irandrews:
I don't mind how its done, but the purpose of the read/write flag on a note is to tell the client whether to put an 'edit' button on the screen next to the note. 
I thought we'd accepted that a 'properties' block will be the way to do this, so its out of the spec.

rob42:
..one use for the `meta._attr` on a note is to tell the client whether to put an 'edit' button on the screen next to the note.

panaaj:
Unless I am reading it wrong, it appears that there are no further changes required.

tkurki:
Some clarifications, I hope:
- Node server has per key security / access control lists that operate on Signal K paths, see https://github.com/SignalK/signalk-server-node/blob/master/test/security.js#L65-L93
- `meta._attr` is in the json schema, yes, but it is pretty underspecified. It is not documented in any way in the textual specification.
- Node server does not implement `meta._attr`

I think the best course of action here is that for the use cases at hand we adopt `meta._attr` as outlined by Rob above.

Personally I am not too crazy about using unix command line numeric representation in an API that is supposed to be self documenting and obvious to the most casual observer, but it is one commonly understood way to do this. As Rob says it is just way to communicate the permissions - we have no Signal K API to modify the permissions.

Further down the line we should clear the ambiguity related to `meta._attr` with additional documentation and maybe add more user friendly APIs. _Unix style permissions, often written in `owner:group:other` form, `-rw-r--r--`_ is too vague for my taste. Like, if you don't know this already where can you learn more? do we support setuid and setgid as well? how do the access levels translate to Signal K actions? is execute level relevant somehow?  etc.


That said this PR is mergeable to me.

tkurki:
Sorry, I was a bit hasty saying that this is in mergeable status. I was only reading the discussion, not the actual schema files.

Three problems:
- No test data - there is nothing that checks if the updated schema works the way we think and no full document to serve as an example on how notes are supposed to be used. A trap we have fallen into several times that I would not like to repeat. I suggest crafting one or more samples documents in test/data/full-valid/, they will be automatically validated in the tests.
- `properties` does not specify any structure, which makes it pretty useless for validation. Any document with anything in `properties` will fail validation. [Unknown property names](https://stackoverflow.com/questions/32044761/json-schema-with-unknown-property-names) are doable in JSON schema. I guess all types should be allowed as values.
- `groups` use is not documented, as it stands it is too vague to be usable. Earlier in this discussion I suggested that it should be handled like regions, with notes using id pointers, then the real group could have a name and a description

@panaaj sorry about change request this late in the game - mea culpa.

tkurki:
Now that I wrote `groups` I realised another problem: what if a note belongs to several groups? It should be an array of group ids I think. Again a nice sample document would go a long way to clearing things up here.

panaaj:
> * `properties` does not specify any structure, which makes it pretty useless for validation. Any document with anything in `properties` will fail validation. [Unknown property names](https://stackoverflow.com/questions/32044761/json-schema-with-unknown-property-names) are doable in JSON schema. I guess all types should be allowed as values.

I found this regarding the JSON schema and object properties:

> The additionalProperties keyword is used to control the handling of extra stuff, that is, properties whose names are
> not listed in the properties keyword. By default any additional properties are allowed.
> The additionalProperties keyword may be either a boolean or an object. If additionalProperties is a boolean
> and set to false, no additional properties will be allowed.

So it seems to be valid but I could add `"additionalProperties": true` for clarity.

panaaj:
> Now that I wrote `groups` I realised another problem: what if a note belongs to several groups? It should be an array of group ids I think. Again a nice sample document would go a long way to clearing things up here.

The way I view `group` is it represents the name of a collection of notes, and (while a note could technically be associated with more than one group) they generally attached to an item / object. 
A bit like a  `sticky note`...... you write on it, you tear it off and then you stick it to something. 

I think it notes with multiple groups might be over-complicating it a little but happy to be wrong!

I'm happy to draft something to show how notes can be used which will possibly clarify things.. (btw that is the motivation behind #541)

panaaj:
Also thinking if a note needs to be related to a position, area or region should the schema include a `oneOf [ {required: [..]} ]` for these properties?
```            "patternProperties": {
                "^urn:mrn:signalk:uuid:[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-4[0-9A-Fa-f]{3}-[89ABab][0-9A-Fa-f]{3}-[0-9A-Fa-f]{12}$": {
                    "type": "object",
                    "description": "A note about a region, named with a UUID. Notes might include navigation or cruising info, images, or anything",
                    "oneOf": [
                        {
                            "required": ["region"]
                        },
                        {
                            "required": ["position"]
                        },
                        {
                            "required": ["geohash"]
                        }                                                
                    ],
                    "properties": {
```

panaaj:
So then it is OK for a note not to be connected to anything _(position, geography, group, another resource, signalk path)_, basically just a `title` and / or `description`.

I have no problem with this.... it's just that notes generally relate to something.

panaaj:
I have completed the following:
1. created `gitbooks/resources_notes.md` as documentation for notes.
2. created `test/data/full-valid/reources-note-sample.json`
3. updated `schemas/groups/resources.json`

Is there anything else required to finalize this PR to be merged?
Does a link to the  `gitbooks/resources_notes.md` file need to be created or will this be done by the publishing process?

fabdrol:
@tkurki  @panaaj provided some more changes, does it resolve your review?

PR #545, opened 2019-04-16
fix: choice of the value of the notification state

cmotelet:
The purpose of this PR is to clarify the possible values for the states in notifications vs the states in zones.
- For a notification, the possible states are:
`[normal|alert|warn|alarm|emergency]` (deletion of nominal).
- For a zone, the possible states are:
`[nominal|normal|alert|warn|alarm|emergency]` (no change).

The file `gitbook-docs\data_model_metadata.md#alarm-management` was already updated in the documentation.

Please comment to find the best compromise.

### DIFF
--- a/schemas/definitions.json
+++ b/schemas/definitions.json
@@ -505,6 +505,19 @@
       }
     },
 
+    "notificationLevel": {
+      "type": "string",
+      "title": "notificationLevel",
+      "description": "The alarm level when a notification is raised.",
+      "default": "normal",
+      "enum": [
+        "normal",
+        "alert",
+        "warn",
+        "alarm",
+        "emergency"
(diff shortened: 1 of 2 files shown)


tkurki:
Hmm. Zones and `notifications` are connected in the spec: https://signalk.org/specification/1.3.0/doc/notifications.html#expected-implementation-behaviour

If we have zones that define normal and nominal zones they are treated like the more serious notifications: when the value is in that zone the notification is set. I don't think we have anywhere specified that `normal` and `nominal` are special cases and should not set `notifications` values.

So I don't think anything needs to be changed in the spec.

cmotelet:
cancelled following tkurki's explanations

PR #547, opened 2019-05-07
fix: allow merge meta user in signalk schemas

cmotelet:
This PR allows you to merge meta-users into an existing path in the signalk schema.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "SignalK specification schema as an npm module with tests",
   "main": "dist/index.js",
   "scripts": {
-    "test": "babel src --out-dir dist --copy-files && mocha",
+    "test": "babel src --out-dir dist --copy-files && mocha --timeout 10000",
     "prepublish": "npm run js:dist && npm run docs:keys",
     "js:watch": "babel --watch src --out-dir dist --copy-files",
     "js:dist": "babel src --out-dir dist --copy-files",
--- a/src/fullsignalk.js
+++ b/src/fullsignalk.js
@@ -35,10 +35,29 @@ function FullSignalK(id, type, defaults) {
     this.self = this.root.vessels[id];
     signalkSchema.fillIdentity(this.root)
     this.root.self = 'vessels.' + id
+    fillWithUserMetadata(this.self, this.root.self);
   }
   this.sources = {};
   this.root.sources = this.sources;
   this.lastModifieds = {};
+}
+
+function fillWithUserMetadata(obj, path) {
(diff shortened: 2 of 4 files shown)


rob42:
Can you please add some more detail about this? I dont understand what meta-users is exactly.

cmotelet:
When you use the function `FullSignalK(id, type, defaults)`, the last parameter `defaults` and may contain part of the schema but if this object contained `meta`, they were not displayed correctly (especially for the paths existing in the schema).
Look at my issue: https://github.com/SignalK/signalk-server-node/issues/754

rob42:
Ah, so this just adjusts the test code, not the spec definition itself?

cmotelet:
Yes, only the ability to inject metadatas and retrieve them correctly. No change in the spec definition.
And I have discovered an additional case that is not being handled properly, the correction is underway for tomorrow.

tkurki:
If I understand this correctly this will add some entries to metadataByRegex every time a FullSignalK object is created. `metadataByRegex` is a module level singleton, initialised from data that is derived from the static schema files. I think that is a bad idea: the current Node server creates a new FullSignalK object for every http request that queries the full data model. Furthermore what if I create two different FullSignalK objects with different user provided metadata - the paths will end up in the shared `metadataByRegex`.

cmotelet:
Yes, that's right, I forgot the multiple call to FullSignalK !
To fix, if I export a new function into `index.js` like `addMetaForPath(fullPathWithoutMetaStr, metaObj)`
and that this function builds the regex and adds it first to `metadataByRegex`.

In `signalk-server-node` at startup, we will iter in `defaults.json` to find the meta values if there are any.
If there is a meta value we complete them, they must be all there otherwise `addValue` in `fullsignalk.js` does not add any.

Then these metas are added only once in `metadataByRegex` with the new function `addMetaForPath(fullPathWithoutMeta, metaObj)`.
`fullPathWithoutMeta` must be of the full path type with the uuid of the vessel so as not to interfere with other regexes.

Do you agree with this new mode?

And another thing, add values to the schema, from `defaults.json` such as
design.[draft|length|beam|airHeight|aisShipType], sensors.gps.[fromBow|fromCenter] doesn't seem right to me without going through a delta.
This direct addition is not valid if you check with `validSignalK`.

fabdrol:
@cmotelet please implement your proposed changes so we can have a look if it resolves @tkurki 's review.

cmotelet:
closed due to bad implementation.

PR #548, opened 2019-05-31
feature: include all source information under `sources`

sbender9:
This is a change only the code use by node server to build the full tree.

Currently when building a the full tree from deltas, this code will not include things like `hardwareVersion` and `productName` under `/sources` when included in deltas. This changes makes so all extra fields under `source` for deltas are included.


fabdrol:
@sbender9 I don't understand what this change is for, could you provide a bit more detail?

sbender9:
I updated the description, let me know if that makes sense.

Issue #294, opened 2016-11-23
RFC 0005: Streaming values over a range of transports


pod909:
RFC 0005: Streaming values over a range of transports
Author: Ric Morris, pod909@gmail.com
Status; DRAFT
Version: 0.1
Date: 22/11/2016

This document contains ideas and content first suggested by Teppo Kurki, Paul Sumpner, Tim Mathews, Sailoog, Toqduj, Jim Boynes, Rob Huitema, Brian P and others.

### Summary

This document describes a consistent approach to communicating updates to individual values in the Signal K namespace over a range of transports.

Three types of service are described:
_stream	_stream providers can be connected to in order to pull updates from them with the assumption that the underlying transport guarantees data integrity.
_serial	_serial providers can be connected to in order to pull updates from them with the assumption that the underlying transport does not guarantee data integrity
_consumer	_consumer services can be connected to in order to push updates to them with the assumption that the underlying transport guarantees data integrity.

Two types of JOSN object are provided for:
_delta	The Signal K delta object described in the Signal K standard.
_value	The value object described as part of the full Signal K object in the Signal K standard, with each value update sent as an individual message with some additions.

### Motivation

There is a strong desire in the community to send and receive Signal K values over a range of transports. This is consistent with the transport agnostic approach to developing Signal K to date. There is a danger with this approach that a radically different set of rules may be defined for each transport, making implementation and maintenance more difficult.

This proposal suggests a common way of sending Signal K values over a range of transports. Demand has been shown for serial, plane TCP socket, Web Socket and MQTT and the features of these transports has been considered.

### Definitions

Message	A delta or value JSON object.
Provider	A network node that sends messages.
Consumer	A network node that receives messages.
Service	A server that acts as a provider or a consumer.
Publisher	A provider that does not send messages unless to receives a subscription.

#### 1.1 Discovery

APPLICATION	The following conditions apply when a service is made available over TCP.

SHOULD 1.1	A service SHOULD be discoverable using mDNS.

MUST 1.2	IF a service is discoverable using mDNS;
it MUST use a service name made up of:
A	_signalk
B	followed by the type of message:  -delta or -value
C	followed by the type of service: -stream, -serial or -consumer
D	followed by the intermediate transport if any e.g. -mqtt
E	followed by ._tcp
e.g.	
_signalk-delta-stream._tcp
_signalk-value-consumer-mqtt._tcp
_signalk-delta-serial._tcp

#### 1.2 Default ports

[...]

PR #559, opened 2019-12-17
feature: specify streaming over TCP in more detail

tkurki:
This PR tries to include the bare minimum additions needed to specify a usable TCP streaming service. The more ambitious #294 stalled. Both Artemis and Node server already have streaming over TCP implemented, just need to check that they match this spec / alter the spec if needed.

### DIFF
--- a/gitbook-docs/SUMMARY.md
+++ b/gitbook-docs/SUMMARY.md
@@ -12,7 +12,7 @@
 * [API]()
   * [URLs and Ports](urls_ports.md)
   * [REST API](rest_api.md)
-  * [Streaming WebSocket API](streaming_api.md)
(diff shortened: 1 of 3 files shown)


tkurki:
I combed through the spec for existing specification for TCP related stuff and found
- 8375 is already specified
- the discovery http response already uses `signalk-tcp`, so I think it makes more sense to define that as the mdns identifier explicitly
- on nth thought I think it makes much more sense to make the default subscription policy `none`, as otherwise the server may send nontrivial amounts of data upon connection before the client sends the unsubscribe message

tkurki:
I force pushed a version that takes the stuff in the previous comment into account.

PR #570, opened 2020-04-29
feature: add detail to delta error

jncarter123:
Previously if the delta was missing a path or value, the logged error message
was "Illegal value in delta: " along witht the pathValue. This is a bit misleading.

The new error message provides detail about whether the path and/or value is
missing along with source info.

Example of new message:
Delta is missing path in {"value":294.17}
from source {"label":"ydwg-2000","type":"NMEA2000","pgn":130312,"src":"41","instance":"0"}

### DIFF
--- a/src/fullsignalk.js
+++ b/src/fullsignalk.js
@@ -154,7 +154,7 @@ function handleNmea2000Source(labelSource, source, timestamp) {
   delete existing.n2k.label
   delete existing.n2k.instance
   delete existing.n2k.type
-  
+
   if(source.instance && !labelSource[source.src][source.instance]) {
     labelSource[source.src][source.instance] = {}
   }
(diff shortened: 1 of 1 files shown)


tkurki:
And sorry about the late response..

src/fullsignalk.js
```diff
@@ -184,10 +184,23 @@ function addValues(context, contextPath, source, timestamp, pathValues) {
 }
 
 function addValue(context, contextPath, source, timestamp, pathValue) {
-  if (_.isUndefined(pathValue.path) || _.isUndefined(pathValue.value)) {
-    console.error("Illegal value in delta:" + JSON.stringify(pathValue));
+  let errMessage = ""
```
tkurki, line 187: For some reason I dislike  string concatenation - probably because the ternary makes it a bit hard to comprehend. Imho a better alternative is to use an array: push the error messages to it, then check for length  and `` `...${errs.join(' and ')}...` `` to combine the messages into a readable error message.

tkurki (changes requested):

PR #571, opened 2020-05-01
[WIP]  add definition for generic data

jncarter123:
The Generic Data specification provides a framework for data that falls
outside of standardized sources and is dependent upon vessel usage. This
currently includes temperature and pressure data.

As an example NMEA provides a list of temperature sources but it also
allows for use defined sources #129-252. The user defined sources are now
defined in this specification. An example of user defined temperature would
be generic.temperatures.userDefined129.0.temperature


tkurki:
What should happen next to this one? Any updated thinking since you submitted it?

tkurki:
No reply, I assume this is stalled. Feel free to reopen.

Issue #163, opened 2016-01-31
Satellites in view


tkurki:
Extend schema to support satellites in view information as available in GSV and PGN 129540

tkurki:
https://github.com/SignalK/n2k-signalk/issues/12

faceless2:
I'll start the ball rolling then:

```
"gnss": {
    "satellites": {
        "constellation": {
            "gps": {
                "1": {
                    "elevation": 85,
                    "azimuth": 24,
                    "distance": /*not sure this is useful */,
                    "snr": -76,
                    "active": true      /* whether the satellite is being used for a fix */
                },
                /* keys continue, key being the PRN of each satellite */
            },
            /* keys continue, being other constellations of satellite eg GLONASS */
        }
    },
}
```

Justification:

1. Keying the satellites by PRN makes updating the model easier, and no arrays are required.
2. An "active" field means you can combine the GSV and GSA NMEA0183 messages into one - these are all the satellites in view, and some of them are also in use (active).
3. I'm not familiar with the N2K PGN referenced so would need clarification on the meaning of "Range Residuals" is and what values of "Status" are other than "Used". If it's just "Used or "Not Used' then the "active" field would cover it.

tkurki:
Is there something that should appear as a sibling of constellation? Why should it be there?

Other than that I'm good with this - can you take it further?

I have no in depth knowledge of the 129540 details either.

faceless2:
You're correct - either "constellation" or "satellites" is redundant and could be removed. "satellites" is more immediately intuitive, all else being equal I'd suggest we keep that one.

By "take it further" you're going to make me learn to use Git and submit a pull request aren't you? Argh. I knew this day would come, but I have been putting it off :-) Will try to get up to speed on this and on the spec structure.

sumps:
Its easier than you think, I had never used it and when I met the guys at METs I promised I would have a go. Once I sat down and started, I found the GitHub part was easy and the only thing holding me back was my knowledge (zero) of JSON. 
You do not need to clone a repository locally on your PC, just start by going to one of the files in the Master branch that you want to change and once you are looking at the contents of the file just click on the Edit (pencil) icon. This allows you to immediately start editing the file and automatically creates an auto-named branch of Master with just the file you are editing. Once you have done the change, you click to create a PR, add your notes and justification for the change and post it. Then others will review and merge it for you. Go on give it a go !
[...]

PR #572, opened 2020-05-10
Feat: add satellites definition to GNSS object in navigation group (non-breaking)

fabdrol:
This PR implements the following paths in the `navigation` group, with the goal of making it possible to express the satellites of a GNSS fix in Signal K. Use cases include:

- Making more informed decisions on the quality of a fix
- Showing a polar graph with all satellites relative to the receiver
- Etc

The PR would add the following paths to the specification, without breaking anything else:
```
navigation.gnss.satellites.<PRN code>.elevation
navigation.gnss.satellites.<PRN code>.pseudoRandonNoiseCode
navigation.gnss.satellites.<PRN code>.azimuth
navigation.gnss.satellites.<PRN code>.signalToNoiseRatio
```

Satellites are identified by PRN code, which is unique to each satellite. `signalToNoiseRatio` has a unit `dB` (decibells), which is *not* a SI base unit, but sanctioned by SI anyway.

### DIFF
--- a/schemas/groups/navigation.json
+++ b/schemas/groups/navigation.json
@@ -305,13 +305,13 @@
         "layline": {
           "type": "object",
           "description": "The layline crossing the current course",
-          "properties":{
-            "distance":{
+          "properties": {
+            "distance": {
               "description": "The current distance to the layline",
               "$ref": "../definitions.json#/definitions/numberValue",
               "units": "m"
             },
-            "time":{
+            "time": {
               "description": "The time to the layline at current speed and heading",
               "$ref": "../definitions.json#/definitions/numberValue",
               "units": "s"
@@ -321,13 +321,13 @@
         "oppositeLayline": {
           "type": "object",
           "description": "The layline parallell to current course",
-          "properties":{
-            "distance":{
+          "properties": {
+            "distance": {
               "description": "The current distance to the layline",
               "$ref": "../definitions.json#/definitions/numberValue",
               "units": "m"
             },
-            "time":{
+            "time": {
               "description": "The time to the layline at current speed and heading",
               "$ref": "../definitions.json#/definitions/numberValue",
               "units": "s"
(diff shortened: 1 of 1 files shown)


tkurki:
Imho all spec additions need corresponding sample/test data, making sense of json schema is hard enough.

It says _satellites in view_. How do satellites get purged from full?

What about when you have more than one gps?

Is the data set the same regardless of the positioning system?

I would like to see this addition tied to concrete implementations in n2k-signalk and nmea0183 parser.

Using `numberValue` puts source data and timestamps under each individual value. Aren’t all the values transmitted as a single message?

Furthermore it would be great to exercise this data to create a sample ui, perhaps a map of the sky like receivers show?

fabdrol:
@tkurki we had some discussion about the merits of providing test data and use cases as you suggested in your post. I think it's a bad order to provide implementation before specification, as I strongly believe that implementation should follow specification. Also, I think that demanding implementation in order to test specification will slow down changes massively. 

That said, I will commit to making PRs in n2k-signalk and nmea0138 parser once this is approved and part of the specification. 

In order to answer your questions about GPS data (which are relevant to this PR, imo), I've asked @paul_dy to weigh in on the questions about GPS systems in general, and about test data.

sumps:
Here is some test data - from our GPS160 in "Tri-Nav" professional mode which shows the latest GNSS sentences. Note the different Talker IDs - GP = GPS, GA = Galileo, GL = GLONASS and GN = GNSS (combined). The actual position sentences are GN as they are calculated using all three constellations, while the GSV sentences have the relevant talker ID based on which satellite signals it is related to.

I could set the GPS160 to a GPS only or GALILEO only mode and then the Talker ID for all sentences would be GP or GA, etc.

[GPS160-Pro-Mode.zip](https://github.com/SignalK/specification/files/4890444/GPS160-Pro-Mode.zip)

The GSA sentence lists all satellite PRNs for the satellites used in the position fix, while the GSV sentences show the status of all of the satellites currently in view. You would need to purge all no longer used satellites from Full, every second so Full only lists the satellites currently in view.

The datasets NMEA0183 or NMEA2000 should be the same for all GNSS systems, assuming correct adoption of the standard. 

If you had multiple GNSS systems on the boat, each one could be outputting this level of data and so you would need to support multiple sources.

sbender9:
I’m agreement that this is the time to go and do the implementation. That way you can validate that the spec is complete before merging it. Too many times something comes up in implementation that was missed when writing the spec.

tkurki:
>additions need corresponding sample/test data, making sense of json schema is hard enough.

I was referring to example/test full & delta data in this comment. We have previously merged schema PRs that were not actually doing what they were supposed to. TDD, you know. 

And I would start the TDD process from real world data in 0183 parser, now that Paul has graciously provided some.

If I were to start SK in Github today I'd seriously consider a monorepo approach. Then the NMEA0183, N2K and specification changes could be a single PR, bringing the RI and specification forward in atomic steps. Not that it needs atomicity, just trying to demonstrate the thinking that the spec and RI should move forward together. We still have inconsistencies that you've witnessed yourself like https://github.com/SignalK/signalk-server-node/pull/1076 from adding stuff to the specification and failing to match it in implementation.

See #562 for an example of trying to practise what I'm preaching here.

> slow down changes massively

Things are slowed down by not following up on the PRs and/or getting stuck in debate. My questions above still remain unanswered...

tkurki:
See #163 for earlier work in this direction.

PR #580, opened 2020-07-27
New propulsion paths

ba58smith:


ba58smith:
Don't merge yet - I still need to add some test data.

PR #581, opened 2020-07-27
Bilge paths

ba58smith:


ba58smith:
I still need to see about adding some test data for this, so don't merge it yet. Ooops!

PR #582, opened 2020-07-30
Add vacuum, current, voltage

ba58smith:
npm test run and passed.

### DIFF
--- a/schemas/groups/propulsion.json
+++ b/schemas/groups/propulsion.json
@@ -96,7 +96,7 @@
         "transmission": {
           "type": "object",
           "description": "The transmission (gear box) of the named engine",
-	  "properties": {
+          "properties": {
             "gear": {
               "type": "object",
               "description": "Currently selected gear the engine is in i.e. Forward, Reverse, etc.",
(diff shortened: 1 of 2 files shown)


schemas/groups/propulsion.json
```diff
@@ -213,6 +213,21 @@
               "description": "Average fuel rate of consumption",
               "$ref": "../definitions.json#/definitions/numberValue",
               "units": "m3/s"
+            },
+            "vacuum": {
+              "description": "Vacuum measured at the fuel filter - an indication of the state of the fuel filter. The higher the vacuum, the more clogged the filter.",
+              "$ref": "../definitions.json#/definitions/numberValue",
+              "units": "Pa"
+            },
+            "current": {
```
tkurki, line 222: I have trouble having _current_ and _voltage_ under `fuel`. Maybe they should go one level up in the hierachy?

Furthermore we generate the release notes from PRs and it would make much more sense if there were separate entries for vacuum and support for electrical engines.

tkurki (changes requested):

schemas/groups/propulsion.json
ba58smith: Yeah, not sure what I was thinking - maybe amps and volts at the fuel pump? But to be honest, I did this PR so long ago, I don't even have the git files that this change came from. And the `vacuum` item isn't really all that important - I just did it for my own boat, honestly. :-) Feel free to close this PR if you don't want to mess with making the change yourself. Sorry, @tkurki !

PR #584, opened 2020-08-02
Add generator as a new top level

ba58smith:
per @tkurki, moved common electrical properties out of /groups/electrical.json into definitions.json.

then, created /groups/generator.json, along with valid test data.

`npm test` fails with:
`     AssertionError: There should be no missing schema uris, but found https://signalk.org/specification/1.0.0/schemas/groups/generator.json: expected [ Array(1) ] to have a length of 0 but got 1`
@tkurki said to submit the PR with this, and that he would investigate.

### DIFF
--- a/schemas/definitions.json
+++ b/schemas/definitions.json
@@ -904,6 +904,326 @@
           }
         }
       }
-    }
+    },
+    "identity": {
+          "type": "object",
+          "title": "Electrical ID",
+          "description": " Common ID items shared by electrical items",
+          "properties": {
+            "name": {
+              "type": "string",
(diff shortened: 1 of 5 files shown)


tkurki:
New schema files need to be added also here: https://github.com/SignalK/specification/blob/master/src/index.js#L22-L35

ba58smith:
@tkurki - OK. But I'm now away from the boat, and my Pi, for three weeks. I can fix this when I return, or you can make that change (for this PR, and for the new Bilge top level PR), so you can merge the result.

ba58smith:
@turki, I added the new files where you said to add them (in your comment above), but still, it fails the test, and I have no idea why.

PR #588, opened 2020-10-02
style: add comments

c33howard:
Add code comments to the code to aid understanding.

I'm trying to wrap my head around some of the details of the signalk implementation and adding comments seemed like the most efficient way to do this.  I figured others would benefit from the work.

There's some very light refactoring in here where it simplified the explanation.

### DIFF
--- a/src/fullsignalk.js
+++ b/src/fullsignalk.js
@@ -49,7 +49,7 @@ FullSignalK.prototype.retrieve = function() {
 
 FullSignalK.prototype.addDelta = function(delta) {
   this.emit('delta', delta);
-  var context = findContext(this.root, delta.context);
+  var context = findContextAndUpdateIdentity(this.root, delta.context);
   this.addUpdates(context, delta.context, delta.updates);
   this.updateLastModified(delta.context);
 };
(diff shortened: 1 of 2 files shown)


src/fullsignalk.js
```diff
@@ -76,66 +76,114 @@ FullSignalK.prototype.deleteContext = function(contextKey) {
...
+
   return context;
 }
 
 FullSignalK.prototype.addUpdates = function(context, contextPath, updates) {
-  var len = updates.length;
-  for (var i = 0; i < len; ++i) {
-    this.addUpdate(context, contextPath, updates[i]);
-  }
+  let that = this;
```
tkurki, line 118: Use an arrow function to get rid of `that` usage.

src/fullsignalk.js
```diff
@@ -76,66 +76,114 @@ FullSignalK.prototype.deleteContext = function(contextKey) {
...
+  // descend into the sources element of the context, creating elements as needed
   parts.reduce((cursor, part) => {
     if(typeof cursor[part] === 'undefined') {
       return cursor[part] = {}
     }
     return cursor[part]
   }, this.sources)
+
+  // Uh, shouldn't something be done with the result of the reduce?  What if
+  // the pointed to value isn't found?
```
tkurki, line 158: You already documented what this does: _creating elements as needed_, using reduce cursor as a temp variable.

src/fullsignalk.js
```diff
@@ -177,72 +225,100 @@ function handleOtherSource(sourceLeaf, source, timestamp) {
 }
 
 function addValues(context, contextPath, source, timestamp, pathValues) {
-  var len = pathValues.length;
-  for (var i = 0; i < len; ++i) {
-    addValue(context, contextPath, source, timestamp, pathValues[i]);
-  }
+  pathValues.forEach(function(pathValue) {
```
tkurki, line 228: An arrow function would be a bit more readable I think.

tkurki (changes requested):
All in all a really nice job! I've done the same a few times in open source projects, documenting how an algorithm works just to wrap my head around it. Nice to be on the receiving end!

Would you mind addressing my minor comments?

src/fullsignalk.js
c33howard: I'm still confused.  All this does is create elements, it doesn't actually update anything.  Notice that context and timestamp aren't referenced in the body of the method.  (Which means my comment is wrong, it's descending into this.sources, not context.)

src/fullsignalk.js
c33howard: Ah, nice.  Thanks.  I'm new to javascript and this was helpful.

src/fullsignalk.js
c33howard: OK

src/fullsignalk.js
tkurki: Ok, so 
- it is handling this.sources and `context` is not needed nor used here, should be removed
- it is not updating timestamp, which it should be doing, so a bug - add a FIXME comment in this documentation PR?
- your comment about descending into sources of the context is not correct, so to get this PR merged please remove that

PR #592, opened 2020-10-28
fix: clarify PUT use

sbender9:


gitbook-docs/put.md
```diff
@@ -1,11 +1,12 @@
 # PUT Requests
 
-PUT requests are sent to a server to request a change to a value. For example, a client would use PUT to switch the
-anchor light on or off, change the heading of the autopilot, or set position of the anchor.
+PUT requests are sent to a server to request an action to be taken. For example, a client would use PUT to switch the anchor light on or off, change the heading of the autopilot, or change the current input on a stereo.
+
+Note that this is very different than updating the current state of something. Use a delta update message to report on somethings current value. For example. to report on the current wind speed from a sensor. See [Delta Format](data_model.md)
```
tkurki, line 5: I would leave the first sentence out, because imho it suffers from the same confusion _update the current state_ versus _report the current value_.

_somethings current value_ => "the current value of someting"?

Punctuation is off in _For example._

I usually try to express everything so that the link is inline and not use a separate _See X_ expression.

gitbook-docs/put.md
```diff
@@ -1,11 +1,12 @@
...
 
 See [Request/Response](request_response.md) for more details on request/response in Signal K.
 
-## Making a Request to Change a Value
+## Making a Request To Take an Action
 
 To change a value, a PUT request should be sent via HTTP or using a Signal K __put__ delta.
```
tkurki, line 11: This needs a little rephrasing also?

gitbook-docs/put.md
```diff
@@ -1,11 +1,12 @@
 # PUT Requests
 
-PUT requests are sent to a server to request a change to a value. For example, a client would use PUT to switch the
-anchor light on or off, change the heading of the autopilot, or set position of the anchor.
+PUT requests are sent to a server to request an action to be taken. For example, a client would use PUT to switch the anchor light on or off, change the heading of the autopilot, or change the current input on a stereo.
```
tkurki, line 3: Would something like this be a useful addition:

_An action as opposed to reporting a new value is characterised by the possibility of failure: for example the autopilot may not accept the new heading or we may simply fail to communicate with the equipment in question and the request will fail with a timeout._

tkurki (changes requested):
Definitely good changes, let's get this merged.

What do you think of merging the PUT and Request response segments? There's quite a bit of overlap there, no need to keep them separate.

joelkoz:
I have a question about this as I am working on an automated interior light switch (for my engine room to be exact).  Has there been a determination of what device(s) owns the state?  For my light switch example, my intuition would tell me that the light switch devices knows if it is on or off. It can report its state obviously. However, when an external PUT request comes in to the server, what happens? Does the sk server simply update its internal model then send out an update to all subscribers? Does the request get forwarded to the device who owns the state, and it is up to the device to acknowledge the PUT, change its state, then report an update via a delta push? In the first example, the sk server owns the state. In the second, the switch does

The easy implementation is for the sk server to update its internal state, and push out that update. If the device that owns it can not accommodate, it could always send a delta to change it back. While that is the easy implementation,  in practice it does involve some quick "on/off again" chatter that might affect other automations in the system.

I am actively searching for this answer, which brought me to the specs, and then to this PR.  This is a detail that would at least be worth mentioning in this section, even if the answer is "its currently undefined" or "its up to the implementation to define this exact behavior."

sbender9:
When a PUT requests comes in, it is forwarded to the device/plugin/etc .that publishes the path. It responds per the Request/Reponse specification. The server does not update its internal state at all. The device needs to send out an update delta when the state actually changes.

PR #596, opened 2020-11-14
[WIP] SKv2 Food for Thought

tkurki:
I am adding this PR and the document thereof to serve as a living document for stuff that we could or should implemement in the next "big" Signal K version.

Not sure how things will work out, but the idea is to discuss in the PR and then edit the doc to contain a somewhat coherent view of things.

Let's see how this turns out - please, discuss here if you have something on your mind or amend this PR by issuing PRs to it.

### DIFF
--- a/v2-food-for-thought.md
+++ b/v2-food-for-thought.md
@@ -10,6 +10,12 @@ There are now (2020-11-14) no real plans as to when and how the next version sho
 
 # Stuff
 
+#### From a single data model to APIs with functionality
+
+V1 was mostly about a single, hierarchical data model. In the end the hierarchy does not provide that much value, other than grouping the data for humans. The problem is that a single hierarchy is not expressive enough and we end up in endless discussions about should the engine alternator be part of the propulsion or the electrical system.
+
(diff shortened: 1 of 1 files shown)


panaaj:
Some inital thoughts based on my recent experience...
1. Where there are keys that represent values of interest that have multiple source types (e.g. headingTrue, heading Magnetic or courseGreatCircle, courseRhumbline) provide a path (eg. heading, course) which is populated with the preferred source (nominated by the user in server settings) that way apps can just query the `heading` or `course` path to get the correct value.
This will promote a conistency across all apps without preferences being set per application.

2. Where keys/paths are intended for use by specific devices (e.g. autopilot -> steering.autopilot) the spec should provide clear guidance on how it is intended this path is used. Preferably with example use cases which include sequence diagrams.
I suspect we don't want a scenario where a chart plotter overwriting values placed there by the autopilot.

preeve9534:
Tidy up the convention adopted for tank keys.

My suggestion would be to move the essentially arbitrary fluid type name out of the path and into meta information so that tank keys simply become: tanks.*n*.  In application contexts where Signal K sees a named fluid type in its data feed (e.g. N2K) then the name supplied by the data feed would provide a default value for the meta.

tkurki:
Added these to the doc. You can also do PRs to add to this  branch, then we have a comment thread for each addition.

PR #599, opened 2020-11-26
Add resources section

panaaj:
Added resources section to consider use of resources beyond the standard ones defined in v1 spec.


tkurki:
Merged as 7d6eeb06fc4674af99f8eee907048733fd4c51c1

PR #601, opened 2020-12-09
feat: add new schema group "winches"

preeve9534:
## Motivation

1. The next release of the N2K specification will include three new Windlass Network Messages PGNs (see [this Technical Bulletin](https://www.nmea.org/Assets/20190613%20windlass%20amendment,%20128776,%20128777,%20128778.pdf).
2. The Signal K specification has no framework for windlass data.

## Proposal
1. Accept that a windlass is a type of winch (as is a capstan and maybe even hoist) and focus this discussion on supporting winches as a top-level group in Signal K.
2. Accept that we model the winches group structure simply as winches._id_, where _id_ is an arbitrary identifier (in the case of an N2K connected winch this would be the installation's N2K instance number).
3. Take this draft pull request as an RFC which will allow refinement of the detail of how we represent winch data in Signal K.

### Windlass identity

Common to all winches and just like some other installation identities in Signal K.
```
identity {
  name:
  location:
  dateInstalled:
  manufacturer {
    name:
    model:
    URL:
  }
}
```

### Winch properties

Common to all winches.
```
type: [ "capstan" | "generic" | "hoist" | "windlass" ]
state: [ "deploying" | "retrieving" | "stopped" ] // what the winch is currently doing
speed: n, n>=1 // logical speed - a meta object may map this into the physical values
inService: [ "no" | "yes" ] // whether or not the winch is available for use
lastOperation: { // changes when state and or speed change
  direction: [ "deploy" | "retrieve" ]
  speed: n, n>=1
  duration: s
}
meta: { // static winch data
  availableSpeeds: [
    {
      id: n,n>=1,
      rate: m/s
    }
  ]
  powerType: [ "electrical" | "hydraulic" | "manual" ]
  spoolDimensions: {
    diameter: m
    length: m
  }
}
```

### Windlass and capstan properties

Additional properties for windlasses and capstans.
```
deploymentState: [ "docked" | "nearlyDocked" | "deployed" | "fullyDeployed" ]
rodeType: [ "rope" | "chain" ]
rodeCounter: m
lineSpeed: m/s
```

This draft pull request includes a very draft winches.json that reflects the straw man for those who like that sort of thing.

### DIFF
--- a/schemas/groups/winches.json
+++ b/schemas/groups/winches.json
@@ -45,13 +45,18 @@
       "description": "Properties common to all winch types",
       "type": "object",
       "properties": {
+	"type": {
+	  "description": "The type of winch installation",
+	  "type": "string",
+	  "enum": [ "capstan", "generic", "hoist", "windlass" ]
+	},
         "state": {
           "description": "Current operating state of the winch",
           "type": "string",
           "enum": [ "deploying", "retrieving", "stopped" ]
         },
         "speed": {
-          "description": "Winch current speed in range 1..n",
+          "description": "Winch current speed in range 1..n (1 is slowest)",
           "$ref": "../definitions.json#/definitions/numberValue"
         },
         "inService": {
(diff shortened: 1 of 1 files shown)


mmfriedel:
Anyway to upvote this ? I would like to see it in the spec.

PR #606, opened 2021-03-23
Update dependencies

tkurki:
Based on #605 so merge that first.


tkurki:
Related to https://github.com/SignalK/signalk-server/issues/1295

PR #609, opened 2021-03-30
dewPoint -> dewPointTemperature and humidity -> relativeHumidity

GHBLoos:
I think this should be `dewPointTemperature` instead of just `dewPoint`

### DIFF
--- a/schemas/groups/environment.json
+++ b/schemas/groups/environment.json
@@ -28,8 +28,8 @@
           "units": "ratio",
           "$ref": "../definitions.json#/definitions/numberValue"
         },
-        "dewPoint": {
-          "description": "Dewpoint in zone",
+        "dewPointTemperature": {
+          "description": "Dewpoint temperature in zone",
           "units": "K",
           "$ref": "../definitions.json#/definitions/numberValue"
         },
--- a/schemas/groups/environment.json
+++ b/schemas/groups/environment.json
@@ -81,7 +81,7 @@
           "$ref": "../definitions.json#/definitions/numberValue",
           "units": "Pa"
         },
-        "humidity": {
+        "relativeHumidity": {
           "description": "Current outside air relative humidity",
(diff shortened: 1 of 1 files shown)


GHBLoos:
Also `outside\humidity` should be `outside\relativeHumidity`

tkurki:
The problem with renaming paths is that it is essentially a breaking change: anything already using that path is rendered invalid per the schema and will lose the connection to the metadata in the schema, including the unit.

Per semantic versioning this would mean a new major version. I am not very keen to do that for something this small.

But we have also other lighter options:
(a)  don't change the path, just the description
(b) add the new path, keep the old one
(c) add the new path, keep the old one, mark it DEPRECATED

I don't think we have a proper way to mark a path deprecated, but adding a text like _DEPRECATED, replaced by xx_ is a cheap way to make this visible in the paths reference.

GHBLoos:
Yeah, I understand that changing the path without informing is not the right way. I am  not sure it will give any troubles, because a path can still be used. But it's good to have a protocol how changes in paths are implemented.
So we should agree on that protocol. Maybe like this
Minor versions -> don't delete the old path, add a new path (or any of the other optiona a, b and c above)
Major versions -> delete the old path

tkurki:
Re schema versioning: did you read first the documentation at first https://signalk.org/specification/1.5.0/doc/versioning.html?

>not sure it will give any troubles

I already said that any existing use of this path *will lose unit metadata* with this change, so as it is this PR is not mergeable in my opinion.

What do you want to do next?

GHBLoos:
Sorry, I should have RTFM first. I will do that, but that probably won't be one of the next days. 
I think it's very impressive what you guys have already done and described. I am just not at the right level at the moment. That will take some time.

tkurki:
Created #625 that deprecates the wrong versions and adds the correct ones.

PR #627, opened 2021-10-23
add "estimatedTimeOfArrival" to "nextPoint"

panaaj:
Addition of `estimatedTimeOfArrival` to `navigation/course/nextPoint` to hold the estimated arrival time to the current destination location.

Currently only estimatedTimeOfArrival key is in `activeRoute` and that is the eta to arrive at the final destination at the end of a route.


tkurki (changes requested):
Like #628, would be great to have test data for this.

panaaj:
added tests

PR #628, opened 2021-10-23
Add "arrivalCircle" to "nextPoint"

panaaj:
Addition of `arrivalCircle` to `nextPoint` to allow an arrival alarm to be implemented.

### DIFF
--- a/schemas/groups/navigation.json
+++ b/schemas/groups/navigation.json
@@ -94,6 +94,11 @@
                 "position": {
                   "description": "The position of nextPoint in two dimensions",
                   "$ref": "../definitions.json#/definitions/position"
+                },
+                "arrivalCircle": {
+                  "description": "The radius in meters of a circle centered around nextPoint position, when entered will signal arrival at destination",
+                  "units": "m",
+                  "$ref": "../definitions.json#/definitions/numberValue"
                 }
               }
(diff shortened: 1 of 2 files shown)


tkurki (changes requested):
I think it would be useful to add this to https://github.com/SignalK/specification/blob/master/test/data/full-valid/course-full_tree.json#L48

Oh I just noticed that the hrefs in that doc put resources under a vessel...

panaaj:
Added tests for `nextPoint.arrivalCircle`.
I have noticed two things in the test data that don't align with the text in the proposed CourseAPI.
1. `href` test value is under vessels _("href": "/vessels/vessels.urn:mrn:imo:mmsi:230099999/resources/waypoints/...")_
 when in practice `resources` is at the root of `./signalk/v1/api`.
2. `arrivalCircleEntered` notification is defined as `notifications.arrivalCircleEntered` where the proposed notification path in the Course API PR is `notifications.navigation.course.arrivalCircleEntered` 

Should these be updated in the test file?

tkurki:
I'll merge this, submit a new PR for fixes?

tkurki:
Oh, please rebase. Would do it myself nicely if the branch were under this repo ;-)

PR #648, opened 2022-12-26
fix: deprecated package updated

KEGustafsson:
Update npm package to remove `core-js` warning about deprecated version.
Links to Signal K Server vulnerabilities and deprecated fix.
https://github.com/SignalK/signalk-server/pull/1508

Tested with updated npmjs version of signalk-schema@1.6.0 version with SignalK Server fixes.
As version 1.6.0 is not visible here and 1.7.0 seemed to cause some issues, This might need more investigation.
Which commit is npmjs 1.6.0?

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
   "dependencies": {
     "JSONStream": "^0.7.4",
     "debug": "^4.3.1",
-    "json-schema-ref-parser": "^9.0.7",
+    "@apidevtools/json-schema-ref-parser": "^9.1.0",
     "lodash": "^4.17.21",
     "tv4": "^1.2.7",
     "tv4-formats": "^3.0.3"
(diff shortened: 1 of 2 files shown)


tkurki:
[Tag v1.6.0](https://github.com/SignalK/specification/commits/v1.6.0) is here in GH and

```
git merge-base v1.7.0 v1.6.0
c71c367032488576121d105b72f10d9afa164c11
```

shows that all we are missing from master / v1.7.0 history is 03c5b03b5506ba113e2960a7a031820c8b748707 (v1.6.0 version bump) and c9d733b65277addd592a8490340ce7967a7344e4 that seems to be the same as 38147bc0ff7d16edb78d3bc55f131940799f528b that we have in master.

So nothing to fix i think.

PR #652, opened 2023-03-28
Add DC-DC convertors

stefanor:
Like chargers, but without stages and modes of charge.

### DIFF
--- a/schemas/groups/electrical.json
+++ b/schemas/groups/electrical.json
@@ -520,6 +520,38 @@
         }
       }
     },
+    "converters": {
+      "description": "Data about a voltage step up/down DC-DC converter",
+      "patternProperties": {
+        "(^[A-Za-z0-9]+$)": {
+          "type": "object",
+          "title": "Charger",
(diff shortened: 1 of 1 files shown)


bergie:
Some Victron DC-DC converters can run as a pure converter, or as a charger (with charge modes). Is the idea that the device would then show up either as a converter or a charger depending on its configuration?

stefanor:
> Some Victron DC-DC converters can run as a pure converter, or as a charger (with charge modes). Is the idea that the device would then show up either as a converter or a charger depending on its configuration?

That's certainly a possibility.

My thinking primary thinking is that it's about how the device is used. If it's used as a battery bank charger, then `electrical.chargers` is more appropriate. If it's the source of energy for a bus, then `electrical.converters` is more appropriate.

stefanor:
> Some Victron DC-DC converters can run as a pure converter, or as a charger (with charge modes). Is the idea that the device would then show up either as a converter or a charger depending on its configuration?

Given that, would it make more sense to include `chargerQualities`?

PR #653, opened 2023-04-14
Solar Chargers: Metrics for Panel Power and Yield Today

stefanor:
This is used by signalk-venus-plugin and signalk-victron-ble.


tkurki:
Thanks!

PR #654, opened 2023-04-16
Add power to DC Qualities

stefanor:
signalk-derived-data calculates power for batteries: https://github.com/SignalK/signalk-derived-data/blob/master/calcs/batteryPower.js

This is a useful metric for all DC systems.


elafargue:
Agree, this would be extremely useful

PR #660, opened 2024-02-18
Waypoint ID added to Course Definitions

Techstyleuk:
Updating the specification to be inline with how its parsed in some of the nmea0183-signalk hooks.  for example, APB.  see: https://github.com/SignalK/nmea0183-signalk/issues/239
This is a partial fix

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,15 +39,15 @@
   },
   "homepage": "https://github.com/SignalK/specification",
   "dependencies": {
-    "JSONStream": "^0.7.4",
-    "debug": "^4.3.1",
     "@apidevtools/json-schema-ref-parser": "^9.1.0",
+    "debug": "^4.3.1",
+    "JSONStream": "^0.7.4",
     "lodash": "^4.17.21",
     "tv4": "^1.2.7",
     "tv4-formats": "^3.0.3"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


panaaj:
This should be aligned with the Course API where possible.

Techstyleuk:
I think between:

Course API - the Spec. - nmea0183-signalk (APB)

there are inconsistencies, APB writes the Waypoint as ......nextPoint.ID, but this is not in the spec.

I added to the spec.  If this shouldn't be in the spec. or if Course API should be updated to reflect the change, let me know what you are thinking?

Having read the API, I am not sure what I would put to close that loop out

panaaj:
I think that there are some paths written by plugins that addressed a specific need / use case at the time.
That said,  there are also some areas to address from an API perspective, for example  aligning data coming from NMEA devices.

For me, if the waypoint id is meaningful (has relevance) only to the device that transmitted it, I would question the need for a path in the spec.
But if we are clear about what role the id has, and if it is of wider interest to other connected devices, then we should add it to the spec and describe the value it holds and the relationship to other course paths (Course paths need to be managed as a collection). 

Some questions that probably need to be answered....
- Is the id only relevant in the context of a waypoint being part of a route?
- What relationship is there between this is and a SignalK waypoint resource?

PR #662, opened 2024-03-17
Add Meteo schema

panaaj:
This PR seeks to add paths to the `environment` group that relate to the processing of AIS weather messages.

By defining the additional attibutes and metadata clients will have available weather station observation and forecast values with the correct units.


panaaj:
@tkurki can you please review and advise if there is anything further required to move this forward.

dirkwa:
Picking this up from the Signal K server side — these meteo/environment units are exactly what's missing in practice today.

Context: a user reported environment metadata (e.g. `environment.water.level` → `{units: "m"}`) disappearing in 2.28.0-beta. The root cause was a server-side regression (now fixed in SignalK/signalk-server#2744 + #2742), but it surfaced because FreeboardSK injects ~18 of these weather units at runtime as a workaround — precisely because they aren't in the base. Its plugin even carries a `@todo remove after merge of specification#662`. So this PR stalling has a concrete downstream cost.

The 18 paths currently relying on that runtime workaround (units the base has no definition for):

```
environment.outside.uvIndex (UVI), cloudCover (ratio), feelsLikeTemperature (K),
horizontalVisibility (m), absoluteHumidity (ratio), precipitationVolume (m)
environment.wind.averageSpeed (m/s), gust (m/s), gustDirectionTrue (rad)
environment.water.level (m), current.set (rad), current.drift (m/s)
environment.water.waves.significantHeight (m), period (s), direction (rad)
environment.water.swell.significantHeight (m), period (s), directionTrue (rad)
```

One open question for @tkurki: on the server thread you've indicated `@signalk/path-metadata` should be the master for well-known paths and units going forward (moving away from the JSON-schema-defined full tree). If that's the direction, should these meteo units land in `@signalk/path-metadata` (which this PR could mirror), or stay authored here in `environment.json` with path-metadata derived from it? Happy to help wire whichever way you prefer — once the units are in the base, FreeboardSK can drop its runtime workaround entirely.

PR #664, opened 2024-04-26
Zones-meta and Notification updates

godind:
Update to change ffrom alarm to notification. Adjusted zones and  operationnal concept.


mdbook/src/notifications.md
```diff
@@ -1,36 +1,58 @@
...
+|------------|--------|
+| normal     | Server generated - The normal value (default) |
+| nominal    | All systems OK |
+| alert      | Indicates a safe or normal condition which is brought to the operators attention to impart information |
+| warn       | Indicates a condition that requires immediate attention but not immediate action |
+| alarm      | Indicates a condition which is outside the specified acceptable range. Immediate action is required to prevent loss of life or equipment damage |
+| emergency  | The value indicates a life-threatening condition |
+
+## Resolving and silencing notifications
+* To silence a notification, the serve or clients, send delta notification replica without the `sound` item the method array.
```
panaaj, line 45: Actioning a notification is an operation of the notification manager. 
As this applies to implementation the expected operation should be outlined here.

godind: I thought about this yesterday, I was including Alarm Clearer plugin as a standard setup in this writing. What do you recommend?

godind: It would be nice to be able to explain how consumers should work the system.

sbender9:
Can you write a larger summary of the changes? Really hard to parse and make sense of the diffs.

tkurki:
Diff ignoring whitespace changes is better https://github.com/SignalK/specification/pull/664/files?w=1

godind:
There was no squash commit for some unknown reason.

Maybe read from my fork here:
https://github.com/godind/specification/tree/Zones-meta/mdbook/src

Not sure how else to do it but pasting the whole thing here. Or I could create another branch with a single commit.

tkurki:
The PR is based on that branch so there is no difference whether you look at the PR or the branch.

If you don't squash then there is no squash, just your commits.

But in terms of understanding what has changed squashing or the long list of commits with less than understandable commit messages is not the problem. You have made so many changes to line breaks without changing the content that it is needlessly hard to understand what are the meaningful changes. This is what Scott is asking - the PR description is pretty cryptic. So please can you summarise how and why this PR changes things?

tkurki:
https://cbea.ms/git-commit/#why-not-how

godind:
What do you suggest to fix this? I'm not sure how I can resubmit in one change without line breaks. I formatted so I could read myself.

It's a learning curve.

Thanks

tkurki:
>So please can you summarise how and why this PR changes things?

godind:
### Notification changes
- Reformulates the base conceptual approach from Alarm to Notification and expresses that and alarms a notification severity state.
- Defines the Normal state as the default data state when path values matches no other zones.
- Changes the approach to clear a Notification from the prior spec which was to send an update the notification path with:
`
"values": [
        {
          "path": "notifications.mob",
          "value": null
        }
      ]
`
to sending an update the notification path with normal state and an empty method:
`
"value": {
            "message": "",
            "state": "normal",
            "method": []
          }
`
- Adds the concept of Silencing and Resolving Notifications: To silence, update notification by removing `sound` array value. Resolve, update notification by setting state to `normal`.
- Moved Alarm Management from Metadata page to this page and renamed to Notification Management as I thought it made more sense in the page.
- Fixed link to metadata page.

### Metadata
- Focus is primarily on zones, reaffirming the base conceptual approach from Alarm to Notification and expresses that and alarms a notification severity state.
- Also reaffirms normal as the state where no zones match the value.
- Added zones configuration exemples to showing where zones state could repeat themselves for different value ranges.
- removed Alarm Management section. I thought it should go in Notifications page as it's about notification management process than meta config settings.

Please read the output as this is my first spec update. I wrote this based on my exchange with Scott and what I understood. I'm not without fault...

I hope what I wrote helps!

PR #665, opened 2024-07-28
Fix state table formatting

mairas:
The table didn't get rendered properly. While the markdown syntax would appear to be correct, maybe the used processor requires whitespace around the header separators.


mairas:
Probably not the root cause since the previous sections are incorrectly formatted as well.

PR #670, opened 2025-09-09
add computer SI units to schema

godind:
Add two SI units
1- B for byte (memory, storage, etc)
2- bps for bit per seconds (bandwidth)

### DIFF
--- a/schemas/definitions.json
+++ b/schemas/definitions.json
@@ -274,7 +274,7 @@
           "quantityDisplay": "Byte",
           "description": "Digital quantity. 1 byte = 8 bits."
         },
-        "bps": {
+        "b/s": {
           "display": "bps",
           "quantity": "Data Transfer Rate",
           "quantityDisplay": "bit/s",


tkurki (changes requested):
These are not SI units to be precise here, but I think useful additions nonetheless.

Please add also at https://github.com/SignalK/specification/blob/1978f556d8790be2b3bfdc9fffcf82a9a913e3c2/scripts/processSchemaFiles.js#L28

schemas/definitions.json
```diff
@@ -267,6 +267,18 @@
...
           "quantityDisplay": "\u03bc",
           "description": "Viscosity in pascal seconds"
+        },
+        "B": {
+          "display": "B",
+          "quantity": "Digital Storage Capacity",
+          "quantityDisplay": "Byte",
+          "description": "Digital quantity. 1 byte = 8 bits."
+        },
+        "bps": {
```
tkurki:
```suggestion
        "b/s": {
```

I think this would be more consistent internally, compared to the other entries that also use slash.

